PR #56, opened 2021-02-02
fix peering guide note + missed generated docs

bcmdarroch:
Fixes this line in the Peering guide:
Screen Shot 2021-02-02 at 2 22 58 PM

Which should look like:
Screen Shot 2021-02-02 at 2 23 06 PM

Also adds up-to-date generated docs, looks like one of the last cleanup PRs missed them.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -37,7 +37,7 @@ provider "aws" {
 // Create an HVN
 resource "hcp_hvn" "example_hvn" {
   hvn_id         = "hcp-tf-example-hvn"
-  cloud_provider = var.cloud_provider
+  cloud_provider = "aws"
   region         = var.region
   cidr_block     = "172.25.16.0/20"
 }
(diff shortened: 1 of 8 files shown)


docs/index.md
```diff
@@ -68,16 +67,15 @@ resource "hcp_aws_network_peering" "example_peering" {
...
+  hvn_id     = hcp_hvn.example_hvn.hvn_id
+  cluster_id = "hcp-tf-example-consul-cluster"
+  tier       = "development"
 }
 ```
 
 ## Schema
 
-### Required
+### Optional
```
roaks3, line 78: Hmm, was this changed by the docs generator? I agree with the change, but wondering if we should also change these 2 fields in the scheme to be optional: https://github.com/hashicorp/terraform-provider-hcp/blob/main/internal/provider/provider.go#L37

roaks3 (approved):
One question, but otherwise LGTM

docs/index.md
bcmdarroch: Oh good point. I wonder why this changed if it's still required in the provider 🤔

PR #57, opened 2021-02-03
Validate Consul datacenter and lowercase the default

alexmunda:
While updating some of the HCP Consul E2E tests to use the new provider, I noticed that we were not validating that the Consul cluster `datacenter` does not contain uppercase letters. 
This PR validates the `datacenter` as such, as well as cleans up some comments.


bcmdarroch (approved):
Good catch! I wonder if there's more validation we need to do, or if the e2e tests will be enough 💭

roaks3 (approved):
Looks great!

PR #58, opened 2021-02-04
HCPE-830 - Add TGW attachment resource

roaks3:
Ticket: https://hashicorp.atlassian.net/browse/HCPE-830

Adds the `hcp_aws_transit_gateway_attachment` resource and data source. The implementation is largely pulled from the initial implementation: https://github.com/hashicorp/cloud-terraform-provider-hcp-internal/pull/8

The inclusion of `aws` in the resource name came from some discussion captured [here](https://hashicorp.atlassian.net/browse/HCPE-872), which might change moving forward.

A simple example of what the resource looks like:
```
resource "hcp_aws_transit_gateway_attachment" "example" {
  hvn_id                        = hcp_hvn.main.hvn_id
  transit_gateway_attachment_id = "example-tgw-attachment"
  transit_gateway_id            = aws_ec2_transit_gateway.example.id
  resource_share_arn            = aws_ram_resource_share.example.arn
  destination_cidrs             = [aws_vpc.example.cidr_block]
}
```

### To Test

The example included here can be used to test the functionality of this resource. Note that the AWS config cannot be the same account as the HVN.

### DIFF
--- a/docs/data-sources/aws_transit_gateway_attachment.md
+++ b/docs/data-sources/aws_transit_gateway_attachment.md
@@ -35,7 +35,7 @@ data "hcp_aws_transit_gateway_attachment" "test" {
 
 - **created_at** (String) The time that the Transit gateway attachment was created.
 - **destination_cidrs** (List of String) The list of associated CIDR ranges. Traffic from these CIDRs will be allowed for all resources in the HVN. Traffic to these CIDRs will be routed into this Transit gateway attachment.
-- **expires_at** (String) The time after which the Transit gateway attachment will be considered expired if it hasn't transitioned into 'Accepted' or 'Active' state.
+- **expires_at** (String) The time after which the Transit gateway attachment will be considered expired if it hasn't transitioned into `ACCEPTED` or `ACTIVE` state.
 - **organization_id** (String) The ID of the HCP organization where the Transit gateway attachment is located. Always matches the HVN's organization.
 - **project_id** (String) The ID of the HCP project where the Transit gateway attachment is located. Always matches the HVN's project.
 - **provider_transit_gateway_attachment_id** (String) The Transit gateway attachment ID used by AWS.
--- a/docs/data-sources/aws_transit_gateway_attachment.md
+++ b/docs/data-sources/aws_transit_gateway_attachment.md
@@ -28,8 +28,8 @@ data "hcp_aws_transit_gateway_attachment" "test" {
 ### Optional
 
 - **id** (String) The ID of this resource.
-- **state** (String) The state of the Transit gateway attachment.
 - **timeouts** (Block, Optional) (see [below for nested schema](#nestedblock--timeouts))
+- **wait_for_active_state** (Boolean) If `true`, the Transit gateway attachment information will not be provided until it is in an `ACTIVE` state. Default `false`.
 
 ### Read-only
 
--- a/docs/data-sources/aws_transit_gateway_attachment.md
+++ b/docs/data-sources/aws_transit_gateway_attachment.md
@@ -29,7 +29,7 @@ data "hcp_aws_transit_gateway_attachment" "test" {
 
 - **id** (String) The ID of this resource.
 - **timeouts** (Block, Optional) (see [below for nested schema](#nestedblock--timeouts))
-- **wait_for_active_state** (Boolean) If `true`, the transit gateway attachment information will not be provided until it is in an `ACTIVE` state. Default `false`.
+- **wait_for_active_state** (Boolean) If `true`, Terraform will wait for the transit gateway attachment to reach an `ACTIVE` state before continuing. Default `false`.
 
 ### Read-only
 
@@ -39,7 +39,7 @@ data "hcp_aws_transit_gateway_attachment" "test" {
 - **organization_id** (String) The ID of the HCP organization where the transit gateway attachment is located. Always matches the HVN's organization.
 - **project_id** (String) The ID of the HCP project where the transit gateway attachment is located. Always matches the HVN's project.
 - **provider_transit_gateway_attachment_id** (String) The transit gateway attachment ID used by AWS.
-- **transit_gateway_id** (String) The ID of the transit gateway in AWS.
+- **transit_gateway_id** (String) The ID of the user-owned transit gateway in AWS.
 
 <a id="nestedblock--timeouts"></a>
 ### Nested Schema for `timeouts`
--- a/docs/data-sources/aws_transit_gateway_attachment.md
+++ b/docs/data-sources/aws_transit_gateway_attachment.md
@@ -1,0 +1,51 @@
+---
+page_title: "hcp_aws_transit_gateway_attachment Data Source - terraform-provider-hcp"
+subcategory: ""
+description: |-
+  The AWS Transit gateway attachment data source provides information about an existing Transit gateway attachment.
+---
+
+# Data Source `hcp_aws_transit_gateway_attachment`
+
+The AWS Transit gateway attachment data source provides information about an existing Transit gateway attachment.
+
+## Example Usage
+
+```terraform
+data "hcp_aws_transit_gateway_attachment" "test" {
+  hvn_id                        = var.hvn_id
+  transit_gateway_attachment_id = var.transit_gateway_attachment_id
+}
+```
+
+## Schema
+
+### Required
+
+- **hvn_id** (String) The ID of the HashiCorp Virtual Network (HVN).
+- **transit_gateway_attachment_id** (String) The ID of the Transit gateway attachment.
+
+### Optional
+
+- **id** (String) The ID of this resource.
+- **state** (String) The state of the Transit gateway attachment.
+- **timeouts** (Block, Optional) (see [below for nested schema](#nestedblock--timeouts))
+
+### Read-only
+
+- **created_at** (String) The time that the Transit gateway attachment was created.
+- **destination_cidrs** (List of String) The list of associated CIDR ranges.
+- **expires_at** (String) The time after which the Transit gateway attachment will be considered expired if it hasn't transitioned into 'Accepted' or 'Active' state.
+- **organization_id** (String) The ID of the HCP organization where the Transit gateway attachment is located. Always matches the HVN's organization.
+- **project_id** (String) The ID of the HCP project where the Transit gateway attachment is located. Always matches the HVN's project.
+- **provider_transit_gateway_attachment_id** (String) The Transit gateway attachment ID used by AWS.
+- **transit_gateway_id** (String) The ID of the Transit gateway in AWS.
+
+<a id="nestedblock--timeouts"></a>
+### Nested Schema for `timeouts`
+
+Optional:
+
+- **default** (String)
+
+
--- a/docs/data-sources/aws_transit_gateway_attachment.md
+++ b/docs/data-sources/aws_transit_gateway_attachment.md
@@ -2,12 +2,12 @@
 page_title: "hcp_aws_transit_gateway_attachment Data Source - terraform-provider-hcp"
 subcategory: ""
 description: |-
-  The AWS Transit gateway attachment data source provides information about an existing Transit gateway attachment.
+  The AWS Transit Gateway Attachment data source provides information about an existing transit gateway attachment.
 ---
 
 # Data Source `hcp_aws_transit_gateway_attachment`
 
-The AWS Transit gateway attachment data source provides information about an existing Transit gateway attachment.
+The AWS Transit Gateway Attachment data source provides information about an existing transit gateway attachment.
 
 ## Example Usage
 
@@ -23,23 +23,23 @@ data "hcp_aws_transit_gateway_attachment" "test" {
 ### Required
 
 - **hvn_id** (String) The ID of the HashiCorp Virtual Network (HVN).
-- **transit_gateway_attachment_id** (String) The user-settable name of the Transit gateway attachment in HCP.
+- **transit_gateway_attachment_id** (String) The user-settable name of the transit gateway attachment in HCP.
 
 ### Optional
 
 - **id** (String) The ID of this resource.
 - **timeouts** (Block, Optional) (see [below for nested schema](#nestedblock--timeouts))
-- **wait_for_active_state** (Boolean) If `true`, the Transit gateway attachment information will not be provided until it is in an `ACTIVE` state. Default `false`.
+- **wait_for_active_state** (Boolean) If `true`, the transit gateway attachment information will not be provided until it is in an `ACTIVE` state. Default `false`.
 
 ### Read-only
 
-- **created_at** (String) The time that the Transit gateway attachment was created.
-- **destination_cidrs** (List of String) The list of associated CIDR ranges. Traffic from these CIDRs will be allowed for all resources in the HVN. Traffic to these CIDRs will be routed into this Transit gateway attachment.
-- **expires_at** (String) The time after which the Transit gateway attachment will be considered expired if it hasn't transitioned into `ACCEPTED` or `ACTIVE` state.
-- **organization_id** (String) The ID of the HCP organization where the Transit gateway attachment is located. Always matches the HVN's organization.
-- **project_id** (String) The ID of the HCP project where the Transit gateway attachment is located. Always matches the HVN's project.
-- **provider_transit_gateway_attachment_id** (String) The Transit gateway attachment ID used by AWS.
-- **transit_gateway_id** (String) The ID of the Transit gateway in AWS.
+- **created_at** (String) The time that the transit gateway attachment was created.
+- **destination_cidrs** (List of String) The list of associated CIDR ranges. Traffic from these CIDRs will be allowed for all resources in the HVN. Traffic to these CIDRs will be routed into this transit gateway attachment.
+- **expires_at** (String) The time after which the transit gateway attachment will be considered expired if it hasn't transitioned into `ACCEPTED` or `ACTIVE` state.
+- **organization_id** (String) The ID of the HCP organization where the transit gateway attachment is located. Always matches the HVN's organization.
+- **project_id** (String) The ID of the HCP project where the transit gateway attachment is located. Always matches the HVN's project.
+- **provider_transit_gateway_attachment_id** (String) The transit gateway attachment ID used by AWS.
+- **transit_gateway_id** (String) The ID of the transit gateway in AWS.
 
 <a id="nestedblock--timeouts"></a>
 ### Nested Schema for `timeouts`
--- a/docs/data-sources/aws_transit_gateway_attachment.md
+++ b/docs/data-sources/aws_transit_gateway_attachment.md
@@ -23,7 +23,7 @@ data "hcp_aws_transit_gateway_attachment" "test" {
 ### Required
 
 - **hvn_id** (String) The ID of the HashiCorp Virtual Network (HVN).
-- **transit_gateway_attachment_id** (String) The ID of the Transit gateway attachment.
+- **transit_gateway_attachment_id** (String) The user-settable name of the Transit gateway attachment in HCP.
 
 ### Optional
 
@@ -34,7 +34,7 @@ data "hcp_aws_transit_gateway_attachment" "test" {
 ### Read-only
 
 - **created_at** (String) The time that the Transit gateway attachment was created.
-- **destination_cidrs** (List of String) The list of associated CIDR ranges.
+- **destination_cidrs** (List of String) The list of associated CIDR ranges. Traffic from these CIDRs will be allowed for all resources in the HVN. Traffic to these CIDRs will be routed into this Transit gateway attachment.
 - **expires_at** (String) The time after which the Transit gateway attachment will be considered expired if it hasn't transitioned into 'Accepted' or 'Active' state.
 - **organization_id** (String) The ID of the HCP organization where the Transit gateway attachment is located. Always matches the HVN's organization.
 - **project_id** (String) The ID of the HCP project where the Transit gateway attachment is located. Always matches the HVN's project.
--- a/docs/data-sources/aws_transit_gateway_attachment.md
+++ b/docs/data-sources/aws_transit_gateway_attachment.md
@@ -39,6 +39,7 @@ data "hcp_aws_transit_gateway_attachment" "test" {
 - **organization_id** (String) The ID of the HCP organization where the transit gateway attachment is located. Always matches the HVN's organization.
 - **project_id** (String) The ID of the HCP project where the transit gateway attachment is located. Always matches the HVN's project.
 - **provider_transit_gateway_attachment_id** (String) The transit gateway attachment ID used by AWS.
+- **state** (String) The state of the transit gateway attachment.
 - **transit_gateway_id** (String) The ID of the user-owned transit gateway in AWS.
 
 <a id="nestedblock--timeouts"></a>
--- a/docs/resources/aws_transit_gateway_attachment.md
+++ b/docs/resources/aws_transit_gateway_attachment.md
@@ -88,6 +88,7 @@ resource "aws_ec2_transit_gateway_vpc_attachment_accepter" "example" {
 - **organization_id** (String) The ID of the HCP organization where the transit gateway attachment is located. Always matches the HVN's organization.
 - **project_id** (String) The ID of the HCP project where the transit gateway attachment is located. Always matches the HVN's project.
 - **provider_transit_gateway_attachment_id** (String) The transit gateway attachment ID used by AWS.
+- **state** (String) The state of the transit gateway attachment.
 
 <a id="nestedblock--timeouts"></a>
 ### Nested Schema for `timeouts`
--- a/docs/resources/aws_transit_gateway_attachment.md
+++ b/docs/resources/aws_transit_gateway_attachment.md
@@ -70,10 +70,10 @@ resource "aws_ec2_transit_gateway_vpc_attachment_accepter" "example" {
 
 ### Required
 
-- **destination_cidrs** (List of String) The list of associated CIDR ranges.
+- **destination_cidrs** (List of String) The list of associated CIDR ranges. Traffic from these CIDRs will be allowed for all resources in the HVN. Traffic to these CIDRs will be routed into this Transit gateway attachment.
 - **hvn_id** (String) The ID of the HashiCorp Virtual Network (HVN).
-- **resource_share_arn** (String, Sensitive) The Amazon Resource Name (ARN) of the Resource Share that is needed to grant HCP access to the Transit gateway in AWS.
-- **transit_gateway_attachment_id** (String) The ID of the Transit gateway attachment.
+- **resource_share_arn** (String, Sensitive) The Amazon Resource Name (ARN) of the Resource Share that is needed to grant HCP access to the Transit gateway in AWS. The Resource Share should be associated with the HCP AWS account principal (see [aws_ram_principal_association](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ram_principal_association)) and the Transit gateway resource (see [aws_ram_resource_association](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ram_resource_association))
+- **transit_gateway_attachment_id** (String) The user-settable name of the Transit gateway attachment in HCP.
 - **transit_gateway_id** (String) The ID of the Transit gateway in AWS.
 
 ### Optional
--- a/docs/resources/aws_transit_gateway_attachment.md
+++ b/docs/resources/aws_transit_gateway_attachment.md
@@ -2,12 +2,12 @@
 page_title: "hcp_aws_transit_gateway_attachment Resource - terraform-provider-hcp"
 subcategory: ""
 description: |-
-  The AWS Transit gateway attachment resource allows you to manage a Transit gateway attachment that attaches an HVN to a Transit gateway in AWS.
+  The AWS Transit Gateway Attachment resource allows you to manage a transit gateway attachment that attaches an HVN to a transit gateway in AWS.
 ---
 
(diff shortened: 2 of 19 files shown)


docs/data-sources/aws_transit_gateway_attachment.md
```diff
@@ -0,0 +1,51 @@
+---
```
roaks3, line 1: The `docs/*` files included here were generated with `go generate`, which pulls in the examples and resource/field descriptions.

docs/data-sources/aws_transit_gateway_attachment.md
```diff
@@ -0,0 +1,51 @@
...
+
+### Required
+
+- **hvn_id** (String) The ID of the HashiCorp Virtual Network (HVN).
+- **transit_gateway_attachment_id** (String) The ID of the Transit gateway attachment.
+
+### Optional
+
+- **id** (String) The ID of this resource.
+- **state** (String) The state of the Transit gateway attachment.
```
roaks3, line 31: The `state` field used here was copied from the internal implementation, but I think it should change to a `wait_for_active_state` flag, or `desired_state`. More detail below:

For this use case, I found some prior discussion from the existing AWS and Google providers:
- https://github.com/hashicorp/terraform-provider-aws/pull/3485
- https://github.com/hashicorp/terraform-provider-google/pull/4797

In short, there seems to be two concerns with using the field `state`:
1. It is merged with the computed `state` of the resource, which can make things behave strangely when trying to resolve whether the `state` has been set by the user or the service ([ref](https://github.com/hashicorp/terraform-provider-google/pull/4797#pullrequestreview-353254440)).
2. It is ambiguous whether the field is meant to wait for the specified state, or filter by the specified state. The [AWS provider is already using state fields as filters](https://github.com/hashicorp/terraform-provider-aws/blob/main/aws/data_source_aws_vpc_peering_connection.go#L120-L130), so it could be confusing if users expect that same usage here.

I came across these two general solutions to the problem that I think we should consider:
1. `wait_for_active_state = "true"` ([ref](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ecs_service#wait_for_steady_state))
2. `desired_state = "ACTIVE"` ([ref](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_instance#desired_status))

internal/provider/resource_aws_transit_gateway_attachment.go
```diff
@@ -0,0 +1,319 @@
...
+				ForceNew:    true,
+			},
+			"resource_share_arn": {
+				Description: "The Amazon Resource Name (ARN) of the Resource Share that is needed to grant HCP access to the Transit gateway in AWS.",
+				Type:        schema.TypeString,
+				Required:    true,
+				Sensitive:   true,
+				ForceNew:    true,
+			},
+			"destination_cidrs": {
```
roaks3, line 66: Nit: Our internal implementation used the `destination_cidrs` name that was copied here, but I haven't seen the "destination" qualifier anywhere else. Would `cidrs` be more concise/clear?

internal/provider/resource_aws_transit_gateway_attachment.go
```diff
@@ -0,0 +1,319 @@
...
+	// Wait for TGW attachment creation to complete
+	if err := clients.WaitForOperation(ctx, client, "create Transit gateway attachment", loc, createTGWAttachmentResponse.Payload.Operation.ID); err != nil {
+		return diag.Errorf("unable to create Transit gateway attachment (%s) for HVN (%s) and Transit gateway (%s): %v", tgwAtt.ID, tgwAtt.Hvn.ID, tgwAtt.ProviderData.AwsData.TgwID, err)
+	}
+
+	log.Printf("[INFO] Created Transit gateway attachment (%s) for HVN (%s) and Transit gateway (%s)", tgwAtt.ID, tgwAtt.Hvn.ID, tgwAtt.ProviderData.AwsData.TgwID)
+
+	// Set the globally unique id of this TGW attachment in the state now since
+	// it has been created, and from this point forward should be deletable
+	link := newLink(tgwAtt.Location, TgwAttachmentResourceType, tgwAtt.ID)
```
roaks3, line 194: We will want to move this block up above the `WaitForOperation` call, but holding off on changing this for now because https://github.com/hashicorp/terraform-provider-hcp/pull/59 is still in-progress.

alexmunda: https://github.com/hashicorp/terraform-provider-hcp/pull/59 has been merged. +1 to moving the id setting above the operation polling

docs/data-sources/aws_transit_gateway_attachment.md
bcmdarroch: Iiiinteresting conundrum 🤔 I'm leaning towards #1, unless we can think of any other state the user might want to wait for (maybe PENDING ACCEPTANCE). I've got some questions though.

> It is merged with the computed `state` of the resource
Do you mean terraform a different state that TF is tracking?

Say we go with #1. What happens if the resource never gets set to active? How can we find out its state?

xargs-P: What are the current different states of a TGW connection? 
Also I assume this is different from the state discussion over in: https://github.com/hashicorp/terraform-provider-hcp/pull/59

internal/provider/resource_aws_transit_gateway_attachment.go
bcmdarroch: I like the concise `cidrs`, assuming the AWS provider uses the same. Though I'd defer to @xargs-P

internal/provider/resource_aws_transit_gateway_attachment.go
bcmdarroch: Does that need to happen in any other resource or is this the last one?

internal/provider/resource_hvn.go
```diff
@@ -89,6 +89,11 @@ func resourceHvn() *schema.Resource {
 				Type:        schema.TypeString,
 				Computed:    true,
 			},
+			"provider_account_id": {
```
bcmdarroch, line 92: Let's save this for a follow-up, but I think this is a good point to introduce the pattern for returning cloud-specific outputs.  I'm thinking we'll have an output with the top-level name `aws`, which contains `account_id`, similar to this [vpc_config object](https://github.com/hashicorp/terraform-provider-aws/blob/main/aws/resource_aws_eks_cluster.go#L165-L214) in the the AWS provider's eks_cluster resource. It's a bit of a hack, but it sounds like it's the best we've got while the provider SDK team works out a better solution for input/output objects.

bcmdarroch (approved):
Had some questions and a suggestion for the new `provider_account_id` HVN output, but otherwise nice work! ✨

docs/data-sources/aws_transit_gateway_attachment.md
alexmunda: > Iiiinteresting conundrum 🤔 I'm leaning towards #1, unless we can think of any other state the user might want to wait for (maybe PENDING ACCEPTANCE). I've got some questions though.
> 
> > It is merged with the computed `state` of the resource
> > Do you mean terraform a different state that TF is tracking?
> 
> Say we go with #1. What happens if the resource never gets set to active? How can we find out its state?

Our polling logic should account for any "terminal" state (errored, etc that we can never recover from) and throw an error. 

I prefer the `wait_for_active_state` approach of option 1 as it puts less burden on the user to inspect every possible state. It would also align with doing the same for peering connections as discussed in https://hashicorp.atlassian.net/browse/HCPE-871

internal/provider/resource_aws_transit_gateway_attachment.go
alexmunda: TGW should be the only resource right now that doesn't set the `id` before polling.

internal/clients/tgw.go
```diff
@@ -0,0 +1,56 @@
...
+func WaitForTGWAttachmentState(ctx context.Context, client *Client, tgwAttachmentID string, hvnID string, loc *sharedmodels.HashicorpCloudLocationLocation, state string) (*networkmodels.HashicorpCloudNetwork20200907TGWAttachment, error) {
+	for {
+		tgwAtt, err := GetTGWAttachmentByID(ctx, client, tgwAttachmentID, hvnID, loc)
+		if err != nil {
+			return nil, fmt.Errorf("unable to retrieve Transit gateway attachment (%s): %v", tgwAttachmentID, err)
+		}
+		switch curState := string(tgwAtt.State); curState {
+		case state:
+			return tgwAtt, nil
+		case string(networkmodels.HashicorpCloudNetwork20200907TGWAttachmentStateFAILED):
```
alexmunda, line 42: Are there any other terminal states to be aware of in regards to TGW?

internal/clients/tgw.go
```diff
@@ -0,0 +1,56 @@
...
+
+// WaitForTGWAttachmentState will poll the GET TGW attachment endpoint until
+// it is in the specified state, ctx is canceled, or an error occurs. This
+// is required because AWS can return a newly created TGW attachment before
+// it is ready to be used.
+func WaitForTGWAttachmentState(ctx context.Context, client *Client, tgwAttachmentID string, hvnID string, loc *sharedmodels.HashicorpCloudLocationLocation, state string) (*networkmodels.HashicorpCloudNetwork20200907TGWAttachment, error) {
+	for {
+		tgwAtt, err := GetTGWAttachmentByID(ctx, client, tgwAttachmentID, hvnID, loc)
+		if err != nil {
+			return nil, fmt.Errorf("unable to retrieve Transit gateway attachment (%s): %v", tgwAttachmentID, err)
```
alexmunda, line 37: Does `Gateway` need to be capitalized in these error messages?

internal/provider/data_source_aws_transit_gateway_attachment.go
```diff
@@ -0,0 +1,119 @@
...
+			},
+			"transit_gateway_attachment_id": {
+				Description:      "The ID of the Transit gateway attachment.",
+				Type:             schema.TypeString,
+				Required:         true,
+				ValidateDiagFunc: validateSlugID,
+			},
+			// Optional inputs
+			"state": {
+				Description: "The state of the Transit gateway attachment.",
```
alexmunda, line 36: I think if we keep this approach, we need to list out the possible states that the user can pick and validate them on the input.

internal/provider/resource_aws_transit_gateway_attachment.go
```diff
@@ -0,0 +1,316 @@
...
+				Description: "The ID of the Transit gateway in AWS.",
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+			},
+			"resource_share_arn": {
+				Description: "The Amazon Resource Name (ARN) of the Resource Share that is needed to grant HCP access to the Transit gateway in AWS.",
+				Type:        schema.TypeString,
+				Required:    true,
+				Sensitive:   true,
```
alexmunda, line 60: How come the ARN is sensitive in this case? I usually think of AWS ARNs synonymous to unique IDs, have I been lying to myself this whole time!?

internal/provider/resource_aws_transit_gateway_attachment.go
```diff
@@ -0,0 +1,316 @@
...
+	// Check for an existing HVN
+	_, err := clients.GetHvnByID(ctx, client, loc, hvnID)
+	if err != nil {
+		if clients.IsResponseCodeNotFound(err) {
+			return diag.Errorf("unable to find the HVN (%s) for the Transit gateway attachment", hvnID)
+		}
+
+		return diag.Errorf("unable to check for presence of an existing HVN (%s): %v", hvnID, err)
+	}
+	log.Printf("[INFO] HVN (%s) found, proceeding with create", hvnID)
```
alexmunda, line 141:
```suggestion
	log.Printf("[INFO] HVN (%s) found, proceeding with TGW create", hvnID)
```

docs/data-sources/aws_transit_gateway_attachment.md
roaks3: For reference, possible states for TGW attachment currently are ([ref](https://github.com/hashicorp/cloud-network/blob/master/proto-public/20200907/tgw_attachment.proto#L47-L55)):
```
UNSET
CREATING
PENDING_ACCEPTANCE
ACCEPTED
ACTIVE
FAILED
REJECTED
DELETING
EXPIRED
```

| unless we can think of any other state the user might want to wait for (maybe PENDING ACCEPTANCE)
Yea, it seems like `PENDING_ACCEPTANCE` and `ACTIVE` would be the two states we would be interested in

| Do you mean terraform a different state that TF is tracking?
There's more context in [this link](https://github.com/hashicorp/terraform-provider-google/pull/4797#pullrequestreview-353254440), but I think the concern is that the `state` could be updated from the server during a READ operation, and then you don't really know anymore what state the user is waiting for.

| Say we go with #1. What happens if the resource never gets set to active? How can we find out its state?
Yea, what @alexmunda mentioned sounds like the right solution in that case: store the state of the resource as it gets updated within the operation.

internal/provider/resource_aws_transit_gateway_attachment.go
roaks3: Yep, it looked like Alex got all of the other cases of this in his PR. Will be updating this one to match

internal/provider/resource_hvn.go
roaks3: Yea, I think this would warrant some more discussion, so my preference would be to include that in the general topic around cloud-specific values in our network resources. Definitely agree that this field falls into that category.

internal/clients/tgw.go
roaks3: I'm not 100% sure, but I can check and make sure we're returning once we hit any terminal state. My guess is that we need to include a few more here.

internal/provider/resource_aws_transit_gateway_attachment.go
smaant: this is essentially the same what AWS calls `destination-cidr-block` https://docs.aws.amazon.com/cli/latest/reference/ec2/create-route.html. Relatively soon we will introduce a concept of HVN Routes that would have destination-cidr (ip range) and destination (peering, tgw attachment, etc). I guess we can call it just `cidr`. In the peering, for example, it's called `peer_vpc_cidr_block`.

internal/clients/tgw.go
roaks3: I'm honestly not sure here. Our capitalization guide from design doesn't seem to explicitly call out a case like this: https://docs.google.com/document/d/1ABNkYIp6aLwPYNpuYOFcDFBFCcnH_4ePRaLP-gD_tKE/edit

@joshklekamp Do you have an opinion here on how we capitalize `Transit gateway attachment` within our error messages and logs?

smaant (commented):
A few comments

internal/provider/data_source_aws_transit_gateway_attachment.go
roaks3: Yea, I was waiting to see where we landed with the interface, but will make sure to address documentation, validation, and diffing once we make a choice.

internal/provider/resource_aws_transit_gateway_attachment.go
smaant: "TGW attachment"

internal/provider/resource_aws_transit_gateway_attachment.go
roaks3: Hmm, this came from @smaant , but I haven't actually verified. Anton, do you know if this is a sensitive value?

internal/clients/tgw.go
joshklekamp: It should be sentence case unless we add AWS to the beginning.

joshklekamp: or Amazon 😄

docs/resources/aws_transit_gateway_attachment.md
```diff
@@ -0,0 +1,102 @@
...
+```
+
+## Schema
+
+### Required
+
+- **destination_cidrs** (List of String) The list of associated CIDR ranges.
+- **hvn_id** (String) The ID of the HashiCorp Virtual Network (HVN).
+- **resource_share_arn** (String, Sensitive) The Amazon Resource Name (ARN) of the Resource Share that is needed to grant HCP access to the Transit gateway in AWS.
+- **transit_gateway_attachment_id** (String) The ID of the Transit gateway attachment.
```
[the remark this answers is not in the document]
nombiezinja: Perhaps it makes sense to be explicit about it and say "The user-settable name of the Transit Gateway attachment in HCP"

internal/provider/resource_aws_transit_gateway_attachment.go
smaant: in general ARNs aren't sensitive but we kind of treat them as such. The idea is that Resource Share (RS) is shared with the entire HVN's org AWS account, therefore technically any HVN in this org can use this RS to attach to TGW shared in this RS as long as they know TGW ID and RS ARN therefore we need to keep one them as a secret.

internal/clients/tgw.go
smaant: AWS itself calls it "Transit Gateway"

internal/clients/tgw.go
smaant: technically there're also `REJECTED` and `EXPIRED`, we might add it here too

internal/provider/resource_aws_transit_gateway_attachment.go
```diff
@@ -0,0 +1,316 @@
...
+	"github.com/hashicorp/terraform-provider-hcp/internal/clients"
+)
+
+var tgwDefaultTimeout = time.Minute * 1
+var tgwCreateTimeout = time.Minute * 35
+var tgwDeleteTimeout = time.Minute * 35
+
+func resourceAwsTransitGatewayAttachment() *schema.Resource {
+	return &schema.Resource{
+		Description: "The AWS Transit gateway attachment resource allows you to manage a Transit gateway attachment that attaches an HVN to a Transit gateway in AWS.",
```
nombiezinja, line 23:
```suggestion
		Description: "The AWS Transit gateway attachment resource allows you to manage a Transit Gateway attachment that attaches an HVN to a Transit Gateway you own in AWS.",
```

internal/clients/tgw.go
nombiezinja: Other terminal states in addition to `FAILED` include `REJECTED`, `DELETING`, `EXPIRED`. 
`UNSET` being returned should also indicate something very wrong as well, however this should never happen, we are probably ok not handling this case here.

internal/provider/resource_aws_transit_gateway_attachment.go
```diff
@@ -0,0 +1,316 @@
...
+	// Check if TGW attachment already exists
+	_, err = clients.GetTGWAttachmentByID(ctx, client, tgwAttachmentID, hvnID, loc)
+	if err != nil {
+		if !clients.IsResponseCodeNotFound(err) {
+			return diag.Errorf("unable to check for presence of an existing Transit gateway attachment (%s): %v", tgwAttachmentID, err)
+		}
+
+		log.Printf("[INFO] Transit gateway attachment (%s) not found, proceeding with create", tgwAttachmentID)
+	} else {
+		return diag.Errorf("a Transit gateway attachment with transit_gateway_attachment_id=%s, hvn_id=%s and project_id=%s already exists - to be managed via Terraform this resource needs to be imported into the state. Please see the resource documentation for hcp_transit_gateway_attachment for more information", tgwAttachmentID, hvnID, loc.ProjectID)
```
nombiezinja, line 152: Hmm not sure if this is the pattern we follow everywhere so please ignore this comment if this is a custom thing, but [a check ensuring unique tgwAttachmentID within a project is in place](https://github.com/hashicorp/cloud-network/blob/master/service/network/private/service_create_tgw_attachment.go#L341), and would be part of the output in the subsequent `Create` request.

docs/data-sources/aws_transit_gateway_attachment.md
xargs-P: Thanks for the comments everyone. This is a good thread. I agree with the teams viewpoint on the 1st item.

docs/resources/aws_transit_gateway_attachment.md
```diff
@@ -0,0 +1,102 @@
+---
+page_title: "hcp_aws_transit_gateway_attachment Resource - terraform-provider-hcp"
+subcategory: ""
+description: |-
+  The AWS Transit gateway attachment resource allows you to manage a Transit gateway attachment that attaches an HVN to a Transit gateway in AWS.
+---
+
+# Resource `hcp_aws_transit_gateway_attachment`
+
+The AWS Transit gateway attachment resource allows you to manage a Transit gateway attachment that attaches an HVN to a Transit gateway in AWS.
```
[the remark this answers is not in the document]
xargs-P: @smaant 
>  to specify that a Transit Gateway is owned by a customer (located in their account) 
++ 
> should be located in the same AWS region as the HVN 
I don't think we have any technically limitation stopping users from having the connection cross regions. How about we change the text to "....and _for a performant connection_ should be located in the same AWS region as the HVN" ?

docs/resources/aws_transit_gateway_attachment.md
```diff
@@ -0,0 +1,102 @@
...
+- **destination_cidrs** (List of String) The list of associated CIDR ranges.
+- **hvn_id** (String) The ID of the HashiCorp Virtual Network (HVN).
+- **resource_share_arn** (String, Sensitive) The Amazon Resource Name (ARN) of the Resource Share that is needed to grant HCP access to the Transit gateway in AWS.
+- **transit_gateway_attachment_id** (String) The ID of the Transit gateway attachment.
+- **transit_gateway_id** (String) The ID of the Transit gateway in AWS.
+
+### Optional
+
+- **id** (String) The ID of this resource.
+- **timeouts** (Block, Optional) (see [below for nested schema](#nestedblock--timeouts))
```
xargs-P, line 82: 1. Do we have any other resources that have nested timeouts? 🤔  
2. Can we add this to the example above?

docs/data-sources/aws_transit_gateway_attachment.md
smaant: In case of going with `wait_for_active_state` approach I would recommend deleting `state` input, otherwise you'll have to deal with edge-cases when `wait_for_active_state = true` and `state` is set to something different.

docs/resources/aws_transit_gateway_attachment.md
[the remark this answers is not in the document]
smaant: > I don't think we have any technically limitation stopping users from having the connection cross regions.

@xargs-P we do, AWS doesn't allow cross region TGW VPC attachments. If customer wants to have their TGW and HVN in different regions they will need to create a new TGW in the same region as HVN and peer these TGWs. But TGW attachment from HVN have to be to a TGW in the same region.

docs/data-sources/aws_transit_gateway_attachment.md
roaks3: Agreed, thanks for the comments! I'll proceed with a `wait_for_active_state` flag and change `state` to be read-only.

docs/data-sources/aws_transit_gateway_attachment.md
```diff
@@ -0,0 +1,51 @@
...
+## Schema
+
+### Required
+
+- **hvn_id** (String) The ID of the HashiCorp Virtual Network (HVN).
+- **transit_gateway_attachment_id** (String) The ID of the Transit gateway attachment.
+
+### Optional
+
+- **id** (String) The ID of this resource.
```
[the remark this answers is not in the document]
roaks3: So this `id` represents our unique Terraform id stored in the state (format is `/project/123/hashicorp.network.tgw-attachment/456`), but the description here is not particularly descriptive, and I think it might actually be confusing for users. Note that it exists here because it is autogenerated by `go generate`.

I think we should leave it for now, because it aligns with the generator and our other resources, but I've created a ticket to see if it makes sense to modify our docs generation to not include these: https://hashicorp.atlassian.net/browse/HCPE-883

docs/resources/aws_transit_gateway_attachment.md
roaks3: These are actually autogenerated by `go generate`, and allow the user to specify custom timeouts for their operations. We do have these on all of our other resources/data sources currently, so the changes here align with that.

We could include these in the example code, but I think it could be a bit distracting for a user? I would expect a user to only need to set these timeouts in very rare situations, if at all. In the case that these are not useful, we may want to consider just removing them from the documentation (similar to the `id` https://hashicorp.atlassian.net/browse/HCPE-883)

docs/resources/aws_transit_gateway_attachment.md
```diff
@@ -0,0 +1,102 @@
...
+- **project_id** (String) The ID of the HCP project where the Transit gateway attachment is located. Always matches the HVN's project.
+- **provider_transit_gateway_attachment_id** (String) The Transit gateway attachment ID used by AWS.
+- **state** (String) The state of the Transit gateway attachment.
+
+<a id="nestedblock--timeouts"></a>
+### Nested Schema for `timeouts`
+
+Optional:
+
+- **create** (String)
```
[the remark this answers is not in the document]
roaks3: These fields, and the `default` field in the data source, are meant for setting custom timeouts for the operations. They render [like this](https://registry.terraform.io/providers/hashicorp/hcp/latest/docs/resources/hvn#nested-schema-for-timeouts) on the docs site, and are currently documented this way for all of our resources due to `go generate`.

Similar to my comment on the `id`, I think we should leave these they way they are for consistency, but make some sort of adjustment moving forward so that they are more clear to users. Possible options include removing these blocks from the docs, or adding actual descriptions to explain what these fields do.

internal/provider/resource_aws_transit_gateway_attachment.go
roaks3: Ah yea, this came up with the HVN resource, and we ended up with the pattern you see here where we do a GET first: https://github.com/hashicorp/terraform-provider-hcp/pull/3#discussion_r556234766

We might be able to accomplish the same thing with the CREATE call, but I think we would need to be careful about what guarantees we have from the endpoint (eg. it looks like we can get at least two types of `AlreadyExists` response codes), and the main driver here was being able to supply an actionable error message.

nombiezinja: Thanks for explaining, this sounds good to me, resolving this comment

internal/clients/tgw.go
roaks3: Thanks for the additional context here (the diagram was really helpful too). I've updated the function to actually specify the "pending" states, instead of terminal states, to align with [the helper that the provider SDK offers](https://github.com/hashicorp/terraform-plugin-sdk/blob/master/helper/resource/state.go). This helper is used frequently in other providers, and seems to be a safer approach if we introduce new states in the future.

docs/data-sources/aws_transit_gateway_attachment.md
```diff
@@ -0,0 +1,51 @@
...
+### Read-only
+
+- **created_at** (String) The time that the Transit gateway attachment was created.
+- **destination_cidrs** (List of String) The list of associated CIDR ranges. Traffic from these CIDRs will be allowed for all resources in the HVN. Traffic to these CIDRs will be routed into this Transit gateway attachment.
+- **expires_at** (String) The time after which the Transit gateway attachment will be considered expired if it hasn't transitioned into `ACCEPTED` or `ACTIVE` state.
+- **organization_id** (String) The ID of the HCP organization where the Transit gateway attachment is located. Always matches the HVN's organization.
+- **project_id** (String) The ID of the HCP project where the Transit gateway attachment is located. Always matches the HVN's project.
+- **provider_transit_gateway_attachment_id** (String) The Transit gateway attachment ID used by AWS.
+- **transit_gateway_id** (String) The ID of the Transit gateway in AWS.
+
```
smaant, line 43: `state` is missing

internal/clients/tgw.go
roaks3: Verified with Josh that we should use sentence case here (so `Transit gateway attachment` at the beginning of a sentence, but `transit gateway attachment` in the middle of a sentence). There are two exceptions where we refer to the resource/data source by the full name in their top-level description, and in those cases we use `AWS Transit Gateway Attachment`. I've updated the error messages and descriptions to reflect the correct capitalization.

docs/data-sources/aws_transit_gateway_attachment.md
```diff
@@ -0,0 +1,51 @@
+---
+page_title: "hcp_aws_transit_gateway_attachment Data Source - terraform-provider-hcp"
+subcategory: ""
+description: |-
+  The AWS Transit gateway attachment data source provides information about an existing Transit gateway attachment.
```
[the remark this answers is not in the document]
roaks3: Resolving based on conversation here: https://github.com/hashicorp/terraform-provider-hcp/pull/58#discussion_r574561134

docs/resources/aws_transit_gateway_attachment.md
roaks3: Created https://hashicorp.atlassian.net/browse/HCPE-890 to address these `timeouts` fields across all resources

internal/provider/resource_aws_transit_gateway_attachment.go
roaks3: Given @smaant 's response, I'm going to leave this as `destination_cidrs`. Also note that the UI field is labeled "Destinations", so I think this aligns well with that.

smaant (approved):

smaant:
Great job @roaks3, thank you for handling this!

nombiezinja (approved):
Awesome work, appreciate you taking this on!

PR #60, opened 2021-02-10
add README example for using local binary

bcmdarroch:
Ran into an issue where I needed to upgrade to v0.14 to use my handcrafted organic locally built provider, so I thought I'd add this caveat to the README.


roaks3 (approved):
👍  This looks good, but I did want to note that I have some changes coming around this same block as part of the contribution documentation

PR #63, opened 2021-02-11
HCS-1760: Add Consul versions data source

alexmunda:
Based off the [`hcs_consul_versions` data source](https://registry.terraform.io/providers/hashicorp/hcs/latest/docs/data-sources/consul_versions), we can allows users to explicitly set their Consul cluster version to the latest recommended Consul version that HCP supports. This means that any time the recommended HCP Consul version changes, someone utilizing the `recommended` field of the `consul_versions` data source will be prompted to upgrade their cluster the next time they run `terraform plan`:
```
❯ terraform plan
hcp_hvn.example_hvn: Refreshing state... 
hcp_consul_cluster.example_consul_cluster: Refreshing state... 

An execution plan has been generated and is shown below.
  ~ update in-place

  # hcp_consul_cluster.example_consul_cluster will be updated in-place
  ~ resource "hcp_consul_cluster" "example_consul_cluster" {
        id                            = "/project/XXX/hashicorp.consul.cluster/hcp-tf-example-consul-cluster"
      ~ min_consul_version            = "v1.9.2" -> "v1.9.3"
        # (19 unchanged attributes hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```

This data source also informs users of the Consul versions that HCP supports, in the case where they are looking to use a specific Consul version.

If HCP offers preview versions of Consul in the future, users would be able to discover and set `min_consul_version` of their cluster using this data source.

```
❯ tf show
# data.hcp_consul_versions.versions:
data "hcp_consul_versions" "versions" {
    available   = [
        "v1.9.3",
        "v1.9.2",
        "v1.9.1",
        "v1.9.0",
        "v1.8.8",
        "v1.8.7",
        "v1.8.6",
        "v1.8.4",
        "v1.8.0",
    ]
    id          = "recommended/v1.9.3/available_len/9/preview_len/0"
    preview     = []
    recommended = "v1.9.3"
}
```

Usage:
```
data "hcp_consul_versions" "versions" {}

resource "hcp_consul_cluster" "example_consul_cluster" {
  hvn_id             = hcp_hvn.example_hvn.hvn_id
  cluster_id         = "hcp-tf-example-consul-cluster"
  tier               = "development"
  min_consul_version = data.hcp_consul_versions.versions.recommended
}
```

### DIFF
--- a/internal/clients/consul_cluster.go
+++ b/internal/clients/consul_cluster.go
@@ -112,14 +112,29 @@ func CreateConsulCluster(ctx context.Context, client *Client, loc *sharedmodels.
 	return resp.Payload, nil
 }
 
-// GetAvailableHCPConsulVersions gets the list of available Consul versions that HCP supports.
-func GetAvailableHCPConsulVersions(ctx context.Context, loc *sharedmodels.HashicorpCloudLocationLocation, client *Client) ([]*consulmodels.HashicorpCloudConsul20200826Version, error) {
+// GetAvailableHCPConsulVersionsForLocation gets the list of available Consul versions that HCP supports for
+// the provided location.
+func GetAvailableHCPConsulVersionsForLocation(ctx context.Context, loc *sharedmodels.HashicorpCloudLocationLocation, client *Client) ([]*consulmodels.HashicorpCloudConsul20200826Version, error) {
 	p := consul_service.NewListVersionsParams()
 	p.Context = ctx
 	p.LocationProjectID = loc.ProjectID
 	p.LocationOrganizationID = loc.OrganizationID
 
 	resp, err := client.Consul.ListVersions(p, nil)
+
+	if err != nil {
+		return nil, err
+	}
+
+	return resp.Payload.Versions, nil
+}
+
+// GetAvailableHCPConsulVersions gets the list of available Consul versions that HCP supports.
+func GetAvailableHCPConsulVersions(ctx context.Context, client *Client) ([]*consulmodels.HashicorpCloudConsul20200826Version, error) {
+	p := consul_service.NewListVersions2Params()
+	p.Context = ctx
+
+	resp, err := client.Consul.ListVersions2(p, nil)
 
 	if err != nil {
 		return nil, err
@@ -147,7 +162,7 @@ func DeleteConsulCluster(ctx context.Context, client *Client, loc *sharedmodels.
 	return deleteResp.Payload, nil
 }
 
-// GetAvailableHCPConsulVersions gets the list of available Consul versions that HCP supports.
+// GetAvailableHCPConsulVersionsForLocation gets the list of available Consul versions that HCP supports.
 func ListConsulUpgradeVersions(ctx context.Context, client *Client, loc *sharedmodels.HashicorpCloudLocationLocation,
 	clusterID string) ([]*consulmodels.HashicorpCloudConsul20200826Version, error) {
 
--- a/internal/clients/consul_cluster.go
+++ b/internal/clients/consul_cluster.go
@@ -162,7 +162,8 @@ func DeleteConsulCluster(ctx context.Context, client *Client, loc *sharedmodels.
 	return deleteResp.Payload, nil
 }
 
-// GetAvailableHCPConsulVersionsForLocation gets the list of available Consul versions that HCP supports.
+// ListConsulUpgradeVersions gets the list of available Consul versions that the supplied cluster
+// can upgrade to.
 func ListConsulUpgradeVersions(ctx context.Context, client *Client, loc *sharedmodels.HashicorpCloudLocationLocation,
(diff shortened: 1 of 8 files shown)


internal/clients/consul_cluster.go
```diff
@@ -147,7 +162,7 @@ func DeleteConsulCluster(ctx context.Context, client *Client, loc *sharedmodels.
 	return deleteResp.Payload, nil
 }
 
-// GetAvailableHCPConsulVersions gets the list of available Consul versions that HCP supports.
+// GetAvailableHCPConsulVersionsForLocation gets the list of available Consul versions that HCP supports.
 func ListConsulUpgradeVersions(ctx context.Context, client *Client, loc *sharedmodels.HashicorpCloudLocationLocation,
```
mkeeler, line 166: Was this supposed to be renamed to match the doc comment?

mkeeler (commented):
Two things.

First is should we expose preview versions in the provider before we support them in HCP itself? If we wanted to call them something other than preview in the future once we went down the path of implementing them then it would potentially be a breaking change to the provider.

Secondly the doc comment I pointed out doesn't match the func name. However the name in the comment is referenced from new code so I am not too sure whats going on here.

internal/clients/consul_cluster.go
alexmunda: Yes great catch!

alexmunda:
> First is should we expose preview versions in the provider before we support them in HCP itself? If we wanted to call them something other than preview in the future once we went down the path of implementing them then it would potentially be a breaking change to the provider.

Great point @mkeeler. Fortunately this implementation is tied to the [`VersionStatus` enum](https://github.com/hashicorp/hcp-sdk-go/blob/09ae52918f56d49c6e786f672326d81d3ad8e0a4/clients/cloud-consul-service/preview/2020-08-26/models/hashicorp_cloud_consul20200826_version_status.go#L34) in the `2020-08-26` version of the consul-service API. If down the road that enum is changed, it would require a new consul-service API version to be created, the `2020-08-26` version would still function as expected, and we could safely update the data source implementation to the new API version's enum without breaking existing functionality.

internal/provider/data_source_consul_versions.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	err = d.Set("preview", previewVersions)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	d.SetId(fmt.Sprintf("recommended/%s/available_len/%d/preview_len/%d", recommendedVersion, len(availableVersions), len(previewVersions)))
```
roaks3, line 93: I don't think this would be blocking, but I'm curious if there are long-term plans for the id here. The intention seems to be to capture the version config at a specific point in time, but I think this format is technically vulnerable to things like 1 version added and 1 version removed, resulting in the same id. FWIW, I noticed AWS handles a similar use case by including a `syncToken` in their [ip-ranges](https://docs.aws.amazon.com/general/latest/gr/aws-ip-ranges.html) response.

roaks3 (approved):
LGTM, my comment on the id used here should not be blocking

internal/provider/data_source_consul_versions.go
alexmunda: Good idea, I will go with a hash of the versions concated.

alexmunda:
```
data "hcp_consul_versions" "versions" {
    available   = [
        "v1.9.3",
        "v1.9.2",
        "v1.9.1",
        "v1.9.0",
        "v1.8.8",
        "v1.8.7",
        "v1.8.6",
        "v1.8.4",
        "v1.8.0",
    ]
    id          = "16fe24158cef6090063cae717b555142"
    preview     = []
    recommended = "v1.9.3"
}
```

internal/consul/version.go
```diff
@@ -37,3 +38,33 @@ func IsValidVersion(version string, versions []*consulmodels.HashicorpCloudConsu
...
+	if len(otherVersions) == 0 {
+		return recommendedVersion
+	}
+
+	// No recommended found, return others
+	if recommendedVersion == "" {
+		return strings.Join(otherVersions, ", ")
+	}
+
+	return fmt.Sprintf("%s, %s", recommendedVersion, strings.Join(otherVersions, ", "))
```
roaks3, line 69: Would it be weird to prepend this string with something like `recommended: ` (and same for line 61)? It might be overkill, but I think there's an edge case where eg. `v1.9.0, v1.8.6, v1.8.4` would be produced by two configurations: no recommended version, and `v1.9.0` recommended version.

roaks3 (approved):
LGTM, just had one comment that might make the ids more robust

internal/consul/version.go
alexmunda: I can definitely get behind adding it for this line. On 61 since there would only be 1 version returned (the recommended one) it seems like overkill 🤷

PR #64, opened 2021-02-12
HCPE-909 - Add TGW attachment imports

roaks3:
Adds import functionality, where the import id is `{hvn_id}:{transit_gateway_attachment_id}`

Most of the functionality matches what is already done with network peering.

One note: because `resource_share_arn` is not available in our responses, it cannot be populated into Terraform state from an import. This is a known case that can be handled with an `ignore_changes` block (I've included this in the documentation here), but the preferred solution would be to include this value in the response, if that is an option here.

### DIFF
--- a/internal/provider/resource_aws_transit_gateway_attachment.go
+++ b/internal/provider/resource_aws_transit_gateway_attachment.go
@@ -2,7 +2,9 @@ package provider
 
 import (
 	"context"
+	"fmt"
 	"log"
+	"strings"
 	"time"
 
 	"github.com/hashicorp/hcp-sdk-go/clients/cloud-network/preview/2020-09-07/client/network_service"
@@ -29,6 +31,9 @@ func resourceAwsTransitGatewayAttachment() *schema.Resource {
 			Default: &tgwDefaultTimeout,
 			Create:  &tgwCreateTimeout,
 			Delete:  &tgwDeleteTimeout,
+		},
+		Importer: &schema.ResourceImporter{
+			StateContext: resourceAwsTransitGatewayAttachmentImport,
 		},
 
 		Schema: map[string]*schema.Schema{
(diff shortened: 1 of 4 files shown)


internal/provider/resource_aws_transit_gateway_attachment.go
```diff
@@ -322,3 +327,33 @@ func setTransitGatewayAttachmentResourceData(d *schema.ResourceData, tgwAtt *net
...
+	client := meta.(*clients.Client)
+
+	idParts := strings.SplitN(d.Id(), ":", 2)
+	if len(idParts) != 2 || idParts[0] == "" || idParts[1] == "" {
+		return nil, fmt.Errorf("unexpected format of ID (%q), expected {hvn_id}:{transit_gateway_attachment_id}", d.Id())
+	}
+	hvnID := idParts[0]
+	tgwAttID := idParts[1]
+	loc := &sharedmodels.HashicorpCloudLocationLocation{
+		ProjectID: client.Config.ProjectID,
```
smaant, line 344: aren't we also need an org-id here?

smaant (approved):
one question, otherwise LGTM

internal/provider/resource_aws_transit_gateway_attachment.go
roaks3: This location is used to create the link format we use for ids (`/project/123/hashicorp.network.tgw-attachment/456`), so we only actually need the project id. The link format used to include more information, but since it is now only the project id that gets used, I'll make a note to consider simplifying this across our resources 🙏 .

roaks3: Ticket created: https://hashicorp.atlassian.net/browse/HCPE-915

PR #65, opened 2021-02-12
Add beta banner to docs for TGW attachment resources

roaks3:
This adds a banner to the docs for transit gateway attachment resources, since they are still in private beta and behind a feature flag. When the next version of the provider is published, these pages will be visible to all users, so this banner will help prevent confusion.


smaant (approved):
👍

xargs-P (approved):

PR #68, opened 2021-02-17
HCS-1716: Add HCP Consul federation support

riddhi89:
Adds federation support for HCP Consul.

### DIFF
--- a/docs/guides/consul-federation.md
+++ b/docs/guides/consul-federation.md
@@ -7,7 +7,7 @@ description: |-
 
 # Federate a new HCP Consul cluster with an existing one
 
-Once you have a HCP Consul cluster, you can create new Consul cluster to federate with the existing one.
+Once you have a HCP Consul cluster, you can create a new Consul cluster to federate with the existing one.
 
 ```terraform
 resource "hcp_hvn" "example" {
--- a/docs/guides/consul-federation.md
+++ b/docs/guides/consul-federation.md
@@ -1,0 +1,31 @@
+---
+subcategory: ""
+page_title: "Federate HCP Consul clusters - HCP Provider"
+description: |-
+    An example of federating a new HCP Consul cluster with an existing one.
+---
+
+# Federate a new HCP Consul cluster with an existing one
+
+Once you have a HCP Consul cluster, you can create new Consul cluster to federate with the existing one.
+
+```terraform
+resource "hcp_hvn" "example" {
+  hvn_id         = var.hvn_id
+  cloud_provider = var.cloud_provider
+  region         = var.region
+}
+
+resource "hcp_consul_cluster" "primary" {
+  hvn_id     = hcp_hvn.example.hvn_id
+  cluster_id = var.primary_cluster_id
+  tier       = "development"
+}
+
+resource "hcp_consul_cluster" "secondary" {
+  hvn_id       = hcp_hvn.example.hvn_id
+  cluster_id   = var.secondary_cluster_id
+  tier         = "development"
+  primary_link = hcp_consul_cluster.primary.self_link
+}
+```
--- a/examples/resources/hcp_consul_cluster/federation.tf
+++ b/examples/resources/hcp_consul_cluster/federation.tf
@@ -1,0 +1,15 @@
+data "hcp_hvn" "example" {
+  hvn_id = "hvn"
+}
+
+data "hcp_consul_cluster" "primary" {
+  cluster_id = "consul-cluster-primary"
+}
+
+resource "hcp_consul_cluster" "secondary" {
+  hvn_id       = data.hcp_hvn.example.hvn_id
+  cluster_id   = "consul-cluster-secondary"
+  tier         = "development"
+  primary_link = data.hcp_consul_cluster.primary.self_link
+}
+
--- a/internal/clients/consul_cluster.go
+++ b/internal/clients/consul_cluster.go
@@ -132,10 +132,10 @@ func GetAvailableHCPConsulVersions(ctx context.Context, loc *sharedmodels.Hashic
 
 // GetAvailableHCPConsulVersions gets the list of available Consul versions that HCP supports.
 func GetAvailableHCPConsulVersions(ctx context.Context, client *Client) ([]*consulmodels.HashicorpCloudConsul20200826Version, error) {
-	p := consul_service.NewListVersions2Params()
+	p := consul_service.NewConsulServiceListVersions2Params()
 	p.Context = ctx
 
-	resp, err := client.Consul.ListVersions2(p, nil)
+	resp, err := client.Consul.ConsulServiceListVersions2(p, nil)
 
 	if err != nil {
 		return nil, err
--- a/internal/clients/consul_cluster.go
+++ b/internal/clients/consul_cluster.go
@@ -12,13 +12,13 @@ import (
 func GetConsulClusterByID(ctx context.Context, client *Client, loc *sharedmodels.HashicorpCloudLocationLocation,
 	consulClusterID string) (*consulmodels.HashicorpCloudConsul20200826Cluster, error) {
 
-	getParams := consul_service.NewGetParams()
+	getParams := consul_service.NewConsulServiceGetParams()
 	getParams.Context = ctx
 	getParams.ID = consulClusterID
 	getParams.LocationOrganizationID = loc.OrganizationID
 	getParams.LocationProjectID = loc.ProjectID
 
-	getResp, err := client.Consul.Get(getParams, nil)
+	getResp, err := client.Consul.ConsulServiceGet(getParams, nil)
 	if err != nil {
 		return nil, err
 	}
@@ -33,13 +33,13 @@ func GetConsulClusterByID(ctx context.Context, client *Client, loc *sharedmodels
 func GetConsulClientConfigFiles(ctx context.Context, client *Client, loc *sharedmodels.HashicorpCloudLocationLocation,
 	consulClusterID string) (*consulmodels.HashicorpCloudConsul20200826GetClientConfigResponse, error) {
 
-	p := consul_service.NewGetClientConfigParams()
+	p := consul_service.NewConsulServiceGetClientConfigParams()
 	p.Context = ctx
 	p.ID = consulClusterID
 	p.LocationOrganizationID = loc.OrganizationID
 	p.LocationProjectID = loc.ProjectID
 
-	resp, err := client.Consul.GetClientConfig(p, nil)
+	resp, err := client.Consul.ConsulServiceGetClientConfig(p, nil)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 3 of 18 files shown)


mkeeler (commented):
If you don't have federation enabled and attempt to use it from the TF provider how does this present itself to the user?

riddhi89:
@mkeeler : The user would get an error like:

```
hcp_consul_cluster.secondary: Creating...
Error: unable to create Consul cluster (consul-cluster-secondary): [POST /consul/2020-08-26/organizations/{cluster.location.organization_id}/projects/{cluster.location.project_id}/clusters][400] ConsulService_Create default  &{Code:3 Details:[] Error:federation is unavailable for this cluster Message:federation is unavailable for this cluster}
```

i0rek (commented):
This is exciting. Which changes were generated and which are the ones you did? Mostly asking because that would make it a little easier to review.

internal/clients/response.go
```diff
@@ -11,5 +13,9 @@ import (
 // request, and returns true if the response code was 404 not found
 func IsResponseCodeNotFound(err error) bool {
 	var apiErr *runtime.APIError
-	return errors.As(err, &apiErr) && apiErr.Code == http.StatusNotFound
+	if errors.As(err, &apiErr) {
+		return apiErr.Code == http.StatusNotFound
+	} else {
+		return strings.Contains(err.Error(), fmt.Sprintf("[%d]", http.StatusNotFound))
```
roaks3, line 19: This is interesting, did you come across a case that required this condition? I would be slightly wary about changing the behavior of this function because a lot of resources use it to determine if they exist or not. Another wrinkle is that the format of response error messages may change with the work @bcmdarroch is looking at.

internal/provider/resource_consul_cluster.go
```diff
@@ -261,6 +272,29 @@ func resourceConsulClusterCreate(ctx context.Context, d *schema.ResourceData, me
 		return diag.Errorf("specified Consul version (%s) is unavailable; must be one of: %v", consulVersion, availableConsulVersions)
 	}
 
+	// If specified, validate and parse the primary link provided for federation.
+	primary_link, ok := d.GetOk("primary_link")
+	var primary *sharedmodels.HashicorpCloudLocationLink
+	if ok {
+		primary, err = parseLinkURL(primary_link.(string), ConsulClusterResourceType)
```
roaks3, line 279: You might be able to replace some of this with `buildLinkFromURL()`, eg:
```
buildLinkFromURL(primary_link.(string), ConsulClusterResourceType, client.Config.OrganizationID)
```

roaks3 (commented):
🎉  Code LGTM, just had two comments

xargs-P:
Hi @riddhi89  , things are looking good. 👌

Can you add a /examples/guides/ markdown for federation? And also update the example in docs/index.md ? Customers will be very interested in this functionality. So it would be sweet to give them as much help as possible.

internal/clients/response.go
bcmdarroch: So this is a bit of a bandaid until I get the rest of service SDKs wired up with errors. Once https://github.com/hashicorp/hcp-sdk-go/pull/7 merges, I'd like to bump the SDK version, then upgrade the provider and revisit this helper then.

internal/provider/resource_consul_cluster.go
```diff
@@ -261,6 +272,29 @@ func resourceConsulClusterCreate(ctx context.Context, d *schema.ResourceData, me
...
+			}
+			primary.Location.OrganizationID = primaryOrgID
+		}
+		// fetch the primary cluster
+		primaryConsulCluster, err := clients.GetConsulClusterByID(ctx, client, primary.Location, primary.ID)
+		if err != nil {
+			return diag.Errorf("unable to check for presence of an existing primary Consul cluster (%s): %v", primary.ID, err)
+		}
+		primary.Location.Region = primaryConsulCluster.Location.Region
+	}
```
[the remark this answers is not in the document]
bcmdarroch: 💭  The provider does some basic validation, like input format and verifying the existence of dependent resources, but I'd agree that this seems closer to Consul business logic. That being said, I could see the advantage of having the request fail sooner, giving the user quicker feedback. How much later down the line would the consul service check for the primary?

internal/provider/resource_consul_cluster.go
```diff
@@ -417,6 +451,22 @@ func setConsulClusterResourceData(d *schema.ResourceData, cluster *consulmodels.
 		return err
 	}
 
+	// self_link is equal to the terraform ID of the resource
+	if err := d.Set("self_link", d.Id()); err != nil {
```
bcmdarroch, line 455: Why is the attribute `self_link` needed if it's the same as the terraform ID?

bcmdarroch (commented):
This looks great! There's a few open questions I'm curious about: the necessity of `self_link` and whether or not we should push the verification of the primary cluster's existence into consul service.

internal/clients/consul_cluster.go
```diff
@@ -75,9 +75,9 @@ func CreateCustomerRootACLToken(ctx context.Context, client *Client, loc *shared
 // CreateConsulCluster will make a call to the Consul service to initiate the create Consul
 // cluster workflow.
 func CreateConsulCluster(ctx context.Context, client *Client, loc *sharedmodels.HashicorpCloudLocationLocation,
-	clusterID, datacenter, consulVersion string, numServers int32, private, connectEnabled bool, network *sharedmodels.HashicorpCloudLocationLink) (*consulmodels.HashicorpCloudConsul20200826CreateResponse, error) {
+	clusterID, datacenter, consulVersion string, numServers int32, private, connectEnabled bool, network *sharedmodels.HashicorpCloudLocationLink, primary *sharedmodels.HashicorpCloudLocationLink) (*consulmodels.HashicorpCloudConsul20200826CreateResponse, error) {
```
alexmunda, line 78: There are a lot of params to this function and the two links at the end make me think it would be easy to flip the hvn and primary in the future. What do you think about it taking in a single `consulmodels.HashicorpCloudConsul20200826Cluster` struct?

internal/provider/resource_consul_cluster.go
alexmunda: The reason it's a separate field is to be explicit to the user that it is a separate property so they aren't confused between, `id` and `cluster_id`. It also protects against the Terraform id format being changed in the future (not likely but 🤷 ). 

That being said, I think it might be best to "re-build" the link using the `newLink` and `linkUrl` functions just so we aren't relying on the `id`.

internal/clients/consul_cluster.go
riddhi89: Sounds like a good improvement to me. Will update.

internal/provider/resource_consul_cluster.go
riddhi89: @roaks3 : So the reason I am not using the `buildLinkFromURL` helper here is because there could be a scenario (planned for future support), where the primary could belong to a different organization than the default org ID (which is client.Config.OrganizationID). Implementing it this way, handles both cases.

internal/provider/resource_consul_cluster.go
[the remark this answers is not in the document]
riddhi89: Let me verify if this is done in the consul-service. If not, definitely sounds like a good place to include it there vs here.

internal/provider/resource_consul_cluster.go
riddhi89: >might be best to "re-build" the link

Makes sense. Will do.

internal/provider/resource_consul_cluster.go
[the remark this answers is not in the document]
riddhi89: Nevermind. The reason I have this check in here is because I need the primary's region config (primary.Location.Region). Else, the below error is raised:
```hcp_consul_cluster.secondary: Creating...

Error: unable to create Consul cluster (consul-cluster-secondary): [POST /consul/2020-08-26/organizations/{cluster.location.organization_id}/projects/{cluster.location.project_id}/clusters][400] ConsulService_Create default  &{Code:3 Details:[0xc000151e60] Error:config: (consul_config: (primary: (location: (region: cannot be blank.).).).). Message:config: (consul_config: (primary: (location: (region: cannot be blank.).).).).}```

File: internal/clients/response.go (hunk 1 above)
Comment (riddhi89) [reply internal/clients/response.go:19]: Right, the "not found consul cluster" errors were not being marshalled into an apiErr obj. As @bcmdarroch mentioned above, this will be cleaned up as a follow up.

**Comment by riddhi89** [comment]
@i0rek  
> Which changes were generated and which are the ones you did?
All the changes within docs/* were generated.

File: examples/resources/hcp_consul_cluster/federation.tf [hunk 6]
```diff
@@ -0,0 +1,15 @@
+data "hcp_hvn" "example" {
```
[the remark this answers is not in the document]
riddhi89: Done

internal/clients/consul_cluster.go
riddhi89: Done.

internal/provider/resource_consul_cluster.go
riddhi89: Done.

xargs-P:
Thanks @riddhi89 .
Can you also update the example in `docs/index.md` ?

docs/guides/consul-federation.md
```diff
@@ -0,0 +1,31 @@
+---
+subcategory: ""
+page_title: "Federate HCP Consul clusters - HCP Provider"
+description: |-
+    An example of federating a new HCP Consul cluster with an existing one.
+---
+
+# Federate a new HCP Consul cluster with an existing one
+
+Once you have a HCP Consul cluster, you can create new Consul cluster to federate with the existing one.
```
xargs-P, line 10: Thank you adding this guide. ❤️ 
Can we add an `a` to the sentence here?

> Once you have a HCP Consul cluster, you can create a new Consul cluster to federate with the existing one.

i0rek (approved):
Looks great!

internal/provider/resource_consul_cluster.go
```diff
@@ -261,6 +272,29 @@ func resourceConsulClusterCreate(ctx context.Context, d *schema.ResourceData, me
...
 
+	// If specified, validate and parse the primary link provided for federation.
+	primary_link, ok := d.GetOk("primary_link")
+	var primary *sharedmodels.HashicorpCloudLocationLink
+	if ok {
+		primary, err = parseLinkURL(primary_link.(string), ConsulClusterResourceType)
+		if err != nil {
+			return diag.Errorf(err.Error())
+		}
+		if primary.Location.OrganizationID == "" {
```
roaks3, line 283: Nit: With the implementation you have here, you can probably take this block of code out of the `primary.Location.OrganizationID == ""` conditional. `parseLinkURL()` parses the URL without attempting to resolve any additional information, so `primary.Location.OrganizationID` would always be empty.

roaks3 (approved):

internal/provider/resource_consul_cluster.go
riddhi89: Done

xargs-P (approved):
Thank you!

bcmdarroch (approved):

PR #73, opened 2021-03-15
HCPE-992 - Update github action to run unit tests instead of acceptance tests

roaks3:
The GitHub actions we use for PRs previously included running the acceptance tests (this came from the scaffolding). Since we have no acceptance tests yet, this was effectively only running unit tests and did not cause any problems, but when we start to add acceptance tests to this repository the job would fail due to authentication.

This changeset modifies the GitHub action to only run unit tests, so that we can begin to add acceptance tests without issue. Since HCP resources cost real money to provision, we do not plan to include acceptance tests as part of the GitHub checks moving forward.


bcmdarroch (approved):
Makes sense to me! As a follow-up, it'd be worth having a discussion about how we can have some long-running acceptance tests running internally (or perhaps even refactoring existing e2e tests to use this provider could serve the same purpose).

PR #74, opened 2021-03-16
HCPE-998 - Add acceptance tests for HVN resource

roaks3:
Add basic acceptance tests for creating, updating, and destroying an HVN.

Acceptance tests need to be run locally, which requires creating a service principal, and setting the env variables:
- `export HCP_CLIENT_ID=...`
- `export HCP_CLIENT_SECRET=...`
- `make acctest`

### DIFF
--- a/internal/provider/provider_test.go
+++ b/internal/provider/provider_test.go
@@ -1,18 +1,44 @@
 package provider
 
 import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+	"sync"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 )
+
+// testAccProvider is the "main" provider instance
+//
+// This Provider can be used in testing code for API calls without requiring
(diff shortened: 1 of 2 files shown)


bcmdarroch (approved):
Yay glad we're getting this in! Only thing I might add is a section of the README that describes the basics of acceptance tests (which could just be a link to the provider plugin's explanation).

I'm tempted to drop the `-v` from [this line in the Makefile](https://github.com/hashicorp/terraform-provider-hcp/blob/main/GNUmakefile#L17), since the complete output makes it hard to find the actual failure. That being said, maybe we keep it if we want to start requiring acceptance test output in PRs 💭 

💡 Also TIL: `go clean -testcache` lets you run the test with fresh results (otherwise it just returns the cached first successful run)

roaks3:
Ah yea, sorry if this wasn't clear, but this change is only meant to add an example acceptance test that utilizes our apis/sdk. The goal is to unblock anyone writing a new resource or feature that would want to include acceptance tests. We have a follow-up ticket to include some more thorough documentation [here](https://hashicorp.atlassian.net/browse/HCPE-1000).

PR #77, opened 2021-03-18
HCS-1805: Add consul cluster vm size

riddhi89:
- Adds a size attribute to the consul cluster properties. This can be provided during creation to specify t-shirt size of each server.
- Updates the hcp-go-sdk version.
- Cleans up some tier related validation in favor of leveraging sdk provided validation.

### DIFF
--- a/docs/data-sources/consul_cluster.md
+++ b/docs/data-sources/consul_cluster.md
@@ -49,8 +49,7 @@ data "hcp_consul_cluster" "example" {
 - **region** (String) The region where the HCP Consul cluster is located.
 - **scale** (Number) The the number of Consul server nodes in the cluster.
 - **self_link** (String) A unique URL identifying the HCP Consul cluster.
-- **size** (String) The t-shirt size representation of each server VM that this Consul cluster
-is provisioned with. Valid options - x_small, small, medium, large.
+- **size** (String) The t-shirt size representation of each server VM that this Consul cluster is provisioned with. Valid options - x_small, small, medium, large.
 - **tier** (String) The tier that the HCP Consul cluster will be provisioned as.  Only 'development' and 'standard' are available at this time.
 
 <a id="nestedblock--timeouts"></a>
--- a/docs/data-sources/consul_cluster.md
+++ b/docs/data-sources/consul_cluster.md
@@ -49,6 +49,8 @@ data "hcp_consul_cluster" "example" {
 - **region** (String) The region where the HCP Consul cluster is located.
 - **scale** (Number) The the number of Consul server nodes in the cluster.
 - **self_link** (String) A unique URL identifying the HCP Consul cluster.
+- **size** (String) The t-shirt size representation of each server VM that this Consul cluster
+is provisioned with. Valid options - x_small, small, medium, large.
 - **tier** (String) The tier that the HCP Consul cluster will be provisioned as.  Only 'development' and 'standard' are available at this time.
 
 <a id="nestedblock--timeouts"></a>
--- a/docs/data-sources/consul_cluster.md
+++ b/docs/data-sources/consul_cluster.md
@@ -49,8 +49,8 @@ data "hcp_consul_cluster" "example" {
 - **region** (String) The region where the HCP Consul cluster is located.
 - **scale** (Number) The the number of Consul server nodes in the cluster.
 - **self_link** (String) A unique URL identifying the HCP Consul cluster.
-- **size** (String) The t-shirt size representation of each server VM that this Consul cluster is provisioned with. Valid option for development tier - x_small. Valid options for other tiers - small, medium, large.
-- **tier** (String) The tier that the HCP Consul cluster will be provisioned as.  Only 'development' and 'standard' are available at this time.
+- **size** (String) The t-shirt size representation of each server VM that this Consul cluster is provisioned with. Valid option for development tier - `x_small`. Valid options for other tiers - `small`, `medium`, `large`.
+- **tier** (String) The tier that the HCP Consul cluster will be provisioned as.  Only `development` and `standard` are available at this time.
 
 <a id="nestedblock--timeouts"></a>
 ### Nested Schema for `timeouts`
--- a/docs/data-sources/consul_cluster.md
+++ b/docs/data-sources/consul_cluster.md
@@ -49,7 +49,7 @@ data "hcp_consul_cluster" "example" {
 - **region** (String) The region where the HCP Consul cluster is located.
 - **scale** (Number) The the number of Consul server nodes in the cluster.
 - **self_link** (String) A unique URL identifying the HCP Consul cluster.
-- **size** (String) The t-shirt size representation of each server VM that this Consul cluster is provisioned with. Valid options - x_small, small, medium, large.
+- **size** (String) The t-shirt size representation of each server VM that this Consul cluster is provisioned with. Valid option for development tier - x_small. Valid options for other tiers - small, medium, large.
 - **tier** (String) The tier that the HCP Consul cluster will be provisioned as.  Only 'development' and 'standard' are available at this time.
 
 <a id="nestedblock--timeouts"></a>
--- a/internal/clients/consul_cluster.go
+++ b/internal/clients/consul_cluster.go
@@ -3,14 +3,14 @@ package clients
 import (
 	"context"
 
-	"github.com/hashicorp/hcp-sdk-go/clients/cloud-consul-service/preview/2020-08-26/client/consul_service"
-	consulmodels "github.com/hashicorp/hcp-sdk-go/clients/cloud-consul-service/preview/2020-08-26/models"
+	"github.com/hashicorp/hcp-sdk-go/clients/cloud-consul-service/preview/2021-02-04/client/consul_service"
+	consulmodels "github.com/hashicorp/hcp-sdk-go/clients/cloud-consul-service/preview/2021-02-04/models"
 	sharedmodels "github.com/hashicorp/hcp-sdk-go/clients/cloud-shared/v1/models"
 )
 
 // GetConsulClusterByID gets an Consul cluster by its ID
 func GetConsulClusterByID(ctx context.Context, client *Client, loc *sharedmodels.HashicorpCloudLocationLocation,
-	consulClusterID string) (*consulmodels.HashicorpCloudConsul20200826Cluster, error) {
+	consulClusterID string) (*consulmodels.HashicorpCloudConsul20210204Cluster, error) {
 
 	getParams := consul_service.NewGetParams()
 	getParams.Context = ctx
@@ -31,7 +31,7 @@ func GetConsulClusterByID(ctx context.Context, client *Client, loc *sharedmodels
 // The files will be returned in base64-encoded format and will get passed in
 // that format.
 func GetConsulClientConfigFiles(ctx context.Context, client *Client, loc *sharedmodels.HashicorpCloudLocationLocation,
-	consulClusterID string) (*consulmodels.HashicorpCloudConsul20200826GetClientConfigResponse, error) {
+	consulClusterID string) (*consulmodels.HashicorpCloudConsul20210204GetClientConfigResponse, error) {
 
 	p := consul_service.NewGetClientConfigParams()
 	p.Context = ctx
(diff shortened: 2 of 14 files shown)


internal/clients/consul_cluster.go
```diff
@@ -3,22 +3,22 @@ package clients
...
-	consulmodels "github.com/hashicorp/hcp-sdk-go/clients/cloud-consul-service/preview/2020-08-26/models"
+	"github.com/hashicorp/hcp-sdk-go/clients/cloud-consul-service/preview/2021-02-04/client/consul_service"
+	consulmodels "github.com/hashicorp/hcp-sdk-go/clients/cloud-consul-service/preview/2021-02-04/models"
 	sharedmodels "github.com/hashicorp/hcp-sdk-go/clients/cloud-shared/v1/models"
 )
 
 // GetConsulClusterByID gets an Consul cluster by its ID
 func GetConsulClusterByID(ctx context.Context, client *Client, loc *sharedmodels.HashicorpCloudLocationLocation,
-	consulClusterID string) (*consulmodels.HashicorpCloudConsul20200826Cluster, error) {
+	consulClusterID string) (*consulmodels.HashicorpCloudConsul20210204Cluster, error) {
```
bcmdarroch, line 13: oh gosh this is annoying to have to change everywhere. 😅  just wrote up an SDK ticket for seeing if we can strip out that version number, so these don't have to be updated every time the service version bumps

internal/provider/resource_consul_cluster.go
```diff
@@ -37,20 +33,6 @@ var consulCusterResourceCloudProviders = []string{
 	"aws",
 }
 
-// consulClusterResourceTiers is the list of tiers
```
bcmdarroch, line 40: 😎

internal/provider/validators.go
```diff
@@ -106,3 +108,39 @@ func validateDatacenter(v interface{}, path cty.Path) diag.Diagnostics {
...
 	return diagnostics
 }
+
+func validateConsulClusterTier(v interface{}, path cty.Path) diag.Diagnostics {
+	var diagnostics diag.Diagnostics
+
+	err := consulmodels.HashicorpCloudConsul20210204ClusterConfigTier(strings.ToUpper(v.(string))).Validate(strfmt.Default)
+	if err != nil {
+		expectedEnumList := regexp.MustCompile(`\[.*\]`).FindStringSubmatch(err.Error())
+		msg := fmt.Sprintf("expected %v to be one of %v", v, expectedEnumList[0])
```
bcmdarroch, line 118: A little confusing that `expectedEnumList` seems to be a list of lists 🤔 Any way to flatten?

bcmdarroch (approved):
Everything looks good to me from the overall provider perspective! Thanks for looking into the enum validation pattern. I like that your approach keeps the API (by extension of the generated SDK) as the source of truth - definitely an improvement I'd like to add that to our style guide. 

I'll leave it to another consul team member to verify the business logic is all good.

internal/provider/resource_consul_cluster.go
```diff
@@ -421,18 +411,16 @@ func setConsulClusterResourceData(d *schema.ResourceData, cluster *consulmodels.
...
 
-	// TODO: Update this logic when tier becomes a first class value on the cluster
-	for t, numServers := range consulTierToNumServers {
-		if numServers == cluster.Config.CapacityConfig.NumServers {
-			if err := d.Set("tier", t); err != nil {
-				return err
-			}
-			break
-		}
+	if err := d.Set("tier", strings.ToLower(string(cluster.Config.Tier))); err != nil {
```
roaks3, line 418: Nit: Since you're already using `DiffSuppressFunc`, I don't think you need to normalize this with `ToLower` (same with size)

internal/provider/validators.go
```diff
@@ -106,3 +108,39 @@ func validateDatacenter(v interface{}, path cty.Path) diag.Diagnostics {
 
 	return diagnostics
 }
+
+func validateConsulClusterTier(v interface{}, path cty.Path) diag.Diagnostics {
```
roaks3, line 112: Very cool pattern for these `validate*` functions, thanks for adding them (and tests)!

internal/provider/resource_consul_cluster.go
```diff
@@ -135,6 +117,16 @@ func resourceConsulCluster() *schema.Resource {
 				Optional:    true,
 				ForceNew:    true,
 			},
+			"size": {
+				Description:      "The t-shirt size representation of each server VM that this Consul cluster\nis provisioned with. Valid options - x_small, small, medium, large.",
```
bcmdarroch, line 121: Trying to think of how we could ensure this description stays up-to-date with the enum values, but it looks like the SDK doesn't provide them in a list, just constants. 🤔

internal/clients/consul_cluster.go
riddhi89: Thank you! That would be great.

internal/provider/data_source_consul_cluster.go
```diff
@@ -117,6 +117,11 @@ func dataSourceConsulCluster() *schema.Resource {
 				Type:        schema.TypeString,
 				Computed:    true,
 			},
+			"size": {
+				Description: "The t-shirt size representation of each server VM that this Consul cluster\nis provisioned with. Valid options - x_small, small, medium, large.",
```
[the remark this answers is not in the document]
riddhi89: Good catch. This was just a copy pasta of the desc from elsewhere. Removed.

internal/provider/resource_consul_cluster.go
riddhi89: Done!

internal/provider/resource_consul_cluster.go
riddhi89: Yea this was the other place I did not like the hard-coding :'( The compromise was at least the sdk validation will be our gate check. But, open to suggestions if any.

internal/provider/validators.go
```diff
@@ -106,3 +108,39 @@ func validateDatacenter(v interface{}, path cty.Path) diag.Diagnostics {
...
+	var diagnostics diag.Diagnostics
+
+	err := consulmodels.HashicorpCloudConsul20210204ClusterConfigTier(strings.ToUpper(v.(string))).Validate(strfmt.Default)
+	if err != nil {
+		expectedEnumList := regexp.MustCompile(`\[.*\]`).FindString(err.Error())
+		msg := fmt.Sprintf("expected %v to be one of %v", v, expectedEnumList)
+		diagnostics = append(diagnostics, diag.Diagnostic{
+			Severity:      diag.Error,
+			Summary:       msg,
+			Detail:        msg + " (value can be case-insensitive).",
```
Kedevlin, line 122: nit 
```suggestion
			Detail:        msg + " (value is case-insensitive).",
```

internal/provider/validators.go
```diff
@@ -106,3 +108,39 @@ func validateDatacenter(v interface{}, path cty.Path) diag.Diagnostics {
...
+	var diagnostics diag.Diagnostics
+
+	err := consulmodels.HashicorpCloudConsul20210204CapacityConfigSize(strings.ToUpper(v.(string))).Validate(strfmt.Default)
+	if err != nil {
+		expectedEnumList := regexp.MustCompile(`\[.*\]`).FindString(err.Error())
+		msg := fmt.Sprintf("expected %v to be one of %v", v, expectedEnumList)
+		diagnostics = append(diagnostics, diag.Diagnostic{
+			Severity:      diag.Error,
+			Summary:       msg,
+			Detail:        msg + " (value can be case-insensitive).",
```
Kedevlin, line 140: nit 
```suggestion
			Detail:        msg + " (value is case-insensitive).",
```

internal/provider/validators_test.go
```diff
@@ -261,3 +261,83 @@ func Test_validateDatacenter(t *testing.T) {
...
+		"valid tier mixedcase": {
+			input:    "DEVelopment",
+			expected: nil,
+		},
+		"invalid tier": {
+			input: "dev",
+			expected: diag.Diagnostics{
+				diag.Diagnostic{
+					Severity:      diag.Error,
+					Summary:       "expected dev to be one of [UNSET DEVELOPMENT STANDARD]",
```
Kedevlin, line 287: this is interesting because tier=UNSET will fail in Consul Service since it is validated as required. I like the validation functions you added but is there an easy way to remove UNSET as being suggested? (same comment about the VM size validation).

Kedevlin (commented):
Looking good! A few mostly minor comments

internal/provider/validators_test.go
Kedevlin: RE "same comment about the VM size validation" above. I thought more about it and this isn't the same for VM size since it's optional. But still seems weird to be suggesting UNSET as an option.

docs/data-sources/consul_cluster.md
```diff
@@ -49,6 +49,7 @@ data "hcp_consul_cluster" "example" {
 - **region** (String) The region where the HCP Consul cluster is located.
 - **scale** (Number) The the number of Consul server nodes in the cluster.
 - **self_link** (String) A unique URL identifying the HCP Consul cluster.
+- **size** (String) The t-shirt size representation of each server VM that this Consul cluster is provisioned with. Valid options - x_small, small, medium, large.
```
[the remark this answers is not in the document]
riddhi89: Agreed on consistency on external messaging. Is there a link/will there be a link available with this info ? Just so that we do not have to duplicate changes to the messaging in multiple places.

internal/provider/validators_test.go
riddhi89: Yea UNSET is a weird option for sure. I am guessing it originates from the proto file having it [here](https://github.com/hashicorp/cloud-consul-service/blob/master/proto-public/20210204/config.proto#L32) and gets propogated in by the swagger tools. 
Wondering instead of doing any manipulation here, we can avoid generating it in the [SDK](https://github.com/hashicorp/hcp-sdk-go/blob/main/clients/cloud-consul-service/preview/2021-02-04/models/hashicorp_cloud_consul20210204_cluster_config_tier.go#L24) itself 🤔

riddhi89: @bcmdarroch @roaks3 any thoughts ^^? 
That said, I am happy to add some manipulation to remove the UNSET option from the suggested list (might be easier in the short term too). Let me know.

bcmdarroch: Yeah I think in the short-term it'd be best to remove it here. Long-term, I'm thinking of updating the SDK template to provide a constant that's an array of all possible enum values, and we could drop the UNSET there. Then that constant could be used in the validator function, in the documentation, etc.

riddhi89: Great! Thanks @bcmdarroch!

riddhi89: Removed the UNSET option for the short term.

docs/data-sources/consul_cluster.md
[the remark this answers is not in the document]
riddhi89: Updated based on discussion with @srikrishmurthy.

docs/data-sources/consul_cluster.md
```diff
@@ -49,6 +49,7 @@ data "hcp_consul_cluster" "example" {
 - **region** (String) The region where the HCP Consul cluster is located.
 - **scale** (Number) The the number of Consul server nodes in the cluster.
 - **self_link** (String) A unique URL identifying the HCP Consul cluster.
+- **size** (String) The t-shirt size representation of each server VM that this Consul cluster is provisioned with. Valid option for development tier - x_small. Valid options for other tiers - small, medium, large.
```
Kedevlin, line 52: nit: not sure how much this matters but I noticed the style below with tier puts single quotes around the value options. Maybe we should do that here too for consistency?
```suggestion
- **size** (String) The t-shirt size representation of each server VM that this Consul cluster is provisioned with. Valid option for development tier - 'x_small'. Valid options for other tiers - 'small', 'medium', 'large'.
```

Kedevlin (approved):
Looks great! Just one nit comment

docs/data-sources/consul_cluster.md
riddhi89: Done

PR #78, opened 2021-03-19
HCPE-1032 - Add basic consul acceptance test

roaks3:
Add a simple Consul acceptance test, which mainly just verifies that all of the fields get set appropriately on create.

Note for anyone interested in testing this locally: this test was taking >10 mins for me to run.

### DIFF
--- a/internal/provider/resource_hvn_test.go
+++ b/internal/provider/resource_hvn_test.go
@@ -11,7 +11,7 @@ import (
 )
 
 var (
-	hvn = fmt.Sprintf(`
+	testAccHvnConfig = fmt.Sprintf(`
 resource "hcp_hvn" "test" {
 	hvn_id         = "test-hvn"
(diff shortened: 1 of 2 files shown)


internal/provider/resource_hvn_test.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 var (
-	hvn = fmt.Sprintf(`
+	testAccHvnConfig = fmt.Sprintf(`
```
bcmdarroch, line 14: is this a convention we need to follow?

bcmdarroch (approved):
👍 👍 ran these locally and they passed!

internal/provider/resource_hvn_test.go
roaks3: No, I only changed this so that the name was more clear and less likely to be reused by other code in the package

PR #79, opened 2021-03-19
HCPE-1000 - Update contribution docs to include guidance on acceptance tests

roaks3:
Add documentation for running and writing acceptance tests.

This is largely copied from https://github.com/hashicorp/terraform-provider-aws/blob/master/docs/contributing/running-and-writing-acceptance-tests.md, and tweaked for this project specifically.


contributing/writing-tests.md
```diff
@@ -0,0 +1,199 @@
...
+common for a single test function to exercise both the creation of and updates
+to a single resource. Most tests follow a similar structure.
+
+1. Pre-flight checks are made to ensure that sufficient provider configuration
+   is available to be able to proceed - in the case of HCP, `HCP_CLIENT_ID` and `HCP_CLIENT_SECRET` must be set prior
+   to running acceptance tests. This is common to all tests exercising a single
+   provider.
+
+Each `TestStep` is defined in the call to `resource.Test()`. Most assertion
+functions are defined out of band with the tests. This keeps the tests
```
bcmdarroch, line 66: Where do those assertions live?

contributing/writing-tests.md
```diff
@@ -0,0 +1,199 @@
...
+
+    resource.Test(t, resource.TestCase{
+        PreCheck:          func() { testAccPreCheck(t) },
+        ProviderFactories: providerFactories,
+        CheckDestroy:      testAccCheckConsulClusterDestroy,
+        Steps: []resource.TestStep{
+            {
+                Config: testConfig(testAccConsulClusterConfig),
+                Check: resource.ComposeTestCheckFunc(
+                    testAccCheckConsulClusterExists(resourceName),
```
bcmdarroch, line 82: So this assertion you wrote, right? And the one below is a generic one the provider test suite gives you?

bcmdarroch (approved):
This is super informative! I think the only thing missing is a summary of the minimum test assertions to include (based on what you have here: test the resource exists, test the required inputs are given)

contributing/writing-tests.md
bcmdarroch: Oo found 'em: https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/resource/testing.go
I would definitely add a link to those here for reference.

contributing/writing-tests.md
```diff
@@ -0,0 +1,199 @@
...
+$ make testacc TESTARGS='-run=TestAccConsulCluster'
+TF_ACC=1 go test ./... -v -run=TestAccConsulCluster -timeout 120m
+ok  	github.com/hashicorp/terraform-provider-hcp/internal/consul	(cached) [no tests to run]
+=== RUN   TestAccConsulCluster
+--- PASS: TestAccConsulCluster (538.48s)
+PASS
+ok  	github.com/hashicorp/terraform-provider-hcp/internal/provider	539.112s
+```
+
+Entire resource test suites can be targeted by using the naming convention to
```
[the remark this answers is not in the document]
roaks3: I don't think so, this is covered in the Terraform docs: https://www.terraform.io/docs/extend/testing/acceptance-tests/testcase.html

contributing/writing-tests.md
```diff
@@ -0,0 +1,199 @@
...
+For advanced developers, the acceptance testing framework accepts some additional environment variables that can be used to control Terraform CLI binary selection, logging, and other behaviors. See the [Extending Terraform documentation](https://www.terraform.io/docs/extend/testing/acceptance-tests/index.html#environment-variables) for more information.
+
+## Writing an Acceptance Test
+
+Terraform has a framework for writing acceptance tests which minimises the
+amount of boilerplate code necessary to use common testing patterns. The entry
+point to the framework is the `resource.Test()` function.
+
+Tests are divided into `TestStep`s. Each `TestStep` proceeds by applying some
+Terraform configuration using the provider under test, and then verifying that
```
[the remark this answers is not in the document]
roaks3: I think this refers to the primary HCP provider, where an acceptance test can additionally use other providers, eg. an AWS provider to peer with

contributing/writing-tests.md
roaks3: They can also be found in the Terraform docs, which we already link to in a few different places: https://www.terraform.io/docs/extend/testing/acceptance-tests/teststep.html

I don't think we need to add another link here, this content is intended to augment those testing docs provided by Terraform (added a line at the top to clarify that)

contributing/writing-tests.md
roaks3: Yep, that's correct

contributing/writing-tests.md
```diff
@@ -0,0 +1,200 @@
+# Writing Acceptance Tests
+
+Terraform includes an acceptance test harness that does most of the repetitive
+work involved in testing a resource. For additional information about testing
+Terraform Providers, see the [Extending Terraform documentation](https://www.terraform.io/docs/extend/testing/index.html).
+This guide is meant to augment that documentation with information specific to the HCP provider.
```
bcmdarroch, line 6: I might word this a little more strongly: `We suggest reading this documentation first`. I originally read this doc assuming that it would tell me everything I'd need to know to start writing acceptance tests.

roaks3: Ah yea, that definitely makes sense, will update

PR #80, opened 2021-03-19
HCPE-990 - Add PR template

roaks3:

Add a PR template to include reminders and help guide newcomers

### :building_construction: Testing

N/A

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,24 @@
+<!--
+Adding a new resource or datasource? Use this checklist to get started: https://github.com/hashicorp/terraform-provider-hcp/blob/main/contributing/checklist-resource.md
+-->
+
+### :hammer_and_wrench: Description
+
+<!-- What code changed, and why? -->
+
+### :building_construction: Testing
+
+<!-- Include any feature flags that are required for testing. -->
+
+Output from acceptance testing:
+
+<!--
+Replace TestAccXXX with a pattern that matches the tests affected by this PR. More info on acceptance tests here: https://github.com/hashicorp/terraform-provider-hcp/blob/main/contributing/writing-tests.md
+
+For more information on the `-run` flag, see the `go test` documentation at https://tip.golang.org/cmd/go/#hdr-Testing_flags.
+-->
+```
+$ make testacc TESTARGS='-run=TestAccXXX'
+
+...
+```
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -4,7 +4,7 @@ Adding a new resource or datasource? Use this checklist to get started: https://
 
 ### :hammer_and_wrench: Description
 
-<!-- What code changed, and why? -->
+<!-- What code changed, and why? If adding a new resource, what is it and what are its key features? If updating an existing resource, what new functionality was added? -->
 
 ### :building_construction: Testing
 
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -6,9 +6,11 @@ Adding a new resource or datasource? Use this checklist to get started: https://
 
 <!-- What code changed, and why? If adding a new resource, what is it and what are its key features? If updating an existing resource, what new functionality was added? -->
 
-### :building_construction: Testing
+### :building_construction: Acceptance tests
 
-<!-- Include any feature flags that are required for testing. -->
+- [ ] Are there any feature flags that are required to use this functionality?
+- [ ] Have you added an acceptance test for the functionality being added?
+- [ ] Have you run the acceptance tests on this branch?
 
 Output from acceptance testing:
 


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,24 @@
...
+<!--
+Replace TestAccXXX with a pattern that matches the tests affected by this PR. More info on acceptance tests here: https://github.com/hashicorp/terraform-provider-hcp/blob/main/contributing/writing-tests.md
+
+For more information on the `-run` flag, see the `go test` documentation at https://tip.golang.org/cmd/go/#hdr-Testing_flags.
+-->
+```
+$ make testacc TESTARGS='-run=TestAccXXX'
+
+...
+```
```
bcmdarroch, line 24: A few sections I'd like to add:
```
### :link: External Links

<!-- Include any links here that might be helpful for people reviewing your PR. If there are none, feel free to delete this section. -->

### :+1: Definition of Done

- [ ] New <service> resource/data source/import meets all requirements defined in the [resource checklist](https://github.com/hashicorp/terraform-provider-hcp/blob/main/contributing/checklist-resource.md)
- [ ] Tests added
- [ ] Docs updated
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,24 @@
...
+Adding a new resource or datasource? Use this checklist to get started: https://github.com/hashicorp/terraform-provider-hcp/blob/main/contributing/checklist-resource.md
+-->
+
+### :hammer_and_wrench: Description
+
+<!-- What code changed, and why? -->
+
+### :building_construction: Testing
+
+<!-- Include any feature flags that are required for testing. -->
```
bcmdarroch, line 11: What is this section for exactly? For reporting acceptance test output? For demonstrating how to test changes locally? I would add more detail to the comment.

bcmdarroch (approved):
🙌 Looks good to me, though I'd suggest adding a Definition of Done checklist section.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,24 @@
+<!--
+Adding a new resource or datasource? Use this checklist to get started: https://github.com/hashicorp/terraform-provider-hcp/blob/main/contributing/checklist-resource.md
```
bcmdarroch, line 2: Worried this comment might get ignored, so I'd move the link into the checklist I described below.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,24 @@
...
+-->
+
+### :hammer_and_wrench: Description
+
+<!-- What code changed, and why? -->
```
bcmdarroch, line 7: We could add some guiding questions here, like:
```
 If adding a new resource, what is it and what are its key features? If updating an existing resource, what new functionality was added?

.github/PULL_REQUEST_TEMPLATE.md
roaks3: The PR template here is loosely based off of the [AWS provider template](https://raw.githubusercontent.com/hashicorp/terraform-provider-aws/main/.github/PULL_REQUEST_TEMPLATE.md), which I assume has not been causing trouble for them with the amount of PR volume they receive. Could you expand on the value you are expecting from adding these two sections? I would be a bit concerned about adding weight to the template if it would not be used or be redundant (mainly thinking about reducing friction for contributors, and the Cloud Consul team recently [reduced the size of their template](https://hashicorp.atlassian.net/browse/HCS-1757)).

bcmdarroch: Hmm External Links doesn't add much, but I like the Definition of Done section as a final dot-your-i's/cross-your-t's reminder. I think it provides some bare minimum goal posts for a PR

.github/PULL_REQUEST_TEMPLATE.md
roaks3: Added more detail here based off of the Kubernetes provider template

.github/PULL_REQUEST_TEMPLATE.md
roaks3: Ok, that makes sense, but a few comments based on the updates I've made:
- I just added some more info around testing in the `Acceptance tests` section based off of what the Kubernetes provider uses, so now looking at the `Definition of Done` section you have here, I think that the `Acceptance tests added` item is a bit redundant.
- Looking at the `Docs updated` item, I think that we should actually use an automated check like [this](https://github.com/hashicorp/terraform-provider-aws/blob/main/.github/workflows/terraform_provider.yml#L156-L187), so that a checklist item is not necessary.

Do you think it would make sense to still add a `Definition of Done` section with only the `New <service> resource ...` checklist item? Are there other items we think might go in here? Again, mainly bringing this up because the other providers do not appear to use a section like this.

bcmdarroch: Yeah with the changes you made, DoD sections seems unnecessary.

roaks3: Ok cool, going to leave that off for now then, and we can see if we need to adjust

.github/PULL_REQUEST_TEMPLATE.md
roaks3: Since we're omitting a `Definition of Done` for now, I'm going to leave this here in the comment, but agreed that we should be ready to tweak this if we are noticing contributors skipping past it

PR #82, opened 2021-03-23
Update resources to be in line with current patterns

roaks3:
These changes will probably be easiest to review one commit at a time. The updates largely come from items we ran into while working on the TGW resource, but did not yet backport to the other resources:

- **fix:** Ensure context is being passed for all HCP API calls - https://github.com/hashicorp/terraform-provider-hcp/commit/27be4da53d37f8199e87442c8ad2df0ad739c49f
- **improvement:** Update comments, docs, and messages to use correct capitalization for network peering - https://github.com/hashicorp/terraform-provider-hcp/commit/e8015376a8d2f7215bb631a54150222b909769e9
- **improvement:** Log import ID used when an import fails due to parsing - https://github.com/hashicorp/terraform-provider-hcp/commit/8204243661125969be973ea0f742cef229284465
- **improvement:** Add comment to clarify that Links can be sent in API requests - https://github.com/hashicorp/terraform-provider-hcp/commit/45e03702e98254687755f122604370223e7cbdc0
- **improvement:** Update peering wait function to use helper - https://github.com/hashicorp/terraform-provider-hcp/commit/86ad65206fdce9a6cd94872faf09b6e4407c38c2

### DIFF
--- a/internal/provider/resource_aws_network_peering.go
+++ b/internal/provider/resource_aws_network_peering.go
@@ -198,7 +198,7 @@ func resourceAwsNetworkPeeringCreate(ctx context.Context, d *schema.ResourceData
 
 	log.Printf("[INFO] Created network peering (%s) between HVN (%s) and peer (%s)", peering.ID, peering.Hvn.ID, peering.Target.AwsTarget.VpcID)
 
-	peering, err = clients.WaitForPeeringToBePendingAcceptance(ctx, client, peering.ID, hvnID, loc)
+	peering, err = clients.WaitForPeeringToBePendingAcceptance(ctx, client, peering.ID, hvnID, loc, d.Timeout(schema.TimeoutCreate))
 	if err != nil {
 		return diag.FromErr(err)
 	}
--- a/internal/provider/resource_aws_network_peering.go
+++ b/internal/provider/resource_aws_network_peering.go
@@ -22,7 +22,7 @@ var peeringDeleteTimeout = time.Minute * 35
 
 func resourceAwsNetworkPeering() *schema.Resource {
 	return &schema.Resource{
-		Description: "The AWS Network peering resource allows you to manage a Network peering between an HVN and a peer AWS VPC.",
+		Description: "The AWS network peering resource allows you to manage a network peering between an HVN and a peer AWS VPC.",
 
 		CreateContext: resourceAwsNetworkPeeringCreate,
 		ReadContext:   resourceAwsNetworkPeeringRead,
@@ -75,7 +75,7 @@ func resourceAwsNetworkPeering() *schema.Resource {
 			},
 			// Optional inputs
 			"peering_id": {
-				Description:      "The ID of the Network peering.",
+				Description:      "The ID of the network peering.",
 				Type:             schema.TypeString,
 				Optional:         true,
 				ForceNew:         true,
@@ -84,12 +84,12 @@ func resourceAwsNetworkPeering() *schema.Resource {
 			},
 			// Computed outputs
 			"organization_id": {
-				Description: "The ID of the HCP organization where the Network peering is located. Always matches the HVN's organization.",
+				Description: "The ID of the HCP organization where the network peering is located. Always matches the HVN's organization.",
 				Type:        schema.TypeString,
 				Computed:    true,
 			},
 			"project_id": {
-				Description: "The ID of the HCP project where the Network peering is located. Always matches the HVN's project.",
+				Description: "The ID of the HCP project where the network peering is located. Always matches the HVN's project.",
 				Type:        schema.TypeString,
 				Computed:    true,
 			},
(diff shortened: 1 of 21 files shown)


internal/provider/resource_aws_network_peering.go
```diff
@@ -22,7 +22,7 @@ var peeringDeleteTimeout = time.Minute * 35
 
 func resourceAwsNetworkPeering() *schema.Resource {
 	return &schema.Resource{
-		Description: "The AWS Network peering resource allows you to manage a Network peering between an HVN and a peer AWS VPC.",
+		Description: "The AWS Network Peering resource allows you to manage a network peering between an HVN and a peer AWS VPC.",
```
bcmdarroch, line 25: The first `Network Peering` should be lowercased too. Same with the data source (and import maybe?). 
I also noticed this in the `Transit Gateway Attachment` docs.

bcmdarroch (approved):
lookin tidy! 🧹 ✨

internal/provider/resource_aws_network_peering.go
roaks3: Ah yea, this actually came up before for the transit gateway resources and I got the following guidance from Josh: https://github.com/hashicorp/terraform-provider-hcp/pull/58#discussion_r573203443

In short, because there is a leading `AWS`, the entire noun gets title-cased. I'm not sure if the guidance has changed, but if not I think this is correct as-is.

bcmdarroch: Ooh I see 🤔 We might need to fix up some of the consul resources then, they seemed to stick with sentence case even in the descriptions (https://github.com/hashicorp/terraform-provider-hcp/blob/main/internal/provider/resource_consul_cluster.go#L57).

roaks3: Oh boy, we've definitely found some edge cases with these 😄 

@joshklekamp can you verify that we still have this right? The following line that @bcmdarroch linked seems like it would be incorrect, and we would want to capitalize the second `cluster`:

`The Consul cluster resource allows you to manage an HCP Consul cluster.`

joshklekamp: I don't believe we capitalize cluster, only branded features. I'm not familiar with the term "AWS Network Peering resource" though. If it's a feature of AWS's, I say keep it capitalized, but if not, we might want to revise to "AWS network peering resource".

bcmdarroch: According to [this comment](https://github.com/hashicorp/terraform-provider-hcp/pull/58#discussion_r573203443):
- Top-level descriptions are all capitalized: `AWS Transit Gateway Attachment, AWS Network Peering, Consul Cluster` ( only [TGW](https://registry.terraform.io/providers/hashicorp/hcp/latest/docs/resources/aws_transit_gateway_attachment) does this currently, in the first paragraph)
- Everywhere else is sentence-cased: `AWS transit gateway attachment, AWS network peering, Consul cluster`

We can definitely go forward with that, and I'll add a ticket for getting consistency across the resources. 
But this does have me wondering if the top-level description exception is worth having, since it just ends up in the first paragraph, not a heading. 🤔 

Also to throw in another wrench, it looks like AWS capitalizes [Transit Gateways](https://aws.amazon.com/transit-gateway/?whats-new-cards.sort-by=item.additionalFields.postDateTime&whats-new-cards.sort-order=desc) and lower-cases [peering](https://aws.amazon.com/peering/). So do we add a rule that respects the casing of the brand for certain resources?

joshklekamp: It might be helpful to pull in Kaitlin here. I tend to differ to her since design tries to align to the Learn team's writing style. I think the examples you provided @bcmdarroch might need a little revision. I think design's POV would be that: 

- Product-specific features are all capitalized: `AWS Transit Gateway Attachment, AWS Peering Connection`
- Everywhere else is sentence-cased: transit gateway attachment, network peering, , peering connection, Consul cluster

roaks3: @joshklekamp thanks for checking in here! What you have described in your last comment is essentially what we do now, but I think we are just off on how to implement the guidance. It sounds like we want to use `AWS Transit Gateway Attachment` capitalization when we are referring to the AWS-specific feature by name, but we actually don't do that anywhere in these docs. In the top-level descriptions, we are referring to *our* HCP Terraform resource by name (ie. a human-readable version of `aws_transit_gateway_attachment`).

> I'm not familiar with the term "AWS Network Peering resource" though. If it's a feature of AWS's, I say keep it capitalized, but if not, we might want to revise to "AWS network peering resource".

So it sounds like we would want `The AWS network peering resource` for peering, and `The AWS transit gateway attachment resource` for TGW. I can just update both resources in this PR.

PR #84, opened 2021-03-24
HCPE-1034: vault resource scaffolding

bcmdarroch:

This PR adds the schema for the new Vault cluster resource. It takes some inspiration from the Consul cluster resource.

Sample HCL:
```
  # required
  cluster_id = "testing"
  hvn_id = "test-hvn"
  
  # optional
  public_endpoint = true
  initial_vault_version = "1.7.0"
}
```

Sample `terraform plan` output:
```
# hcp_vault_cluster.example will be created
+ resource "hcp_vault_cluster" "example" {
    + cloud_provider               = (known after apply)
    + cluster_id                   = "testing"
    + created_at                   = (known after apply)
    + hvn_id                       = "test-hvn"
    + id                           = (known after apply)
    + namespace                    = "admin"
    + organization_id              = (known after apply)
    + project_id                   = (known after apply)
    + region                       = (known after apply)
    + tier                         = "development"
    + vault_version                = (known after apply)
    + vault_private_endpoint_url   = (known after apply)
    + vault_public_endpoint_url    = (known after apply)
}
```

Sample HCL:
```
resource "hcp_vault_cluster_admin_token" "example" {
  cluster_id = "testing"
}
```

Sample `terraform plan` output:
```
# hcp_vault_cluster_admin_token.example will be created
+ resource "hcp_vault_cluster_admin_token" "example" {
    + cluster_id                   = "testing"
    + token                        = (known after apply)
}
```

### DIFF
--- a/internal/provider/resource_vault_cluster.go
+++ b/internal/provider/resource_vault_cluster.go
@@ -50,7 +50,6 @@ func resourceVaultCluster() *schema.Resource {
 				ForceNew:         true,
 				ValidateDiagFunc: validateSlugID,
 			},
-			// TODO: this is required for consul, but is this complete for vault yet? we can also mark computed output until it's ready.
 			"tier": {
 				Description: "The tier that the HCP Vault cluster will be provisioned as.  Only 'development' and 'standard' are available at this time.",
 				Type:        schema.TypeString,
@@ -129,6 +128,11 @@ func resourceVaultCluster() *schema.Resource {
 			},
 			"vault_private_endpoint_url": {
 				Description: "The private URL for the Vault UI.",
+				Type:        schema.TypeString,
+				Computed:    true,
+			},
+			"created_at": {
+				Description: "The time that the Vault cluster was created.",
 				Type:        schema.TypeString,
 				Computed:    true,
 			},
--- a/internal/provider/resource_vault_cluster.go
+++ b/internal/provider/resource_vault_cluster.go
@@ -2,7 +2,6 @@ package provider
 
 import (
 	"context"
-	"strings"
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
@@ -48,17 +47,6 @@ func resourceVaultCluster() *schema.Resource {
 				ForceNew:         true,
 				ValidateDiagFunc: validateSlugID,
 			},
-			"tier": {
-				Description: "The tier that the HCP Vault cluster will be provisioned as.  Only 'development' and 'standard' are available at this time.",
-				Type:        schema.TypeString,
-				Required:    true,
-				ForceNew:    true,
-				// TODO: generalize consul enum validator
-				// ValidateDiagFunc: validateStringInSlice(vaultClusterResourceTiers, true),
-				DiffSuppressFunc: func(_, old, new string, _ *schema.ResourceData) bool {
-					return strings.ToLower(old) == strings.ToLower(new)
-				},
-			},
 			// optional fields
 			"public_endpoint": {
 				Description: "Denotes that the cluster has a public endpoint for the Vault UI. Defaults to false.",
@@ -77,6 +65,12 @@ func resourceVaultCluster() *schema.Resource {
 				// },
 			},
 			// computed outputs
+			// TODO: once more tiers are supported and can be changed by users, make this a required input.
+			"tier": {
+				Description: "The tier that the HCP Vault cluster will be provisioned as.  Only 'development' is available at this time.",
+				Type:        schema.TypeString,
+				Computed:    true,
+			},
 			"organization_id": {
 				Description: "The ID of the organization this HCP Vault cluster is located in.",
 				Type:        schema.TypeString,
--- a/internal/provider/resource_vault_cluster.go
+++ b/internal/provider/resource_vault_cluster.go
@@ -1,9 +1,53 @@
 package provider
 
-import "github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+import (
+	"context"
+	"time"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+// defaultClusterTimeout is the amount of time that can elapse
+// before a cluster read operation should timeout.
+var defaultVaultClusterTimeout = time.Minute * 5
+
+// createUpdateTimeout is the amount of time that can elapse
+// before a cluster create or update operation should timeout.
+var createUpdateVaultClusterTimeout = time.Minute * 35
+
+// deleteTimeout is the amount of time that can elapse
+// before a cluster delete operation should timeout.
+var deleteVaultClusterTimeout = time.Minute * 25
 
 func resourceVaultCluster() *schema.Resource {
 	return &schema.Resource{
-		Description: "The Vault cluster resource allows you to manage an HCP Vault cluster.",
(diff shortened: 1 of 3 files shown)


internal/provider/resource_vault_cluster.go
```diff
@@ -0,0 +1,153 @@
...
+				ValidateDiagFunc: validateSlugID,
+			},
+			"hvn_id": {
+				Description:      "The ID of the HVN this HCP Vault cluster is associated to.",
+				Type:             schema.TypeString,
+				Required:         true,
+				ForceNew:         true,
+				ValidateDiagFunc: validateSlugID,
+			},
+			// TODO: this is required for consul, but is this complete for vault yet? we can also mark computed output until it's ready.
```
bcmdarroch, line 53: Need to confirm with Vault team.

internal/provider/resource_vault_cluster.go
```diff
@@ -0,0 +1,153 @@
...
+				Type:        schema.TypeBool,
+				Default:     false,
+				Optional:    true,
+				ForceNew:    true,
+			},
+			"namespace": {
+				Description: "The name of the customer namespace.",
+				Type:        schema.TypeString,
+				Optional:    true,
+				// TODO Confirm:
```
bcmdarroch, line 77: Does a new namespace mean new cluster?

internal/provider/resource_vault_cluster.go
```diff
@@ -0,0 +1,153 @@
...
+				// ForceNew:    true,
+			},
+			"max_lease_ttl": {
+				Description: "The max lease time-to-live (TTL) for this Vault cluster.",
+				Type:        schema.TypeString,
+				Optional:    true,
+				// TODO Confirm:
```
bcmdarroch, line 84: Does a new max lease TTL mean new cluster?

internal/provider/resource_vault_cluster.go
```diff
@@ -0,0 +1,153 @@
...
+				// ForceNew:    true,
+			},
+			// TODO: min_vault_version for consistency with Consul?
```
bcmdarroch, line 87: `min_vault_version` or `initial_vault version`? or something else?

internal/provider/resource_vault_cluster.go
thorfour: Sorry, what is required for consul? I'm unsure what this comment is referring to.

internal/provider/resource_vault_cluster.go
thorfour: Yes it would. We don't have a way to rename the namespace for the customer right now. I don't believe there is a way to rename a namespace in Vault.

internal/provider/resource_vault_cluster.go
thorfour: I don't think it technically has to, but I don't believe we have a way to update that currently.

internal/provider/resource_vault_cluster.go
thorfour: I feel like `initial_vault_version` makes more sense semantically. But if it needs to change for consistency I'm sure we can do that.

internal/provider/resource_vault_cluster.go
bcmdarroch: Sorry, to rephrase: Is the `tier` feature complete for Vault clusters? Or is it still hidden behind a feature flag?
I ask because, if it's complete, I can include it as a required input on the cluster (like consul cluster does). If it's not complete, I can make it a computed output that gets automatically set to `development`. Or leave it out entirely until it's ready to be exposed.

Also, a more general question: what parts of the vault cluster API are currently feature flagged?

internal/provider/resource_vault_cluster.go
bcmdarroch: Is there a default name it gets set to? I could include that as a computed output.

internal/provider/resource_vault_cluster.go
xargs-P: What does the Vault API call it today?

xargs-P (commented):
Will the admin/root token resource come in another PR?

internal/provider/resource_vault_cluster.go
bcmdarroch: It's called `initial_vault_version` in the Vault API. The question behind this question is, do we want HCP TF resources to be consistent with the service APIs, regardless of inconsistency across the services, or we do want to enforce some common language across the HCP provider as much as possible?

I might take a look at how the UI handles that problem.

bcmdarroch:
I'll go ahead and add the admin token schema to this one 👍 👍

internal/provider/resource_vault_cluster.go
bcmdarroch: TODO: ticket to change `min_consul_version` on consul service to `initial_consul_version` for consistency.

internal/provider/resource_vault_cluster.go
bcmdarroch: TODO: based on some slack research, it looks like tiers can be used, so we'll make this a required input for now.

internal/provider/resource_vault_cluster.go
bcmdarroch: TODO: investigate cloud-ui's default namespace

xargs-P: @thorfour  if we allow Customers to set the name in TF, will that break any backend processes that are assuming the default namespace we assign ?

bcmdarroch: Confirmed with Thor today that `namespace` can't be changed by the user, so I bumped that down to computed outputs.

internal/provider/resource_vault_cluster.go
```diff
@@ -0,0 +1,137 @@
...
+			},
+			"initial_vault_version": {
+				Description:      "The initial Vault version to use when creating the cluster. Once the cluster is created, this value is no longer used. If not specified, it is defaulted to the version that is currently recommended by HCP.",
+				Type:             schema.TypeString,
+				Optional:         true,
+				ValidateDiagFunc: validateSemVer,
+				DiffSuppressFunc: func(_, old, new string, _ *schema.ResourceData) bool {
+					return input.NormalizeVersion(old) == input.NormalizeVersion(new)
+				},
+				ForceNew: true,
```
roaks3, line 67: I'm probably just missing context, but wanted to ask: do we plan to support version upgrades for Vault through Terraform?

roaks3 (approved):
LGTM, but was the plan here to merge this into `main`? I think it might be worth considering using a feature branch to merge into, so that this work is not in an incomplete state on the `main` branch. For example, running `go generate` on the code in this PR will generate docs for the Vault cluster before it is implemented, which could be confusing/noisy for other contributors developing off of `main`.

internal/provider/resource_vault_cluster.go
bcmdarroch: After some investigation on the consul side, we've decided to go with `min_vault_version` for consistency with the Consul clusters in the HCP and HCS provider. This way, the field can be used as be both the initial version of the cluster and to force version upgrades.

internal/provider/resource_vault_cluster.go
bcmdarroch: Update: at this moment, `tier` can only be `development`, so making the a computed output for now.

internal/provider/resource_vault_cluster.go
bcmdarroch: Ooh you're right, a customer can't upgrade a cluster yet, so this would always be a ForceNew

PR #95, opened 2021-04-05
[HCS-1831] Remove extraneous protocol from fqdn string if present

aclaygray:

This PR updates the logic for generating the `k8sAuthMethodHost` within the `hcp_consul_agent_helm_config` data source.

As it stands now we always prepend an `https://` protocol to the `kubernetes_endpoint` string provided. This updated logic will strip a leading `https://` from the `kubernetes_endpoint` string and will leverage the existing logic to prepend the protocol again as needed.  This will allow for handling of both cases of this string having a leading `https://` and not.

### :building_construction: Local testing
With the following `main.tf`

```
<PROVIDER set-up>

data "hcp_consul_agent_helm_config" "test" {
  cluster_id = "consul-cluster"
  kubernetes_endpoint = "https://catter.com"
}
```

resulted with the following in `terraform.state` file:
```
...
k8sAuthMethodHost: https://catter.com:443
...
```

The above output is seen with the following `main.tf` as well:

```
<PROVIDER set-up>

data "hcp_consul_agent_helm_config" "test" {
  cluster_id = "consul-cluster"
  kubernetes_endpoint = "catter.com"
}
```

### DIFF
--- a/internal/provider/data_source_consul_agent_helm_config.go
+++ b/internal/provider/data_source_consul_agent_helm_config.go
@@ -190,6 +190,19 @@ func generateHelmConfig(name, datacenter, fqdn string, retryJoin []string, expos
 	// replace any escaped double-quotes with single quotes
 	rj = strings.Replace(rj, "\"", "'", -1)
 
+	// trim off any leading `https://` protocol if present.
+	// this protocol will be prepended as expected when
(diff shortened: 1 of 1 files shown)


roaks3 (approved):
LGTM

PR #96, opened 2021-04-06
HCPE-977: Vault admin token resource

bcmdarroch:

This PR adds the Vault cluster admin token resource. A previous closed PR ([#92](https://github.com/hashicorp/terraform-provider-hcp/pull/92)) added a token data source. But to avoid introducing a vulnerability that had new admin tokens generating on every refresh, I decided to make this a resource in order to track the creation time of the admin token and refresh it only when it is about to expire. In the future this expiration tracking will likely be pushed into the API.

Sample HCL:
```
resource "hcp_hvn" "example" {
  hvn_id          = "test-hvn"
  cloud_provider  = "aws"
  region          = "us-west-2"
}

  cluster_id         = "test-vault-cluster"
  hvn_id             = hcp_hvn.example.hvn_id
}

resource "hcp_vault_cluster_admin_token" "example" {
  cluster_id         = hcp_vault_cluster.example.cluster_id
}
```

Sample `terraform plan` output:
```
# hcp_vault_cluster_admin_token.example will be created
+ resource "hcp_vault_cluster_admin_token" "example" {
    + cluster_id                   = "test-vault-cluster"
    + id                           = (known after apply)
    + token                        = (known after apply)
    + created_at                   = (known after apply)
}
```

- [x] Are there any feature flags that are required to use this functionality? ❌

```
$ make testacc TESTARGS='-run=TestAccVaultClusterAdminToken'

--- PASS: TestAccVaultClusterAdminToken (754.93s)
PASS
```


internal/provider/resource_vault_cluster_admin_token.go
```diff
@@ -42,18 +54,160 @@ func resourceVaultClusterAdminToken() *schema.Resource {
...
+	// If the token already exists, this block verifies if it is close to expiration and should be refreshed.
+	createdAt := d.Get("created_at").(string)
+	if createdAt != "" {
+		log.Printf("[INFO] existing admin token found for Vault cluster (%s) [project_id=%s, organization_id=%s]",
+			clusterID,
+			loc.ProjectID,
+			loc.OrganizationID,
+		)
+
+		// The refresh window starts five minutes before the 6h expiry.
```
jfreda, line 165: Awesome, I was going to suggest this exact thing if it wasn't already implemented 🙌

jfreda (approved):
Looks great! 🚀

internal/provider/resource_vault_cluster_admin_token.go
```diff
@@ -33,6 +40,11 @@ func resourceVaultClusterAdminToken() *schema.Resource {
 				ValidateDiagFunc: validateSlugID,
 			},
 			// computed outputs
+			"created_at": {
+				Description: "The time that the admin token was created.",
+				Type:        schema.TypeString,
+				Computed:    true,
+			},
```
roaks3, line 47: You might already have this on a local branch, but I think we still want the `token` below to be `Sensitive`

jfreda: Ooh yes, I double-checked this for the data source PR but missed it on this one - good catch!

roaks3 (approved):
Code looks good, and logic here makes sense to me to limit the risk of accumulating many valid tokens.

Main thing I came across while testing was that `terraform plan` will still create an additional token (if the previous token was expired), but not save it in state, so the operation results in an orphaned token that will expire in 6 hours. To me, this is better than the alternative of _always_ orphaning tokens when doing a plan or apply, but it does seem like we're bumping up against the edges of what is expected on the Terraform side while using these HCP endpoints. It might be worth revisiting the APIs in the future to determine if they could be adjusted to offer more options here.

bcmdarroch:
Glad you pointed that out @roaks3! I've documented that as a known issue on the resource doc page.

PR #98, opened 2021-04-07
Update changelog for 0.4.0 release

bcmdarroch:

We got Vault resources and Consul cluster VM sizes incoming! 🚀

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,10 @@
 ## 0.4.0 (Unreleased)
 
 FEATURES:
-
 - **New resource** `hcp_vault_cluster` (#97)
 - **New resource** `hcp_vault_cluster_admin_token` (#97)
 
 IMPROVEMENTS:
-
 - all: Log import ID used when an import fails due to parsing (#82)
 - all: Add comment to clarify that Links can be sent in API requests (#82)
 - ci: Add github checks (#90)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
+## 0.4.0 (April 6, 2021)
+
+FEATURES:
+
+- **New resource** `hcp_vault_cluster` (#97)
+- **New resource** `hcp_vault_cluster_admin_token` (#97)
+
+IMPROVEMENTS:
+
+- all: General chores (#82)
```
roaks3, line 10: Oh sorry, should have maybe made it more clear, but I thought we'd break these out into individual items:
```
FIXES:
- all: Ensure context is being passed for all HCP API calls (#82)

IMPROVEMENTS
resource/aws_network_peering: Update comments, docs, and messages to use correct capitalization for network peering (#82)
all: Log import ID used when an import fails due to parsing (#82)
all: Add comment to clarify that Links can be sent in API requests (#82)
resource/aws_network_peering: Update peering wait function to use helper (#82)
```

CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
+## 0.4.0 (April 6, 2021)
```
roaks3, line 1: I think we want `(Unreleased)` here for the date, and then the TeamCity job takes care of replacing with the date

roaks3 (approved):
Approving so that it isn't blocked from merging, but I think this also needs the updates to `examples/provider/provider.tf` for the main example snippet that people are most likely to copy-paste (similar to previous PRs: https://github.com/hashicorp/terraform-provider-hcp/pull/87/files)

PR #100, opened 2021-04-08
HCS-1805-hotfix: Set "computed=true" option for the vm size

riddhi89:
- `size` is an optional field on Consul cluster create. If customers have not specified it in the config, without the "computed=true" option it forces recreation of the cluster.
- Also updated the size desc to include the link with more details.

### DIFF
--- a/internal/provider/resource_consul_cluster.go
+++ b/internal/provider/resource_consul_cluster.go
@@ -119,11 +119,11 @@ func resourceConsulCluster() *schema.Resource {
 				ForceNew:    true,
 			},
 			"size": {
-				// TODO: Add a link to the HCP Consul size details when it is available here - https://cloud.hashicorp.com/pricing/consul#FAQ
-				Description:      "The t-shirt size representation of each server VM that this Consul cluster is provisioned with. Valid option for development tier - `x_small`. Valid options for other tiers - `small`, `medium`, `large`.",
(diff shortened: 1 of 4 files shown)


internal/provider/resource_consul_cluster.go
```diff
@@ -119,11 +119,11 @@ func resourceConsulCluster() *schema.Resource {
 				ForceNew:    true,
 			},
 			"size": {
-				// TODO: Add a link to the HCP Consul size details when it is available here - https://cloud.hashicorp.com/pricing/consul#FAQ
-				Description:      "The t-shirt size representation of each server VM that this Consul cluster is provisioned with. Valid option for development tier - `x_small`. Valid options for other tiers - `small`, `medium`, `large`.",
+				Description:      "The t-shirt size representation of each server VM that this Consul cluster is provisioned with. Valid option for development tier - `x_small`. Valid options for other tiers - `small`, `medium`, `large`. For more details - https://cloud.hashicorp.com/pricing/consul",
 				Type:             schema.TypeString,
 				Optional:         true,
 				ForceNew:         true,
+				Computed:         true,
```
bcmdarroch, line 126: Was concerned that this should really be `Default`, but according to [this issue](https://github.com/hashicorp/terraform/issues/21278), this is totally a valid pattern! 📝

bcmdarroch (approved):

PR #102, opened 2021-04-14
Doc updates

bcmdarroch:
This PR updates the warning messages on resources with sensitive input/output. It also adds a warning about the bug released in v0.4.0 and patched in v0.4.1.


docs/data-sources/aws_transit_gateway_attachment.md
```diff
@@ -20,6 +20,8 @@ data "hcp_aws_transit_gateway_attachment" "test" {
 }
 ```
 
+~> **Security Notice** This resource contains sensitive input. Please see this [list of recommendations](https://www.terraform.io/docs/language/state/sensitive-data.html) for storing sensitive information in Terraform.
+
```
roaks3, line 24: I don't believe the data source actually contains any sensitive inputs, so this message might be confusing

docs/index.md
```diff
@@ -9,12 +9,18 @@ description: |-
 
 The HCP provider provides resources to manage [HashiCorp Cloud Platform](https://cloud.hashicorp.com/) (HCP) resources.
 
+~> **Known Issue** There is an issue with v0.4.0 of the HCP Provider in which existing Consul clusters that do not specify size will be force-recreated on the next Apply. Please upgrade to the patch v0.4.1 to avoid this issue.
+
```
roaks3, line 13: I think I understand the intent of this message, but from the user's perspective it may be a little weird. What we will end up with is a `v0.4.2` when this gets released, which will mention that the user should use `v0.4.1`, because `v0.4.0` has an issue. But then when the user selects `v0.4.0` or `v0.4.1` from the version selector, they will not see any messages related to known issues. I'm wondering if we will instead need to rely on a more generic message if we really need something like this, eg. `Please refer our [CHANGELOG]() for critical fixes`

docs/index.md
```diff
@@ -9,12 +9,18 @@ description: |-
...
 ## Authenticating with HCP
 
 The HCP provider supports authentication via a Client ID and a Client Secret. The [authentication guide](guides/auth.md) describes how to obtain client credentials.
 
 ## Example Usage
 
+Most HCP resources have a dependency on the Hashicorp Virtual Network (HVN) resource.
```
roaks3, line 20: Nit: IMO this is so fundamental to Terraform usage that it would be extraneous information for the average user (I'm not aware of any other providers that include a note like this), and for Terraform beginners, we probably need to be directing them to more comprehensive guides to be brought up to speed

templates/resources/consul_cluster_root_token.md.tmpl
```diff
@@ -0,0 +1,20 @@
...
+{{ .Description | plainmarkdown | trimspace | prefixlines "  " }}
+---
+
+# {{.Name}} `{{.Type}}`
+
+~> **Security Notice:** Please see this [list of recommendations](https://www.terraform.io/docs/language/state/sensitive-data.html) for storing sensitive information in Terraform.
+
+{{ .Description | trimspace }}
+
+The cluster root token resource is the token used to bootstrap the cluster's ACL system. Using this resource to create a new root token for a cluster will invalidate the consul root token accessor id and Consul root token secret id properties of the cluster.
```
roaks3, line 14: This line seems like a duplicate of the `Description` above?

templates/resources/consul_cluster_root_token.md.tmpl
```diff
@@ -0,0 +1,20 @@
...
+
+## Example Usage
+
+{{ tffile "examples/resources/hcp_vault_cluster_admin_token/resource.tf" }}
```
roaks3, line 18: I think this one needs to be `hcp_consul_cluster_root_token` instead of `hcp_vault_cluster_admin_token`

roaks3 (changes requested):
Mostly looks good, but had a few comments, and I think the Consul root token docs need a few fixes

docs/index.md
bcmdarroch: I like the idea of adding the generic message, I can do that. @xargs-P wants to keep the specific 0.4.0 warning in for 30 days, then we'll remove it. We're adding the same 30-day warning to the CLI (which will show up when the provider initializes). We want to be extra noisy about that issue, since it could result in data loss.

docs/data-sources/aws_transit_gateway_attachment.md
bcmdarroch: Oops you're right, thought the ARN might be an output

roaks3 (approved):

PR #111, opened 2021-04-22
added self_link to hvn, peering and tgw attachment

smaant:

Added `self_link` attribute to hvn, peering and tgw attachment resources and datasources.

### DIFF
--- a/internal/provider/resource_hvn_test.go
+++ b/internal/provider/resource_hvn_test.go
@@ -5,8 +5,10 @@ import (
 	"fmt"
 	"testing"
 
+	sharedmodels "github.com/hashicorp/hcp-sdk-go/clients/cloud-shared/v1/models"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+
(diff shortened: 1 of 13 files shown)


nombiezinja (commented):
LGTM but am unfamiliar with the repo so will let Brenna do the approval!

bcmdarroch (commented):
Looks good! My only ask: could you update the acceptance tests too? https://github.com/hashicorp/terraform-provider-hcp/blob/main/internal/provider/resource_hvn_test.go

smaant:
@bcmdarroch good point, thank you for noticing! updated the test

internal/provider/resource_hvn_test.go
```diff
@@ -102,6 +106,41 @@ func testAccCheckHvnExists(name string) resource.TestCheckFunc {
 	}
 }
 
+func testSelfLink(name string, expectedID, expectedType string) resource.TestCheckFunc {
```
bcmdarroch, line 109: thanks for adding this! probably something we could share with future tests :)

smaant: yes, this is how I saw it too

bcmdarroch (approved):

PR #115, opened 2021-04-28
HCCP91 hvn routes data source

nombiezinja:

Jira: https://hashicorp.atlassian.net/browse/HCCP-90

Adds the `hcp_hvn_route` data source.; another PR will follow to add the resource.

This data source is the first in the provider to use self_link (change introduced in https://github.com/hashicorp/terraform-provider-hcp/pull/111) as references in lieu of ids (e.g. hvn_id)

Example output from read:
Screen Shot 2021-04-30 at 1 03 42 PM


internal/provider/data_source_hvn_route.go
```diff
@@ -0,0 +1,98 @@
...
+		Schema: map[string]*schema.Schema{
+			// Required inputs
+			"hvn": {
+				Description: "The `self_link` of the HashiCorp Virtual Network (HVN).",
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+			"destination_cidr": {
+				Description: "The destination CIDR of the HVN route",
+				Type:        schema.TypeString,
```
nombiezinja, line 29: Thinking of introducing CIDR validation here, but wanted to see what the team thinks before adding it.

smaant: I'm not sure it's necessary, backend already does the validation

internal/provider/data_source_hvn_route.go
```diff
@@ -0,0 +1,98 @@
...
+	log.Printf("[INFO] Reading HVN route for HVN (%s) with destination_cidr=%s ", hvn, destination)
+	route, err := clients.ListHVNRoutes(ctx, client, hvnLink.ID, destination, "", "", loc)
+	if err != nil {
+		return diag.Errorf("unable to retrieve HVN route for HVN (%s) with destination_cidr=%s: %v",
+			hvn, destination, err)
+	}
+
+	// ListHVNRoutes call should return 1 and only 1 HVN route.
+	if len(route) != 1 {
+		return diag.Errorf("Unexpected number of HVN routes returned: %d", len(route))
```
smaant, line 83: I think if route not found (len == 0) then we can return a more specific error, like "route with destination CIDR X not found"

internal/provider/data_source_hvn_route.go
```diff
@@ -0,0 +1,98 @@
...
+			},
+			// Computed outputs
+			"self_link": {
+				Description: "A unique URL identifying the HVN route.",
+				Type:        schema.TypeString,
+				Computed:    true,
+			},
+			"target": {
+				Description: "The `self_link` identifying the target of the HVN route.",
+				Type:        schema.TypeString,
```
smaant, line 40: I think we should make it an object with a single `self_link` field inside for now, otherwise we block ourselves from extending it if we need to in the future

internal/provider/resource_hvn_route.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	if err := d.Set("self_link", selfLink); err != nil {
+		return err
+	}
+
+	// Set self_link identifying the target of the HVN route.
+	var targetLink string
+
+	switch route.Target.HvnConnection.Type {
```
smaant, line 31: I think you can just use this type to build url, not necessary to have cases for everything

nombiezinja: i was thinking we need to handle the case when target has an unknown type but you're right, that's unnecessary since (a) unlikely to happen and (b) will not raise an error since it is still just a string

internal/provider/data_source_hvn_route.go
```diff
@@ -0,0 +1,104 @@
...
+			"self_link": {
+				Description: "A unique URL identifying the HVN route.",
+				Type:        schema.TypeString,
+				Computed:    true,
+			},
+			"target": {
+				Description: "The target of the HVN route.",
+				Type:        schema.TypeMap,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
```
smaant, line 42: @bcmdarroch do you know if we can extend this in the future, say make it more complex than just a `string: string` kind of object?

nombiezinja: Screen Shot 2021-04-28 at 4 14 58 PM
it does seem like we can keep nesting this, but will let Brenna give a better answer since I'm new to writing providers

smaant: what I meant is now we are going to have it like:
```
target = {
  self_link = "url"
}
```
what if in the future we would want to make it:
```
target = {
  self_link = "url"
  some_number = 1
  some_object = {
     foo = "bar"
  }
}
```
how would be transition from first to second? Is the second is even possible in TF?

nombiezinja: looked at docs_as far as i can tell_, i don't think so, the elements must be homogenous

internal/provider/data_source_hvn_route.go
```diff
@@ -0,0 +1,104 @@
...
+	log.Printf("[INFO] Reading HVN route for HVN (%s) with destination_cidr=%s ", hvn, destination)
+	route, err := clients.ListHVNRoutes(ctx, client, hvnLink.ID, destination, "", "", loc)
+	if err != nil {
+		return diag.Errorf("unable to retrieve HVN route for HVN (%s) with destination_cidr=%s: %v",
+			hvn, destination, err)
+	}
+
+	// ListHVNRoutes call should return 1 and only 1 HVN route.
+	if len(route) > 1 {
+		return diag.Errorf("Unexpected number of HVN routes returned for destination_cidr=%s: %d", destination, len(route))
```
daneharrigan, line 86: Can you help me understand this error? cloud-network removed it's overlapping CIDR validation, right? Could a customer create two HVN Routes with the same CIDR?

internal/provider/data_source_hvn_route.go
bcmdarroch: Generally, the provider should duplicate any straightforward input validation (i.e. doesn't require an API call) so requests fail early. Just remembered terraform provider SDK provides a [CIDR helper](https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/validation/network.go#L84) 🙌

internal/provider/data_source_hvn_route.go
nombiezinja: Customers can't create HVN routes with identical CIDRs, however in the case this changes in the future (however unlikely), we should still handle the case of ListHVNRoutes returning more than 1 route

internal/provider/data_source_hvn_route.go
```diff
@@ -0,0 +1,104 @@
...
+			Default: &hvnRouteDefaultTimeout,
+		},
+		Schema: map[string]*schema.Schema{
+			// Required inputs
+			"hvn": {
+				Description: "The `self_link` of the HashiCorp Virtual Network (HVN).",
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+			"destination_cidr": {
```
roaks3, line 27: Just curious about the required input fields here: why was `destination_cidr` chosen as opposed to some sort of identifier, like the `ID`? Are we expecting this to be more user-friendly?

internal/provider/data_source_hvn_route.go
roaks3: I don't know much about this either, but looking at an example like this, it might be possible to do? https://github.com/hashicorp/terraform-provider-aws/blob/4bd169a4786a8a35674fa7b93b4dadc17d8d9090/aws/resource_aws_eks_cluster.go#L165-L214

internal/provider/data_source_hvn_route.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+}
+
+func dataSourceHVNRouteRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*clients.Client)
+
+	hvn := d.Get("hvn").(string)
+	var hvnLink *sharedmodels.HashicorpCloudLocationLink
+
+	hvnLink, err := parseLinkURL(hvn, "hashicorp.network.hvn")
```
roaks3, line 66: We have a `HvnResourceType` constant that you can use here instead of the raw string

internal/provider/data_source_hvn_route.go
nombiezinja: thanks for the example! perhaps `type` should be `schema.TypeList` then, to allow more extensibility

xargs-P:
@nombiezinja do you have example output for the data source with more than one route? Is it just a list?
Also I do not think we support route priority at the moment, but if so, how would priority be reflected in the list?

internal/provider/data_source_hvn_route.go
smaant: we won't have routes with identical CIDRs, there's no way to know how to route traffic in this case

internal/provider/data_source_hvn_route.go
smaant: it's similar to https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/route, allows finding what target is associated with specific CIDR

smaant:
@xargs-P this datasource is intended to return only single route (there could be only one route for each CIDR)

bcmdarroch (commented):
Nice work on this @nombiezinja! Definitely a complex challenge for your first provider contribution.

I had some comments, and a proposal. This data source could be made more flexible if we turned it into something like [AWS's subnet IDs data source](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/subnet_ids), which always returns a list.

I would like to know more about how users might use this data source. Will they usually need the list, or just one route, or both? I'm leaning towards having both a singular and plural version of this data source for the greatest amount of flexibility. 

Happy to make some time to chat about this over zoom!

internal/provider/data_source_hvn_route.go
smaant: maybe full link info (id/type/location/etc) maybe when we introduce not just `hvn_connection` types of targets something else, I just don't want to lock us into not being able to extend it if we need it

xargs-P:
> @xargs-P this datasource is intended to return only single route (there could be only one route for each CIDR)

Got it @smaant . Related to priority, are we tracking? If so, is that expressed in the data source?

smaant:
@xargs-P actually just discussed with Ti, we would take Brenna's suggestion and make it be able to return multiple routes. For routes priority is always from more specific cidr to less specific, this is a standard behavior for route tables:
 - https://docs.aws.amazon.com/vpc/latest/userguide/VPC_Route_Tables.html#route-table-routes
 - https://docs.microsoft.com/en-us/azure/virtual-network/virtual-networks-udr-overview#how-azure-selects-a-route

internal/provider/data_source_hvn_route.go
nombiezinja: spoke to @smaant , we will make this field `target_link` instead to keep the structure flat and simple for now. in the future we may add `target` if we need to

internal/provider/data_source_hvn_route.go
daneharrigan: @smaant linked me to [this DB trigger](https://github.com/hashicorp/cloud-network/blob/b39b5a6d46b4caf1a42a6132e821cf66d1b2b62a/models/migrations/20210323110208_routes_uniqueness_check.up.sql). I was missing what allowed overlapping, but enforced unique CIDRs.

nombiezinja:
> @xargs-P actually just discussed with Ti, we would take Brenna's suggestion and make it be able to return multiple routes. For routes priority, it is always from more specific cidr to less specific, this is a standard behavior for route tables:
> 
> * https://docs.aws.amazon.com/vpc/latest/userguide/VPC_Route_Tables.html#route-table-routes
> * https://docs.microsoft.com/en-us/azure/virtual-network/virtual-networks-udr-overview#how-azure-selects-a-route

Just looked into the topic of data source returning multiple values a bit more; it looks like the existing pattern for writing providers is to have one data source for singular entity, and one for plurals. examples:

https://github.com/hashicorp/terraform-provider-aws/blob/main/aws/data_source_aws_outposts_outpost.go
https://github.com/hashicorp/terraform-provider-aws/blob/main/aws/data_source_aws_outposts_outposts.go

https://github.com/hashicorp/terraform-provider-aws/blob/main/aws/data_source_aws_security_group.go
https://github.com/hashicorp/terraform-provider-aws/blob/main/aws/data_source_aws_security_groups.go

seems like it might make sense to keep the current data source the way it is, and keep `destination_cidr` a required param. when the need arises to query all routes for an HVN, we can create another data source for plural hvn routes, following existing patterns like the examples above

xargs-P:
Got it 👍 @nombiezinja

xargs-P (approved):
Thanks for the examples 👌

@bcmdarroch good for you?

roaks3:
@xargs-P Brenna is out today, but she did ping me before she signed off yesterday to say that the singular version of this data source is good for her if it's good for you 😄 . So sounds like we're happy to move forward with the data source the way it is in this PR

internal/provider/data_source_hvn_route.go
```diff
@@ -0,0 +1,102 @@
...
+			"hvn": {
+				Description: "The `self_link` of the HashiCorp Virtual Network (HVN).",
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+			"destination_cidr": {
+				Description:      "The destination CIDR of the HVN route",
+				Type:             schema.TypeString,
+				Required:         true,
+				ValidateDiagFunc: validateIsStartOfPrivateCIDRRange,
```
nombiezinja, line 31: Added validation for `destination_cidr`. Originally I was planning to use the IsCIDR helper as @bcmdarroch had suggested, and use `validation.All` following the example here(https://github.com/hashicorp/terraform-provider-aws/blob/main/aws/resource_aws_eks_cluster.go#L133). However, I went with a custom validator for 2 reasons: 

1-  terraform-provider-hcp only uses `ValidateDiagFunc`, and `validation.All` would not work 
2- `ValidateFunc` is deprecated but most importantly
3- HCP create HVN route endpoint validation (https://github.com/hashicorp/cloud-network/blob/master/service/network/private/service_create_hvn_route.go#L218) uses the validation rule here: https://github.com/hashicorp/cloud-sdk/blob/master/validation/rule_cidrblock.go#L0-L1; this validation is reused in several placess acros HCP endpoints taking a CIDR parameter. `validateIsStartOfPrivateCIDRRange` validates based on the same rules: 

a) CIDR is valid 
b) CIDR is private CIDR block
c) Address is at beginning of CIDR range

smaant: I would suggest to remove validation here (and probably in other places too), the least duplicated logic we have in Tf provider the easier it will be to maintain. Basic request validation will be done by the backend very quickly and won't impact UX of the TF provider.

roaks3: Thanks for adding client-side validation for this! Looks great, and agreed that this validator should be used for our other CIDRs

roaks3 (approved):
LGTM

smaant (approved):

internal/provider/data_source_hvn_route.go
roaks3: Oops, just missed @smaant 's response on this. IMO it is reasonable to have this be validated client-side (I'm assuming it is generic and unlikely to change over time), and another slight benefit is that the user will get feedback on the `plan` step instead of seeing it later during `apply`. But otherwise, I do agree that this could be deferred to the backend and the experience would be similar.

nombiezinja: Replaced the custom validator with `IsCIDR` as we are looking to gain some traction on this work, and merged this PR

Also spoke to provider team (@bcmdarroch & @roaks3 ) as well as some Terraform practitioners; my recommendation is to ensure the custom validator is added back at some point. The aforementioned 3 rules are rudimentary and are unlikely to be removed from the cloud-sdk validation package, hence drift is less of a concern. If the validation rules in cloud-sdk were to change, we are unlikely to remove rules, and only likely to add additional rules, in which case the custom validator covering 3 existing rules will be a good idea in terms of improving the practitioner experience for this provider.

PR #116, opened 2021-05-03
Make peer_vpc_cidr_block(peering resource) and destination_cidrs(TGW attachment resource) optional 

nombiezinja:
https://hashicorp.atlassian.net/browse/HCCP-91

peer_vpc_cidr_block(peering resource) and destination_cidrs(TGW attachment resource) are optional instead of mandatory.

Prepartory PR for HCCP-91 HVN route resource.

### DIFF
--- a/internal/provider/resource_aws_network_peering.go
+++ b/internal/provider/resource_aws_network_peering.go
@@ -66,14 +66,14 @@ func resourceAwsNetworkPeering() *schema.Resource {
 					return strings.ToLower(old) == strings.ToLower(new)
 				},
 			},
+			// Optional inputs
 			"peer_vpc_cidr_block": {
 				Description:  "The CIDR range of the peer VPC in AWS.",
 				Type:         schema.TypeString,
-				Required:     true,
+				Optional:     true,
 				ForceNew:     true,
 				ValidateFunc: validation.IsCIDR,
 			},
-			// Optional inputs
 			"peering_id": {
 				Description:      "The ID of the network peering.",
 				Type:             schema.TypeString,
--- a/docs/resources/aws_network_peering.md
+++ b/docs/resources/aws_network_peering.md
@@ -53,13 +53,13 @@ resource "aws_vpc_peering_connection_accepter" "peer" {
 
 - **hvn_id** (String) The ID of the HashiCorp Virtual Network (HVN).
 - **peer_account_id** (String) The account ID of the peer VPC in AWS.
-- **peer_vpc_cidr_block** (String) The CIDR range of the peer VPC in AWS.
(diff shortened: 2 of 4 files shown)


internal/provider/resource_aws_network_peering.go
```diff
@@ -66,14 +66,14 @@ func resourceAwsNetworkPeering() *schema.Resource {
 					return strings.ToLower(old) == strings.ToLower(new)
 				},
 			},
+			// Optional inputs
 			"peer_vpc_cidr_block": {
 				Description:  "The CIDR range of the peer VPC in AWS.",
 				Type:         schema.TypeString,
-				Required:     true,
+				Optional:     true,
```
bcmdarroch, line 73: Can the API provide a default if the CIDR block is not set? We've run into issues in the past where combining `Optional` and `ForceNew` has led to unintended deletion of resources. That shouldn't be an issue here, since this was a required field and therefore should exist on all peering resources. Still, would be ideal to include `Computed` if the API can provide a default.

nombiezinja: This PR will result in the provider being in an interim state for these 2 resources, the eventual state that we are aiming for is releasing a major version that removes these 2 parameters altogether, delegating the creation of the routes to the HVN route APIs completely. At this moment the RPCs for create peering and create TGW attachment has stopped validating the existence of these parameters, and either 

(a) creates an HVN route for the peering/TGW attachment if `destination_cidr`/`peering_vpc_cidr_block` is present or 
(b) creates the TGW attachment and peering without an HVN route if `destination_cidr`/`peering_vpc_cidr_block` is not passed

I wasn't aware of how `ForceNew` could accidentally result in the deletion of an existing resource, in your opinion as the provider repo owner, do you think we should remove it for this interim state?

smaant: As Ti mentioned, we are planning to deprecate/remove this field whatsoever (deprecate and eventually remove for peerings and remove right now for tgw attachments). We can use empty string as default if necessary.

smaant (approved):

smaant:
Discussed with @bcmdarroch and @nombiezinja and realized that it might be not ideal to recreate peering if user used to have a CIDR in the peering definition and removed it. We decided to find a way to check if cidr got erased (used to have some value and empty now) then trigger removing that HVN route, if cidr changed then return an error that this behavior no longer supported and they need to use HVN routes instead. In both ways we shouldn't delete the peering.

nombiezinja:
Spoke with @smaant and @bcmdarroch, in addition to the change discussed above, we will make this PR part of the feature branch for the HVN route resource. Renamed branch to HCCP-91-HVN-route-resource, closing this in favour of a new PR with the newly renamed branch

PR #117, opened 2021-05-07
Add beta banner to HVN route data source doc

nombiezinja:
This adds a banner to the docs for HVN route data source, since the feature is still in private beta and behind a feature flag, and the provider relevant work is still under development. When the next version of the provider is published, these pages will be visible to all users, so this banner will help prevent confusion.


bcmdarroch (approved):
Perfect ✨

PR #118, opened 2021-05-07
Set ForceNew to false for public_endpoint

bplotnick:
I don't believe changing public_endpoint requires a recreation of the cluster. At least it doesn't in the UI


bcmdarroch:
Hi @bplotnick, thanks for your contribution! Unfortunately there are some further changes required before this can merge - mainly, implementing Update for this resource, which must call [this endpoint](https://github.com/hashicorp/hcp-sdk-go/blob/main/clients/cloud-vault-service/preview/2020-11-25/client/vault_service/vault_service_client.go#L64). You're welcome to make those changes, but otherwise I've asked our cloud Vault team for help with this.

bcmdarroch:
Completed in #152

PR #121, opened 2021-05-12
Remove TGW beta banner from docs for GA

nombiezinja:
This removes the banner to the docs for transit gateway attachment resources, in preparation for the feature going GA


templates/resources/aws_transit_gateway_attachment.md.tmpl
```diff
@@ -7,8 +7,6 @@ description: |-
 
 # {{.Type}} ({{.Name}})
 
--> **Note:** This feature is currently in private beta. If you would like early access, please [contact our sales team](https://www.hashicorp.com/contact-sales).
```
bcmdarroch, line 10: Let's delete this file too and regen. Without the banner, there's no need for the custom template.

bcmdarroch (approved):

PR #122, opened 2021-05-17
HCCP-91 HVN route resource and import

nombiezinja:
Ticket: https://hashicorp.atlassian.net/browse/HCCP-91

## Description 

Adds the hcp_hvn_route resource and import. The implementation follows examples of the existing resources, with the exception of using `self_link` to refer to HVNs and Target resources, rather than the ID of the HVN or Target.

A simple example of what the resource looks like for a TGW attachment:

```
resource "hcp_hvn_route" "tgwroute" {
  hvn_link         = hcp_hvn.main.self_link
  destination_cidr = aws_vpc.example.cidr_block
  hvn_route_id     = "exampletgwhvnroute"
  target_link      = hcp_aws_transit_gateway_attachment.example.self_link
}
```

A simple example of what the resource looks like for a peering: 

```
resource "hcp_hvn_route" "peering-route" {
  hvn_link         = hcp_hvn.test-hvn.self_link
  destination_cidr = "192.168.0.0/20"
  hvn_route_id     = "examplepeeringroute"
  target_link      = hcp_aws_network_peering.example.self_link
}
```

## To Test

The example included here can be used to test the functionality of this resource. 
Please note that 
- if creating a route for a TGW attachment, the AWS config cannot be the same account as the HVN

## Additional notes

- ~This PR also makes the fields `peer_vpc_cidr_block` and `destination_cidr` optional. This is an interim solution, as based on discussions with product we have reached consensus that we will make the breaking changes in the next release by simply removing these fields. However, for the purpose of this PR these fields need to be optional, otherwise submitting a value during creation of peering/tgw attachment will result in a route having been created for these resources, causing an error during later HVN route creation~

- This PR does not include acceptance tests, as they will be in a subsequent PR, along with the acceptance tests for TGW attachment resrouce 

- A follow-up PR will also remove `peer_vpc_cidr_block` and `destination_cidr` fields from the peering and TGW attachment resource, as aforementioned. It will also make the user-settable ID field required for both of these resrouces. It is not included in this PR as the provider team has agreed that it will make this review easier.


examples/resources/hcp_hvn_route/resource.tf
```diff
@@ -0,0 +1,59 @@
+provider "aws" {
```
nombiezinja, line 1: The example here is for TGW attachment, I was thinking perhaps we could also include an example for creating an hvn route for a peering in this same file, but not sure if that would be sensible, wanted to hear what the team thinks

internal/provider/data_source_hvn_route.go
```diff
@@ -52,6 +52,7 @@ func dataSourceHVNRoute() *schema.Resource {
 				Type:        schema.TypeString,
 				Computed:    true,
 			},
+			// TODO add updated_at after public API returns updated_at for HVN route
```
nombiezinja, line 55: This work has been recently merged (https://github.com/hashicorp/cloud-network/pull/526), we will need to release a new version of the SDK in order to leverage the change, my suggestion is to do it in a follow up PR

internal/provider/resource_hvn_route.go
```diff
@@ -1,14 +1,275 @@
...
+				Description: "The state of the HVN route.",
+				Type:        schema.TypeString,
+				Computed:    true,
+			},
+			"created_at": {
+				Description: "The time that the HVN route was created.",
+				Type:        schema.TypeString,
+				Computed:    true,
+			},
+			// TODO add updated_at after public API returns updated_at for HVN route
```
nombiezinja, line 81: Please see https://github.com/hashicorp/terraform-provider-hcp/pull/122/files#r633792131

examples/resources/hcp_hvn_route/resource.tf
bcmdarroch: Yeah I'm fine with either both examples or just peering, since I think that's slightly more commonly used.

nombiezinja: Added peering to example!

internal/provider/resource_hvn_route.go
```diff
@@ -1,14 +1,290 @@
...
+			log.Printf("[WARN] HVN route for HVN (%s) with destination CIDR %s not found, removing from state", hvnLink.ID, destination)
+			d.SetId("")
+			return nil
+		}
+
+		return diag.Errorf("unable to retrieve HVN route for HVN (%s) with destination CIDR %s: %v", hvnLink.ID, destination, err)
+	}
+
+	if len(route) != 1 {
+		return diag.Errorf("Unexpected number of HVN route returned when waiting for route with destination CIDR of %s for HVN (%s) to be Active", destination, hvnLink.ID)
```
bcmdarroch, line 239:
```suggestion
		return diag.Errorf("Unexpected number of HVN routes returned when waiting for route with destination CIDR of %s for HVN (%s) to be Active", destination, hvnLink.ID)
```

bcmdarroch (commented):
This looks good! Left a few comments sprinkled throughout - only major one about dropping the cidr block. 

I'm hoping to do another test run tomorrow - accidentally kicked it off on my env with routes feature flag turned off. 🙈 Are acceptance tests planned for a follow-up? I can help with those, since we don't have any multi-provider test examples yet.

internal/provider/resource_aws_network_peering.go
```diff
@@ -66,14 +66,14 @@ func resourceAwsNetworkPeering() *schema.Resource {
 					return strings.ToLower(old) == strings.ToLower(new)
 				},
 			},
+			// Optional inputs
```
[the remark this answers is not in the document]
nombiezinja: Formerly we agreed to keep that change out of this PR but happy to comply whatever makes sense to you. I am going on call tomorrow so will defer to @antonakv for handling the complete removal of these fields and adding the ID 

I think making another feature branch and pointing this branch to the feature branch makes sense, but will see how Anton would like to handle it.

internal/provider/resource_hvn_route.go
```diff
@@ -1,14 +1,290 @@
...
+	if err != nil {
+		if clients.IsResponseCodeNotFound(err) {
+			return diag.Errorf("unable to create route for HVN (%s) with the destination CIDR of %s, target (%s) does not exist ", hvnLink.ID, destination, target)
+		}
+		return diag.Errorf("unable to create route for HVN (%s) with the destination CIDR of %s, unable to check for existence of target (%s)", hvnLink.ID, destination, target)
+	}
+
+	targetLink.Location.Region = retrievedHvn.Location.Region
+
+	hvnRouteParams := network_service.NewCreateHVNRouteParams()
```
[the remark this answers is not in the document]
nombiezinja: That makes sense. I was following example here: https://github.com/hashicorp/terraform-provider-hcp/blob/HCCP-91-HVN-route-resource/internal/provider/resource_aws_network_peering.go#L160. Given that some resource functions seem to use a client wrapper while others dont, I noted the existing pattern seems to be `extract to helper when it is to be reused elsewhere". Given create is not used elsewhere for HVN routes, I was thinking it makes sense to not use a client wrapper, like the example in peering, what do you think?

internal/provider/resource_hvn_route.go
```diff
@@ -101,11 +104,14 @@ func resourceHvnRouteCreate(ctx context.Context, d *schema.ResourceData, meta in
 
 	target := d.Get("target_link").(string)
 	var targetLink *sharedmodels.HashicorpCloudLocationLink
-	// TODO this needs to handle more than just peering type
 	targetLink, err = buildLinkFromURL(target, PeeringResourceType, loc.OrganizationID)
+	if err != nil {
```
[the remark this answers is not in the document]
nombiezinja: Good call,added comment and improved error return for this as well

docs/resources/hvn_route.md
```diff
@@ -0,0 +1,139 @@
...
+subcategory: ""
+description: |-
+  The HVN route resource allows you to manage an HVN route.
+---
+
+# hcp_hvn_route (Resource)
+
+The HVN route resource allows you to manage an HVN route.
+
+## Example Usage
```
smaant, line 13: I think we should limit example to either TGW or peering, not both for simplicity. Peering would probably the best choice just because it is the most common type of connection and it's much simple.

docs/resources/hvn_route.md
```diff
@@ -0,0 +1,139 @@
...
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- **destination_cidr** (String) The destination CIDR of the HVN route
+- **hvn** (String) The `self_link` of the HashiCorp Virtual Network (HVN).
+- **hvn_route_id** (String) The ID of the HVN route.
+- **target_link** (String) A unique URL identifying the target of the HVN route.
```
smaant, line 111: we should specify here what targets are supported

internal/clients/hvn_route.go
```diff
@@ -28,3 +31,73 @@ func ListHVNRoutes(ctx context.Context, client *Client, hvnID string,
...
+// the LIST endpoint
+func hvnRouteRefreshState(ctx context.Context, client *Client, destination string, hvnID string, loc *sharedmodels.HashicorpCloudLocationLocation) resource.StateRefreshFunc {
+	return func() (interface{}, string, error) {
+		route, err := ListHVNRoutes(ctx, client, hvnID, destination, "", "", loc)
+		if err != nil {
+			return nil, "", err
+		}
+
+		if len(route) != 1 {
+			return nil, "", fmt.Errorf("Unexpected number of HVN route returned when waiting for route with destination CIDR of %s for HVN (%s) to be Active", destination, hvnID)
```
smaant, line 75: I think error message isn't correct here

internal/provider/resource_aws_network_peering.go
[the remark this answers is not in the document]
smaant: let's probably keep this PR focused on HVN routes only and do the rest in a separate PR because it would require updating docs, examples, tests, etc

internal/provider/resource_hvn_route.go
```diff
@@ -1,14 +1,293 @@
...
+	retrievedHvn, err := clients.GetHvnByID(ctx, client, loc, hvnLink.ID)
+	if err != nil {
+		if clients.IsResponseCodeNotFound(err) {
+			return diag.Errorf("unable to find the HVN (%s) for the HVN route", hvnLink.ID)
+		}
+
+		return diag.Errorf("unable to check for presence of an existing HVN (%s): %v", hvnLink.ID, err)
+	}
+
+	if retrievedHvn.State != networkmodels.HashicorpCloudNetwork20200907NetworkStateSTABLE {
```
smaant, line 127: let's please remove this, including checking HVN existence - this is way to much knowledge for the TF provider which is suppose to be a thin client. Especially given that this resource takes an `hvn_link` which is almost guaranty that it won't be just hardcoded but rathe would have an hvn resource/data-source next to it.

internal/provider/resource_hvn_route.go
```diff
@@ -1,14 +1,293 @@
...
+			return diag.Errorf("unable to check for presence of an existing route for HVN (%s) with the destination CIDR of %s: %v", hvnLink.ID, destination, err)
+		}
+
+		return diag.Errorf("an HVN route with destination=%s, hvn_id=%s and project_id=%s already exists - to be managed via Terraform this resource needs to be imported into the state. Please see the resource documentation for hcp_hvn_route for more information", destination, hvnLink.ID, loc.ProjectID)
+	} else {
+		log.Printf("[INFO] HVN route with destination CIDR of %s for HVN (%s) not found, proceeding with HVN route create", destination, hvnLink.ID)
+	}
+
+	// Check if the target exists.
+	switch targetLink.Type {
```
smaant, line 146: same, let's please delete it, it's not a concern of TF provider to do such a business-logic specific validation

internal/provider/resource_hvn_route.go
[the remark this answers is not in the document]
smaant: agree, it makes sense to move a bunch of this code into `clients.CreateHVNRoute`

internal/provider/resource_hvn_route.go
smaant: also, this validation won't happen on `plan` as far as I understand, so there's absolutely no benefits for doing this validation here comparing to calling backend and receiving error from it

internal/provider/resource_hvn_route.go
```diff
@@ -1,14 +1,293 @@
...
+	destination := d.Get("destination_cidr").(string)
+	loc := &sharedmodels.HashicorpCloudLocationLocation{
+		OrganizationID: client.Config.OrganizationID,
+		ProjectID:      client.Config.ProjectID,
+	}
+
+	route, err := clients.ListHVNRoutes(ctx, client, hvnLink.ID, destination, "", "", loc)
+	if err != nil {
+		if clients.IsResponseCodeNotFound(err) {
+			log.Printf("[WARN] HVN route for HVN (%s) with destination CIDR %s not found, removing from state", hvnLink.ID, destination)
```
smaant, line 233: @bcmdarroch is removing datasource from the state is an idiomatic thing to do in this case?

smaant: should we just return an error and fail tf apply? otherwise we just continue operations hoping that user would read output

internal/provider/resource_hvn_route.go
```diff
@@ -1,14 +1,293 @@
...
+	}
+
+	if len(route) != 1 {
+		return diag.Errorf("Unexpected number of HVN routes returned when waiting for route with destination CIDR of %s for HVN (%s) to be Active: %d", destination, hvnLink.ID, len(route))
+	}
+
+	// The HVN route failed to provision properly so we want to let the user know and
+	// remove it from state.
+	if route[0].State == networkmodels.HashicorpCloudNetwork20200907HVNRouteStateFAILED {
+		log.Printf("[WARN] HVN route for HVN (%s) with destination CIDR %s failed to provision, removing from state", hvnLink.ID, destination)
```
smaant, line 248: @bcmdarroch same question here, why don't we just fail the tf apply?

internal/provider/resource_hvn_route.go
```diff
@@ -1,14 +1,293 @@
...
+	hvnLink, err := buildLinkFromURL(hvn, HvnResourceType, loc.OrganizationID)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	log.Printf("[INFO] Deleting HVN route (%s)", routeID)
+	resp, err := clients.DeleteHVNRouteByID(ctx, client, hvnLink.ID, routeID, loc)
+
+	if err := clients.WaitForOperation(ctx, client, "delete HVN route", loc, resp.Operation.ID); err != nil {
+		if strings.Contains(err.Error(), "execution already started") {
```
smaant, line 281: where do we return this error?

smaant (changes requested):

internal/provider/resource_aws_network_peering.go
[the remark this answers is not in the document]
bcmdarroch: Agreed, separate PRs for each change would be best. So you'll move these changes to peering and TGW into a new PR?

docs/resources/hvn_route.md
bcmdarroch: Yeah peering sounds good.

internal/provider/resource_hvn_route.go
bcmdarroch: I agree that checking the HVN state isn't necessary. However, checking for the existence of the resource is an established pattern in all our other resources, as it allows us to provide additional information in the logs, like:
```
2021-05-11T11:15:14.170-0700 [INFO]  plugin.terraform-provider-hcp: 2021/05/11 11:15:14 [INFO] HVN (my-favorite-hvn) not found, proceeding with create: timestamp=2021-05-11T11:15:14.170-0700
2021-05-11T11:15:14.170-0700 [INFO]  plugin.terraform-provider-hcp: 2021/05/11 11:15:14 [INFO] Creating HVN (my-favorite-hvn): timestamp=2021-05-11T11:15:14.170-0700
```

internal/provider/resource_hvn_route.go
bcmdarroch: Yes, this is a pattern. If it the API does not return the route, that means it was deleted and thus TF needs to update its state to reflect that. Setting the ID to nil in the line below is the same as deleting it from the state. The error in L238 should then be displayed, so the user will know there was a change in the state.

internal/provider/resource_hvn_route.go
bcmdarroch: Same answer, this updates the TF state during a `terraform refresh`.

internal/provider/resource_hvn_route.go
bcmdarroch: Looks like this is a pattern from the peering resource, which had these comments:
```
		// If the HVN has already been deleted
		// TODO: Peerings can be deleted automatically by the network monitor workflow ]
		// and when deleting the HVN is deleted can cause an already started error.
		// It would be nicer if we could return a better error message or tie the operations together.
```

I'm not sure if the same thing applies to routes. Ideally we should remove the special case, or else come up with a new error message.

internal/clients/hvn_route.go
nombiezinja: Previous block calls the LIST endpoint with the destination cidr filter, and should only return 1 single route, therefore this error is here. What do you think is incorrect about it?

internal/provider/resource_aws_network_peering.go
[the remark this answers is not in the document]
nombiezinja: For the purpose of this PR the changes to peering and TGW is necessary, otherwise submitting a value during creation of peering/tgw attachment will result in a route having been created for these resources, causing an error during later HVN route creation

internal/provider/resource_hvn_route.go
smaant: this PR didn't add `updated_at` for hvn routes, so we don't expect it to be added, removed that TODO

docs/resources/hvn_route.md
smaant: done

docs/resources/hvn_route.md
smaant: done

internal/provider/resource_hvn_route.go
smaant: discussed with Brenna on zoom and agreed that this is unnecessary here to duplicate backend logic

internal/provider/resource_hvn_route.go
smaant: same, discussed with Brenna on zoom and agreed that this is unnecessary here to duplicate backend logic

docs/resources/hvn_route.md
```diff
@@ -0,0 +1,89 @@
...
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- **destination_cidr** (String) The destination CIDR of the HVN route
+- **hvn** (String) The `self_link` of the HashiCorp Virtual Network (HVN).
+- **hvn_route_id** (String) The ID of the HVN route.
+- **target_link** (String) A unique URL identifying the target of the HVN route. Examples of the target: [`aws_network_peering`](aws_network_peering.md), [`aws_transit_gateway_attachment`](aws_transit_gateway_attachment.md)
```
smaant, line 61: @bcmdarroch I wonder if you know if these links would work that way in the TF registry

internal/provider/resource_hvn_route.go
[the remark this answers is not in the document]
smaant: done

internal/provider/resource_aws_network_peering.go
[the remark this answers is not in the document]
smaant: removed all not HVN route related changed from this PR

internal/provider/data_source_hvn_route.go
```diff
@@ -20,23 +20,27 @@ func dataSourceHVNRoute() *schema.Resource {
...
 				Type:        schema.TypeString,
 				Required:    true,
 			},
-			"destination_cidr": {
-				Description:  "The destination CIDR of the HVN route",
-				Type:         schema.TypeString,
-				Required:     true,
-				ValidateFunc: validation.IsCIDR,
+			"hvn_route_id": {
+				Description: "The HVN route ID",
```
bcmdarroch, line 29:
```suggestion
				Description: "The ID of the HVN route.",
```

internal/provider/data_source_hvn_route.go
```diff
@@ -20,23 +20,27 @@ func dataSourceHVNRoute() *schema.Resource {
...
+				Required:    true,
 			},
 			// Computed outputs
 			"self_link": {
 				Description: "A unique URL identifying the HVN route.",
 				Type:        schema.TypeString,
 				Computed:    true,
 			},
+			"destination_cidr": {
+				Description: "The destination CIDR of the HVN route",
```
bcmdarroch, line 40:
```suggestion
				Description: "The destination CIDR of the HVN route.",
```

internal/provider/link.go
```diff
@@ -102,18 +102,27 @@ func linkURL(l *sharedmodels.HashicorpCloudLocationLink) (string, error) {
 // parseLinkURL parses a link URL into a link. If the URL is malformed, an
 // error is returned.
 //
+// If `expectedType` is provided it will be matched against the resource from
+// the URL and if they don't much function returns an error. If `expectedType`
```
bcmdarroch, line 106:
```suggestion
// the URL and if they don't match the function returns an error. If `expectedType`
```

bcmdarroch: Thanks for adding this comment!

internal/provider/resource_hvn_route.go
```diff
@@ -1,14 +1,238 @@
...
+			},
+			"hvn_route_id": {
+				Description:      "The ID of the HVN route.",
+				Type:             schema.TypeString,
+				Required:         true,
+				ForceNew:         true,
+				ValidateDiagFunc: validateSlugID,
+			},
+			"destination_cidr": {
+				Description:  "The destination CIDR of the HVN route",
```
bcmdarroch, line 53:
```suggestion
				Description:  "The destination CIDR of the HVN route.",
```

docs/resources/hvn_route.md
bcmdarroch: Yes this should work.

examples/data-sources/hcp_hvn_route/variables.tf
```diff
@@ -1,9 +1,9 @@
-variable "hvn" {
+variable "hvn_link" {
   description = "The `self_link` of the HashiCorp Virtual Network (HVN)."
   type        = string
 }
 
-variable "destination_cidr" {
-  description = "The destination CIDR of the HVN route."
+variable "hvn_route_id" {
+  description = "The HVN route ID."
```
bcmdarroch, line 7:
```suggestion
  description = "The ID of the HVN route ID."
```

internal/provider/resource_hvn_route.go
smaant: currently backend will just return 503 for the second, third, etc delete requests of the same hvn route, I'll fix it so it would just return the same operation if delete is already in progress, either way this special case won't be necessary here

bcmdarroch (approved):
Looking good! I haven't gotten the chance to test this out locally but code all looks great. Just had some nitpicks on documentation.

docs/data-sources/hvn_route.md
```diff
@@ -25,8 +25,8 @@ data "hcp_hvn_route" "example" {
 
 ### Required
 
-- **destination_cidr** (String) The destination CIDR of the HVN route
-- **hvn** (String) The `self_link` of the HashiCorp Virtual Network (HVN).
+- **hvn_link** (String) The `self_link` of the HashiCorp Virtual Network (HVN).
+- **hvn_route_id** (String) The HVN route ID
```
smaant, line 29:
```suggestion
- **hvn_route_id** (String) The ID of the HVN route ID.
```

docs/resources/hvn_route.md
```diff
@@ -0,0 +1,89 @@
...
+  target_link      = hcp_aws_network_peering.example.self_link
+}
+```
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- **destination_cidr** (String) The destination CIDR of the HVN route
```
smaant, line 58:
```suggestion
- **destination_cidr** (String) The destination CIDR of the HVN route.
```

docs/data-sources/hvn_route.md
```diff
@@ -36,6 +36,7 @@ data "hcp_hvn_route" "example" {
 ### Read-Only
 
 - **created_at** (String) The time that the HVN route was created.
+- **destination_cidr** (String) The destination CIDR of the HVN route
```
smaant, line 39:
```suggestion
- **destination_cidr** (String) The destination CIDR of the HVN route.
```

smaant (approved):

Issue #133, opened 2021-06-03
Allow for more streamlined vault provider integration


rgbrizzlehizzle:
### Description

Hey, cannot tell you how excited I am for this provider!  I have a small QoL feature request that I think people will like.  

Right now, the `hcp_vault_cluster.vault_public_endpoint_url` is mistmatched with what a standard vault provider would expect.  Users need to do something like this 

```hcl
provider "vault" {
  address = join("", ["https://", hcp_vault_cluster.bkbn.vault_public_endpoint_url, ":8200"])
  token = hcp_vault_cluster_admin_token.bkbn.token
}
```

in order to point a vault provider at a HCP vault cluster.

Ideally, this should be streamlined to something like

```hcl
 provider "vault" {
  address = hcp_vault_cluster.bkbn.vault_public_full_url // Not a fan of this name but you get the point 😃 
  token = hcp_vault_cluster_admin_token.bkbn.token
}
```

### New or Affected Resource(s)

* hcp_0.6.0

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

bcmdarroch:
Thanks for submitting this issue @rgbrizzlehizzle! I'm currently investigating whether we can fix this at the API level, so will keep you posted!

rgbrizzlehizzle:
closing, resolved in hcp provider v `0.8.0`, thanks @bcmdarroch 🙏

Issue #124, opened 2021-05-18
Consul and Vault URLs aren't correct


nadenf:
### Terraform Version and Provider Version
```
0.15.3
```

### Affected Resource(s)

* hcp_consul_cluster
* hcp_vault_cluster

### Terraform Configuration Files

```hcl
output consul_url {
  description = ""
  value = hcp_consul_cluster.consul.consul_private_endpoint_url
}

output vault_url {
  description = ""
  value = hcp_vault_cluster.vault.vault_private_endpoint_url
}
```

### Expected Behavior

URL outputs should be actual URLs i.e. prefixed with https://

### Actual Behavior

URLs such as the following are generated:

```
consul.private.consul.xxxxxxxxxx.aws.hashicorp.cloud
```

Which then cause errors when you pass them into the Consul and Vault providers.

bcmdarroch:
Hi @nadenf! Thanks for adding this issue. I'll bring this request up to our cloud Consul and Vault teams and provide an update when I hear from them.

nadenf:
Closed by #145

PR #145, opened 2021-06-11
Clusters return complete url instead of DNS names

bcmdarroch:

Addresses https://github.com/hashicorp/terraform-provider-hcp/issues/124 and https://github.com/hashicorp/terraform-provider-hcp/issues/133. While it's possible in the future the backend responses will be updated to return a complete URL, for now the provider will handle the translation.

✖️ Are there any feature flags that are required to use this functionality?

```
$ make testacc TESTARGS='-run=TestAccConsulCluster'

--- PASS: TestAccConsulCluster (551.46s)

$ make testacc TESTARGS='-run=TestAccVaultCluster'

--- PASS: TestAccVaultCluster (683.37s)
```


internal/provider/resource_consul_cluster.go
```diff
@@ -450,12 +451,12 @@ func setConsulClusterResourceData(d *schema.ResourceData, cluster *consulmodels.
 	}
 
 	if publicEndpoint {
-		if err := d.Set("consul_public_endpoint_url", cluster.DNSNames.Public); err != nil {
+		if err := d.Set("consul_public_endpoint_url", fmt.Sprintf("https://%s:8200", cluster.DNSNames.Public)); err != nil {
```
riddhi89, line 454: Hmm, any reason the port has been included here? And which port is this 🤔

jfreda (commented):
This looks good for Vault, but I think the Consul port needs to be changed (and `testAccCheckFullURL` refactored accordingly).

internal/provider/resource_consul_cluster.go
bcmdarroch: This is the [port the UI sets](https://github.com/hashicorp/cloud-ui/blob/81c0cbed5a0b5b8211184f6e26891adf420c625f/engines/vault/addon-test-support/pages/components/page/detail.js#L7-L8) in the copy component on the cluster details page. Not sure where it came from originally though.

riddhi89: Hmm, I think dropping it altogether might be better. It keeps consistent with the HCP UI too?

riddhi89: This is what I get when I copy the url from the UI:
```https://consul-cluster-primary.consul.11eb5071-85f5-1eb2-992c-0242ac110003.aws.hashicorp.cloud
https://consul-cluster-primary.private.consul.11eb5071-85f5-1eb2-992c-0242ac110003.aws.hashicorp.cloud```
Does not look like there are any ports attached.

jfreda: Oh whoops, I think I reviewed at like the same time that @riddhi89 did yesterday and I missed this thread 🙂  HCP Vault uses Vault's default port 8200 for its API, but I believe that HCP Consul just uses port 443 for its API so its URL shouldn't need to include a port (but will let @riddhi89 confirm this for sure).

riddhi89: Yup. We can make Consul HTTP API requests using HTTPS/443 port:

bcmdarroch: Thanks! 🙌

riddhi89 (approved):

jfreda (approved):

PR #151, opened 2021-06-18
Update changelog for 0.8.0

bcmdarroch:

In preparation for release of 0.8.0!

```
$ make testacc

--- PASS: TestAccHvnPeering (224.95s)
--- PASS: TestAccTGWAttachment (514.70s)
--- PASS: TestAccConsulCluster (543.60s)
--- PASS: TestAccConsulSnapshot (611.61s)
--- PASS: TestAccHvnRoute (220.75s)
--- PASS: TestAccHvn (74.41s)
--- PASS: TestAccVaultCluster (653.02s)

ok  	github.com/hashicorp/terraform-provider-hcp/internal/provider	2843.669s
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@
 Please remove any functions that adjust the output from the `vault_private_endpoint_url`, `vault_public_endpoint_url`, `consul_private_endpoint_url`, and `consul_public_endpoint_url` when upgrading to this version. ⚠️
 
 IMPROVEMENTS:
-* resource/vault_cluster: `tier` is now an optional input (#144)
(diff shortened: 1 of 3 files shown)


roaks3 (commented):
I actually had one comment on the other items in the changelog:

I see that the endpoint URL changes are listed as fixes instead of breaking changes, but I believe they will break previously-working HCL if a user upgrades to this version of the provider and is already appending the protocol/port. I know that we're still pre-1.0, so breaking changes are somewhat expected and can be captured in a minor version bump, but I still think we need to be explicit in the changelog about them (ie. listing them under a `BREAKING CHANGES:` header). Potentially even worth including some guidance for transitioning existing HCL?

zackiles (approved):

PR #160, opened 2021-07-02
HCPE-1198: add CIDR guidance to HVN resource documentation

bcmdarroch:

Adds some helpful guidance around HVN CIDR blocks, like we offer in the UI.


lackeyjb (approved):

docs/resources/hvn.md
```diff
@@ -10,6 +9,14 @@ description: |-
 
 The HVN resource allows you to manage a HashiCorp Virtual Network in HCP.
 
+We recommend the following when selecting the CIDR block of an HVN:
+
+- If the CIDR block values for your HVN and VPCs overlap, then you will not be able to establish a connection. Ensure that any VPCs you plan to connect do not have overlapping values.
+
+- The default HVN CIDR block value does not overlap with the default CIDR block value for AWS VPCs (172.31.0.0/16). However, if you are planning to use this HVN in production, we recommend adding a custom value instead of using the default.
+
+- The CIDR block value must end between /16 and /25.
```
roaks3, line 18: One other thing since we're adding some more guidance here (although definitely not blocking): I believe our APIs expect CIDR blocks to use the first IP address in the block, and will error if an arbitrary IP address within the block is used (and I think [`cidrsubnet()`](https://www.terraform.io/docs/language/functions/cidrsubnet.html) can help with this?). We could potentially add something to help guide users past that restriction.

roaks3 (approved):

docs/resources/hvn.md
bcmdarroch: Oh I didn't know that.💡 I'll run this by control plane team:
`The CIDR block value should be the first IP address of the desired CIDR block. The [cidrsubnet()](https://www.terraform.io/docs/language/functions/cidrsubnet.html) helper can be used to specify the first address.

bcmdarroch: Got 👍 from control place. Went ahead and added a couple more rules that we're enforcing on the backend.

PR #163, opened 2021-07-08
Fix terraform updates for min_consul_version after revert

chapmanc:
If an update fails subsequent terraform runs consider the state correct. This is because the current cluster version is not validated against the min_consul_version.

Added  min_consul_version to read and set it to the consul_version. We can suppress the diff when the cluster version is greater than or equal to the what is set in the terraform by the user.

- ~~[ ] Are there any feature flags that are required to use this functionality?~~
- ~~[ ] Have you added an acceptance test for the functionality being added?~~
- [X] Have you run the acceptance tests on this branch?
-  Ran manual tests extensively in cases where user setup would cause update errors. Validated that it works correctly.

### DIFF
--- a/internal/provider/resource_consul_cluster.go
+++ b/internal/provider/resource_consul_cluster.go
@@ -94,7 +94,7 @@ func resourceConsulCluster() *schema.Resource {
 				ValidateDiagFunc: validateSemVer,
 				DiffSuppressFunc: func(_, old, new string, _ *schema.ResourceData) bool {
 					// Suppress diff for non specified value
-					if new == "" {
+					if new == "" || old == "" {
 						return true
 					}
 
--- a/internal/provider/resource_consul_cluster.go
+++ b/internal/provider/resource_consul_cluster.go
@@ -103,7 +103,7 @@ func resourceConsulCluster() *schema.Resource {
 					log.Printf("[DEBUG] Actual Consul Version %v", old)
 					log.Printf("[DEBUG] Current TF Version %v", new)
 					// suppres diff if the specified min_consul_version is <= to the actual consul version
-					return currentTFVersion.LessThanOrEqual(actualConsulVersion) || new == ""
+					return currentTFVersion.LessThanOrEqual(actualConsulVersion)
 				},
 			},
 			"datacenter": {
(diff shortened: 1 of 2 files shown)


lackeyjb (approved):

internal/provider/resource_consul_cluster.go
```diff
@@ -92,9 +93,17 @@ func resourceConsulCluster() *schema.Resource {
...
+					if new == "" {
+						return true
+					}
+
+					actualConsulVersion := version.Must(version.NewVersion(old))
+					currentTFVersion := version.Must(version.NewVersion(new))
+					log.Printf("[DEBUG] Actual Consul Version %v", old)
+					log.Printf("[DEBUG] Current TF Version %v", new)
+					// suppres diff if the specified min_consul_version is <= to the actual consul version
+					return currentTFVersion.LessThanOrEqual(actualConsulVersion)
```
riddhi89, line 106: Hmm, I entered a version less than my cluster's 1.9.6 version, but the diff does not seem to be suppressed 🤔 

resource "hcp_consul_cluster" "mycluster" {
  hvn_id          = hcp_hvn.example.hvn_id
  cluster_id      = "consul-cluster-tf"
  tier            = "development"
  public_endpoint = true
  min_consul_version = "v1.9.3"
}

```
  ~ update in-place

  # hcp_consul_cluster.mycluster will be updated in-place
  ~ resource "hcp_consul_cluster" "mycluster" {
        id                            = "/project/2a385cdb-189b-4a05-8bc4-6b91c62fd703/hashicorp.consul.cluster/consul-cluster-tf"
      + min_consul_version            = "v1.9.3"
        # (23 unchanged attributes hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.

Do you want to perform these actions?```

riddhi89: Nvm, I did not switch to this branch before building the provider 🤦 
All good.

riddhi89 (approved):

internal/provider/resource_consul_cluster.go
chapmanc: I think we identified together that was because the tf build wasn't refreshed.

PR #167, opened 2021-07-15
fix links in changelog

bcmdarroch:

Made all the PR links work in the Changelog! ☺️ 
(I thought this used to work automatically..? Might just be thinking of PR/release note descriptions)


roaks3 (approved):
Looks good! Yea, I believe the `CHANGELOG.md` was always behaving this way where the links were not hooked up. My understanding was that someday [this `changelog-links.sh` script](https://github.com/hashicorp/terraform-provider-hcp/blob/main/scripts/changelog-links.sh) would get hooked up to some sort of CI or something.

bcmdarroch:
OMG so glad you reminded me of that. Looks like that script hasn't been working because 1) we haven't been using the `[GH-nnnn]` syntax for PR links and 2) the provider URL was wrong.

PR #168, opened 2021-07-19
HCS-2061: Add size upgrade field for consul cluster update

ellacai:
### Changes
- remove `ForceNew` on consul cluster size field - we want to update the resource instead of destroy and recreate
- modify `resourceConsulClusterUpdate` to handle size change 
- note: the logic to validate size upgrade paths is already in consul service so it is not needed here

### Testing
- tested size upgrade using terraform with consul cluster in integration (currently behind launchdarkly flag)

### Acceptance Criteria:
- HCP TF provider should call the consul-service update cluster API on a change in the value for size , instead of re-creating the cluster. 
- The PR should be approved from a consul-service engineer and a TCE team engineer
- PR will not be merged until monitoring is added

### DIFF
--- a/internal/provider/resource_consul_cluster.go
+++ b/internal/provider/resource_consul_cluster.go
@@ -131,7 +131,6 @@ func resourceConsulCluster() *schema.Resource {
 				Description:      "The t-shirt size representation of each server VM that this Consul cluster is provisioned with. Valid option for development tier - `x_small`. Valid options for other tiers - `small`, `medium`, `large`. For more details - https://cloud.hashicorp.com/pricing/consul",
 				Type:             schema.TypeString,
 				Optional:         true,
-				ForceNew:         true,
 				Computed:         true,
 				ValidateDiagFunc: validateConsulClusterSize,
 				DiffSuppressFunc: func(_, old, new string, _ *schema.ResourceData) bool {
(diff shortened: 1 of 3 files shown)


lackeyjb (commented):
This looks good, but could we add some updates to the acceptance tests for this change?

internal/provider/resource_consul_cluster.go
```diff
@@ -583,30 +582,59 @@ func resourceConsulClusterUpdate(ctx context.Context, d *schema.ResourceData, me
...
-	// Fetch available upgrade versions
-	upgradeVersions, err := clients.ListConsulUpgradeVersions(ctx, client, cluster.Location, clusterID)
-	if err != nil {
-		return diag.Errorf("unable to list Consul upgrade versions (%s): %v", clusterID, err)
+	// Confirm update fields have been changed
+	sizeChanged := d.HasChange("size")
+	versionChanged := d.HasChange("min_consul_version")
+
+	if !sizeChanged && !versionChanged {
+		return diag.Errorf("at least one of: [min_consul_version, size] is required in order to upgrade the cluster")
```
riddhi89, line 590: when is this error returned? When a config which does not have `ForceNew` enabled and nor is supported for update, is changed?

lackeyjb (approved):
Looks good 👍

riddhi89 (approved):

PR #188, opened 2021-08-24
change bucket to bucket_name to remain consistent with Packer

azr:
Following https://github.com/hashicorp/packer/pull/11219, I'm renaming this field to stay consistent.

### DIFF
--- a/internal/provider/data_source_packer_image_iteration.go
+++ b/internal/provider/data_source_packer_image_iteration.go
@@ -23,7 +23,7 @@ func dataSourcePackerImageIteration() *schema.Resource {
 		},
 		Schema: map[string]*schema.Schema{
 			// Required inputs
-			"bucket": {
+			"bucket_name": {
 				Description:      "The slug of the HCP Packer Registry image bucket to pull from.",
 				Type:             schema.TypeString,
 				Required:         true,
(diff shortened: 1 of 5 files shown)


internal/provider/data_source_packer_image_iteration.go
```diff
@@ -23,7 +23,7 @@ func dataSourcePackerImageIteration() *schema.Resource {
 		},
 		Schema: map[string]*schema.Schema{
 			// Required inputs
-			"bucket": {
+			"bucket_name": {
```
bcmdarroch, line 26: It's totally fine to change this field directly while in private beta, but I wanted to mention this guide for any field name changes that happen after resource goes GA: https://www.terraform.io/docs/extend/best-practices/deprecations.html#provider-attribute-rename

bcmdarroch (approved):

internal/provider/data_source_packer_image_iteration.go
azr: Thanks, yes, totally agreed to be careful with surprise deprecations 🙂

azr:
Thanks for reviewing, will merge at the same time the Packer change is merged ! 🙂

PR #227, opened 2021-12-04
Add Vault Cluster scaling docs

sam11b:
Hello! This is the initial PR for vault cluster scaling TF provider doc additions. We're hoping to have this merged by early next week (Dec 7th). Thanks!


templates/guides/vault-scaling.md.tmpl
```diff
@@ -0,0 +1,19 @@
+---
+subcategory: ""
+page_title: "Resize or scale a Vault cluster - HCP Provider"
+description: |-
+    Change a current HCP Vault cluster in terms of tiers (Dev, Starter, Standard) or sizes (S, M, L).
+---
+
+# Resize or scale a cluster
```
bcmdarroch, line 8: Currently the tier consists of `tier_size` and it's not really possible to change one without the other. I wonder if we should simplify the language here to just `scale`. Right now, seeing `change size OR tier` makes me think I should be able to change them separately.
```suggestion
# Scale a cluster
```

bcmdarroch (approved):
Had a suggestion for changing the wording around tier vs. size, but otherwise looks good to me!

PR #246, opened 2022-02-01
TCE-1772: peering activation data source

bcmdarroch:

This PR adds a new data source, `hcp_peering_connection_activation`. 

The purpose of this data source is to ensure a given peering connection is **active** by the time the Terraform run returns in order to avoid race conditions with other dependent resources. 
Since peering connections generally must return early with some outputs in order to complete the peering, it may take some time before the peering connection is active. This data source polls the peering connection and returns once it verifies the state is active. 

### :ship: Release Note

```release-note
* datasource/hcp_peering_connection_activation: adds the peering activation data source [GH-246]
```
### 🧪 : Try It Out

```
provider "hcp" {}

provider "aws" {
  region = "us-west-2"
}

resource "hcp_hvn" "test" {
	hvn_id         = "test-hvn"
	cloud_provider = "aws"
	region         = "us-west-2"
}

resource "aws_vpc" "vpc" {
  cidr_block = "10.220.0.0/16"
  tags = {
     Name = "hcp-provider-peering-activation-test"
  }
}

resource "hcp_aws_network_peering" "peering" {	
  peering_id      = "test-peering"
  hvn_id          = hcp_hvn.test.hvn_id
  peer_account_id = aws_vpc.vpc.owner_id
  peer_vpc_id     = aws_vpc.vpc.id
  peer_vpc_region = "us-west-2"
}

data "hcp_peering_connection_activation" "activation" {
  peering_id  = hcp_aws_network_peering.peering.peering_id
  hvn_link    = hcp_hvn.test.self_link
}

resource "hcp_hvn_route" "route" {
  hvn_route_id      = "test-route"
  hvn_link          = hcp_hvn.test.self_link
  destination_cidr  = "172.31.0.0/16"
  target_link       = hcp_aws_network_peering.peering.self_link
}

resource "aws_vpc_peering_connection_accepter" "peering-accepter" {
  vpc_peering_connection_id = hcp_aws_network_peering.peering.provider_peering_id
  auto_accept               = true
  tags = {
    Name            = "hcp-provider-peering-activation-test"
    hvn_id          = hcp_hvn.test.hvn_id
    organization_id = hcp_hvn.test.organization_id
    project_id      = hcp_hvn.test.project_id
    peering_id      = hcp_aws_network_peering.peering.peering_id
  }
}
```

```
$ make testacc TESTARGS='-run=TestAccPeeringConnectionActivation'

--- PASS: TestAccPeeringConnectionActivation (637.12s)
```


xargs-P:
@bcmdarroch is this necessary because hcp_aws_network_peering.peering returns before the service is actually ready to be accepted on the aws_vpc_peering_connection_accepter side? 🤔

internal/provider/provider.go
```diff
@@ -46,6 +46,7 @@ func New() func() *schema.Provider {
 				"hcp_hvn":                            resourceHvn(),
 				"hcp_hvn_peering_connection":         resourceHvnPeeringConnection(),
 				"hcp_hvn_route":                      resourceHvnRoute(),
+				"hcp_peering_connection_activation":  resourcePeeringConnectionActivation(),
```
bcmdarroch, line 49: How do we feel about this name? 🤔 another idea was `hcp_peering_wait_for_active`

internal/provider/resource_hcp_peering_connection_activation.go
```diff
@@ -0,0 +1,118 @@
...
+
+	"github.com/hashicorp/terraform-provider-hcp/internal/clients"
+)
+
+var activationDefaultTimeout = time.Minute * 5
+
+// The purpose of this resource is to ensure the given peering connection is active before the Terraform run returns. Since
+// peering connections generally must return early with some outputs in order to complete the peering, it may take some time before the peering connection is active.
+// This resource polls the peering connection and returns once it verifies the state is active.
+func resourcePeeringConnectionActivation() *schema.Resource {
```
bcmdarroch, line 21: On second thought, this might make more sense as a read-only [data source](https://www.terraform.io/language/data-sources) 🤔

bcmdarroch:
@xargs-P When `hcp_aws_network_peering` returns, it _is_ ready to be accepted by `aws_vpc_peering_connection_accepter`. But there is additional setup that has to happen on the HCP side after the peering is accepted in order for it become active and ready for us, which is when race conditions arise for dependent resources like Consul clusters. This resource should alleviate those race conditions.

internal/provider/resource_hcp_peering_connection_activation.go
roaks3: Yea, that's what I was thinking too when reading through. I think a data source might fit better, and probably makes more sense since we don't appear to actually be creating a piece of infrastructure.

internal/clients/peering.go
```diff
@@ -78,7 +79,10 @@ func waitForPeeringToBe(state string) WaitFor {
 
 // WaitForPeeringToBePendingAcceptance will poll the GET peering endpoint until
 // the state is PENDING_ACCEPTANCE, ctx is canceled, or an error occurs.
-var WaitForPeeringToBePendingAcceptance = waitForPeeringToBe(PeeringStatePendingAcceptance)
+var WaitForPeeringToBePendingAcceptance = waitForPeeringToBe(PeeringStatePendingAcceptance, []string{PeeringStateCreating})
```
roaks3, line 82: Nit: An unfortunate side-effect of the second parameter is we lose some readability (as in, "what are those extra states being passed in?"). I believe the AWS provider just creates a devoted function for each case, but if we want to stick with this parameterized function, we may want to consider passing a config object ie. `{Target: PeeringStatePendingAcceptance, Pending: []string{PeeringStateCreating}}`

internal/clients/peering.go
```diff
@@ -78,7 +79,10 @@ func waitForPeeringToBe(state string) WaitFor {
...
 
 // WaitForPeeringToBeAccepted will poll the GET peering endpoint until the state is ACCEPTED, ctx in canceled, or an error occurs.
-var WaitForPeeringToBeAccepted = waitForPeeringToBe(PeeringStateAccepted)
+var WaitForPeeringToBeAccepted = waitForPeeringToBe(PeeringStateAccepted, []string{PeeringStateCreating, PeeringStatePendingAcceptance})
+
+// WaitForPeeringToBeActive will poll the GET peering endpoint until the state is ACTIVE, ctx in canceled, or an error occurs.
```
jasonpilz, line 87:
```suggestion
// WaitForPeeringToBeActive will poll the GET peering endpoint until the state is ACTIVE, ctx is canceled, or an error occurs.
```

internal/provider/resource_hcp_peering_connection_activation.go
jasonpilz: +1 Data source seems to be the better fit, yeah.

internal/clients/peering.go
bcmdarroch: Ooh I like the config idea

bcmdarroch:
After some internal discussion, we decided to solve this problem in a way that didn't necessitate a new data source (see https://github.com/hashicorp/terraform-provider-hcp/pull/248 for new solution)

PR #248, opened 2022-02-04
TCE-1770: Azure peering connection (internal only)

bcmdarroch:

This PR adds support for Azure networking in the form of the new `hcp_azure_peering_connection`. 

⚠️ **This feature is available for internal users only.**

### :ship: Release Note

```release-note
* resource/azure_peering_connection: adds Azure peering resource (internal only)  [GH-248]
```

### 🧪 Example Config
Get creds with `az login`

```
// HCP resources
provider "hcp" {}

locals {
  subscription_id = "<get from az login>" 
  tenant_id       = "<get from az login>"
  vnet_region     = "eastus"
  role_def_name  = "test-hcp-hvn-peering-access"
}

resource "hcp_hvn" "azure" {
  hvn_id         = "test-azure-hvn"
  cloud_provider = "azure"
  region         = local.vnet_region
}

resource "azurerm_resource_group" "rg" {
  name     = "resource-group-test"
  location = "East US"
}

resource "azurerm_virtual_network" "vnet" {
  name                = "test-vnet"
  location            = azurerm_resource_group.rg.location
  resource_group_name = azurerm_resource_group.rg.name

  address_space = [
    "10.0.0.0/16"
  ]
}

resource "hcp_azure_peering_connection" "peering" {
  hvn_link                 = hcp_hvn.azure.self_link
  peering_id               = "test-peering"
  peer_vnet_name           = azurerm_virtual_network.vnet.name
  peer_subscription_id     = local.subscription_id
  peer_tenant_id           = local.tenant_id
  peer_resource_group_name = azurerm_resource_group.rg.name
  peer_vnet_region         = local.vnet_region
}

data "hcp_azure_peering_connection" "peering" {
  hvn_link                 = hcp_hvn.azure.self_link
  peering_id               = hcp_azure_peering_connection.peering.peering_id
  wait_for_active_state  = true
}

resource "hcp_hvn_route" "route" {
  hvn_route_id = "test-route"
  hvn_link = hcp_hvn.azure.self_link
  destination_cidr = "172.31.0.0/16"
  target_link = data.hcp_azure_peering_connection.peering.self_link
}

// Azure resources
provider "azurerm" {
  features {}
}

resource "azuread_service_principal" "principal" {
  application_id = hcp_azure_peering_connection.peering.application_id
}

resource "azurerm_role_definition" "definition" {
  name  = local.role_def_name
  scope = azurerm_virtual_network.vnet.id
  assignable_scopes = [
    azurerm_virtual_network.vnet.id
  ]
  permissions {
    actions = [
      "Microsoft.Network/virtualNetworks/peer/action",
      "Microsoft.Network/virtualNetworks/virtualNetworkPeerings/read",
      "Microsoft.Network/virtualNetworks/virtualNetworkPeerings/write"
    ]
  }
}

resource "azurerm_role_assignment" "role_assignment" {
  principal_id       = azuread_service_principal.principal.id
  role_definition_id = azurerm_role_definition.definition.role_definition_resource_id
  scope              = azurerm_virtual_network.vnet.id
}

```

```
$ make testacc TESTARGS='-run=TestAccAzurePeeringConnection'

--- PASS: TestAccAzurePeeringConnection (991.21s)

$ make testacc TESTARGS='-run=TestAccAwsPeering'

--- PASS: TestAccAwsPeering (764.13s)
```


docs/resources/azure_peering_connection.md
```diff
@@ -0,0 +1,136 @@
...
+}
+
+resource "hcp_azure_peering_connection" "peer" {
+  hvn_link                 = hcp_hvn.hvn.self_link
+  peering_id               = "dev"
+  peer_vnet_name           = azurerm_virtual_network.vnet.name
+  peer_subscription_id     = "50dc4fb8-db16-4c07-8c35-f31467122911"
+  peer_tenant_id           = "09cdb642-fb81-4425-ba72-b882790ecb06"
+  peer_resource_group_name = azurerm_resource_group.rg.name
+  peer_vnet_region         = hcp_hvn.hvn.region
```
pcewing, line 31: Should this be the region of the vnet being peered to the HVN? Or the HVN itself? If it's the former, it might be be better for this to be:
```
peer_vnet_region = azure_rm_virtual_network.region
```
Or whatever the correct output variable is, otherwise this might be misleading if the vnet is in a different region from the HVN.

docs/resources/azure_peering_connection.md
```diff
@@ -0,0 +1,136 @@
...
+
+resource "hcp_azure_peering_connection" "peer" {
+  hvn_link                 = hcp_hvn.hvn.self_link
+  peering_id               = "dev"
+  peer_vnet_name           = azurerm_virtual_network.vnet.name
+  peer_subscription_id     = "50dc4fb8-db16-4c07-8c35-f31467122911"
+  peer_tenant_id           = "09cdb642-fb81-4425-ba72-b882790ecb06"
+  peer_resource_group_name = azurerm_resource_group.rg.name
+  peer_vnet_region         = azure_rm_virtual_network.region
+}
```
manish-hashicorp, line 32: Does this resource succeed when peering reaches pending state?

docs/resources/azure_peering_connection.md
```diff
@@ -0,0 +1,136 @@
...
+
+resource "hcp_peering_connection_activation" "activation" {
```
manish-hashicorp, line 34: Does this succeed when customer accepts the peering connection? If these two are accurate could you mention this in the docs?

examples/resources/hcp_azure_peering_connection/resource.tf
```diff
@@ -0,0 +1,73 @@
...
+  peer_resource_group_name = azurerm_resource_group.rg.name
+  peer_vnet_region         = azure_rm_virtual_network.region
+}
+
+resource "hcp_peering_connection_activation" "activation" {
+  peering_id = hcp_azure_peering_connection.peer.peering_id
+  hvn_link   = hcp_hvn.hvn.self_link
+}
+
+resource "hcp_hvn_route" "route" {
```
manish-hashicorp, line 23: Does this work with Azure? I am going to test now but curious if you noticed any problems with it.

internal/clients/peering.go
```diff
@@ -78,7 +85,19 @@ func waitForPeeringToBe(state string) WaitFor {
 
 // WaitForPeeringToBePendingAcceptance will poll the GET peering endpoint until
 // the state is PENDING_ACCEPTANCE, ctx is canceled, or an error occurs.
-var WaitForPeeringToBePendingAcceptance = waitForPeeringToBe(PeeringStatePendingAcceptance)
+var WaitForPeeringToBePendingAcceptance = waitForPeeringToBe(peeringState{
```
manish-hashicorp, line 88: This is nice. I like the way these functions are composed with `waitForPeeringToBe`.

internal/provider/data_source_aws_network_peering.go
```diff
@@ -31,6 +32,13 @@ func dataSourceAwsNetworkPeering() *schema.Resource {
 				Required:         true,
 				ValidateDiagFunc: validateSlugID,
 			},
+			// Optional inputs
+			"wait_for_active_state": {
```
manish-hashicorp, line 36: This is outside vnet peering scope. Can this be in another PR?

internal/provider/data_source_azure_peering_connection.go
```diff
@@ -0,0 +1,147 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+			"hvn_link": {
+				Description: "The `self_link` of the HashiCorp Virtual Network (HVN).",
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+			// Optional inputs
+			"wait_for_active_state": {
```
manish-hashicorp, line 34: If `hcp_peering_connection_activation` is used to wait for peering connection to reach ACTIVE then is this field required?

internal/provider/data_source_azure_peering_connection.go
```diff
@@ -0,0 +1,147 @@
...
+				Type:        schema.TypeString,
+				Computed:    true,
+			},
+			"project_id": {
+				Description: "The ID of the HCP project where the peering connection is located. Always matches the HVN's project.",
+				Type:        schema.TypeString,
+				Computed:    true,
+			},
+			"application_id": {
+				Description: "The application ID of the HCP VNet backing the HVN.",
```
manish-hashicorp, line 52: I think the following text may make more sense: "The application id is the ID of Azure application whose credentials are used to peer HCP HVN VNet with customer VNet."

internal/provider/data_source_azure_peering_connection.go
```diff
@@ -0,0 +1,147 @@
...
+				Description: "The tenant ID of the peer VNet in Azure.",
+				Type:        schema.TypeString,
+				Computed:    true,
+			},
+			"peer_resource_group_name": {
+				Description: "The resource group name of the peer VNet in Azure.",
+				Type:        schema.TypeString,
+				Computed:    true,
+			},
+			"provider_peering_id": {
```
manish-hashicorp, line 81: Since this is azure specific resource can this be renamed to `azure_peering_id`?

internal/provider/resource_azure_peering_connection.go
```diff
@@ -0,0 +1,400 @@
...
+				Type:        schema.TypeString,
+				Computed:    true,
+			},
+			"project_id": {
+				Description: "The ID of the HCP project where the peering connection is located. Always matches the HVN's project.",
+				Type:        schema.TypeString,
+				Computed:    true,
+			},
+			"application_id": {
+				Description: "The application ID of the HCP VNet backing the HVN.",
```
manish-hashicorp, line 92: Likewise here too: "The ID of Azure application whose credentials are used to peer VNets."

internal/provider/resource_azure_peering_connection.go
```diff
@@ -0,0 +1,400 @@
...
+				Type:        schema.TypeString,
+				Computed:    true,
+			},
+			"provider_peering_id": {
```
manish-hashicorp, line 96: Likewise here too - could this be renamed to `azure_peering_id`?

internal/provider/resource_azure_peering_connection.go
```diff
@@ -0,0 +1,400 @@
...
+			return nil
+		}
+
+		return diag.Errorf("unable to retrieve peering connection (%s): %v", peeringID, err)
+	}
+
+	// The peering connection failed to provision properly so we want to let the user know and
+	// remove it from state
+	if peering.State == networkmodels.HashicorpCloudNetwork20200907PeeringStateFAILED {
+		log.Printf("[WARN] peering connection (%s) failed to provision, removing from state", peering.ID)
```
manish-hashicorp, line 260: It it normal in terraform to remove resource from state if it is in FAILED state? I would expect it to error out and expect user to take action manually rather than surprising them with deletion.

manish-hashicorp (approved):
Looks good! I haven't looked in details at test. I've some suggestions about resource and property names. Please do let me know your thoughts on those.

docs/resources/azure_peering_connection.md
bcmdarroch: Yes, it has to return with the `application_id` for Azure to accept the connection.

docs/resources/azure_peering_connection.md
bcmdarroch: Yes! Good reminder - I still need to update the docs with the new direction we took.

examples/resources/hcp_azure_peering_connection/resource.tf
bcmdarroch: So far it has as long as the peering connection is already active.

internal/provider/data_source_aws_network_peering.go
bcmdarroch: Good point

internal/provider/data_source_azure_peering_connection.go
bcmdarroch: Actually the docs are outdated. The activation datasource was the original solution, but we decided we'd rather follow an existing TGW pattern and use this `wait_for_active_state` flag on the peering datasource.

internal/provider/resource_azure_peering_connection.go
bcmdarroch: Good question. So far we've taken the approach across our resources to not track the resource's `state` as a property - so the user doesn't have much recourse to fix anything in a bad state except to recreate. We delete with no error here so that we don't disrupt the rest of the TF apply. The user _will_ see that the peering is not there on the next TF run, because TF will show in the plan it still needs to be created.

PR #278, opened 2022-03-21
Remove beta notes from hcp packer data sources

laurapacilio:
HCP Packer recently completed a GA launch! This PR removes the notes about HCP Packer data sources being in beta because they're not anymore! 🥳


sylviamoss (approved):
Nice!

lackeyjb (approved):

PR #293, opened 2022-04-20
[DO NOT MERGE] changes for migration to new releases-api

modrake:
This reflects the changes that acccompany the releases-api anticipated to be launched in early May. Included is:

*    Version bump of the terraform provider release gh action
*    Variables added for interfacing with releases-api
*    hc-releases syntax modified for releases-api use
*    metadata file added for ingesting with releases-api


bcmdarroch (approved):
Thanks! 🎉

Issue #222, opened 2021-11-08
Failed HVN resource is not deleted on terraform destroy


crhino:
### Terraform Version and Provider Version

```
Terraform version: 1.0.0
HCP provider version: 0.7.0
```

### Affected Resource(s)

* hcp_hvn
* Also probably others...

### Terraform Configuration Files

```hcl
terraform {
  required_providers {
    hcp = {
      source  = "hashicorp/hcp"
      version = "~> 0.7.0"
    }
  }
}

resource "hcp_hvn" "default" {
  hvn_id         = var.hvn_id
  cloud_provider = var.cloud_provider
  region         = var.region
  cidr_block     = var.hvn_cidr_block
}

data "hcp_consul_versions" "versions" {}

resource "hcp_consul_cluster" "default" {
  cluster_id         = var.cluster_id
  min_consul_version = var.consul_version == "recommended" ? data.hcp_consul_versions.versions.recommended : var.consul_version
  hvn_id             = hcp_hvn.default.hvn_id
  tier               = var.tier
  public_endpoint    = var.public_endpoint
}
```

### Steps to Reproduce

1. `terraform apply`
2. HVN creation fails, example error: 
```
Error: unable to create HVN (cloud-e2e-bbd97b5-70959): create HVN operation (94869ad2-7594-4f49-8ee6-181a6ae4d0a5) failed [code=13, message=terraform apply failed: error applying Terraform: failed to apply Terraform config: error reading Route for Route Table (rtb-00a99602827a4ba82) with destination (0.0.0.0/0): couldn't find resource]
│ 
│   with module.hcp_e2e_consul_service.hcp_hvn.default,
│   on ../../../terraform/modules/hcp_public/hcp/hcp.tf line 10, in resource "hcp_hvn" "default":
│   10: resource "hcp_hvn" "default" {
│ 
╵

Exited with code exit status 1
```
3. `terraform destroy`

### Expected Behavior

During the `terraform destroy`, I expected to see my `FAILED` HVN resource deleted.

### Actual Behavior

```
2021-11-08T21:54:33Z module.hcp_e2e_consul_service.hcp_hvn.default: Refreshing state... [id=/project/xxxxx/hashicorp.network.hvn/xxxxxx]
2021-11-08T21:54:35Z 
2021-11-08T21:54:35Z Note: Objects have changed outside of Terraform
2021-11-08T21:54:35Z 
[...]

PR #326, opened 2022-06-13
Store Consul cluster/snapshot state to fix failed cluster behavior

jjtimmons:

Background: we have an issue where resources are leaked in our E2E tests. This is because we purge all failed clusters from state with `d.SetId("")`. 

Changes:
- add `cluster_state` and `snapshot_state` to Consul clusters and snapshots, respectively
- remove `DELETED` Consul clusters from state (not `FAILED`)
- set Consul client config resource properties _after_ setting the rest of the resource properties. The `GetConsulClusterByID` call can succeed before the `GetConsulClientConfigFiles` API call fails -- we should still set the resource properties we have from the `GetConsulClusterByID` call

UX change:
- `FAILED` clusters/snapshots are stored in state so they're deleted by terraform on the next `terraform apply` or `terraform delete`. This differs from the behavior right now where users have to find and delete the resources in UI manually because Terraform forgets about them

[Example of a failed E2E test here](https://app.circleci.com/pipelines/github/hashicorp/cloud-e2e/112124/workflows/24718995-e963-45e2-890b-ff4694ed3d6b/jobs/112230/steps). We mark the cluster as deleted and remove it from state:

```
2022-06-11T16:11:36Z   # module.hcp_e2e_consul_service.hcp_consul_cluster.default has been deleted
2022-06-11T16:11:36Z   - resource "hcp_consul_cluster" "default" {
2022-06-11T16:11:36Z       - cluster_id         = "cloud-e2e-a541b8d-112230" -> null
2022-06-11T16:11:36Z       - connect_enabled    = true -> null
```

That's the opposite of what we want to do: delete the failed cluster on the next run. And our E2E org becomes filled with Consul clusters and HVNs: https://admin.hcp.to/organizations/orgs/11eb1545-60c8-4b02-9121-0242ac110008/resources

From [docs](https://learn.hashicorp.com/tutorials/terraform/provider-setup):

> When you create something in Terraform but delete it manually, Terraform should gracefully handle it. If the API returns an error when the resource doesn't exist, the read function should check to see if the resource is available first. If the resource isn't available, the function should set the ID to an empty string so Terraform "destroys" the resource in state. The following code snippet is an example of how this can be implemented; you do not need to add this to your configuration for this tutorial.

As far as I can tell, the behavior in this PR will match the [desired behavior of the original PR](https://github.com/hashicorp/terraform-provider-hcp/pull/59):

> Set the ID of resources that need to poll for an operation before polling the operation. This prevents the case where a resource is created in the DB but the async resource operation fails and Terraform has no state entry so it thinks the resource does not exist (even though it does and is in a failed state), so the user can't run terraform destroy to clean up the failed resource.

closes: https://github.com/hashicorp/terraform-provider-hcp/issues/335

### :ship: Release Note

```release-note
* resource/consul: Store cluster+snapshot state [GH-326]
```

```
$ → make testacc TESTARGS='-run=TestAccConsulCluster'
TF_ACC=1 go test ./internal/... -v -run=TestAccConsulCluster -timeout 210m
?       github.com/hashicorp/terraform-provider-hcp/internal/clients    [no test files]
PASS
PASS
=== RUN   TestAccConsulCluster
    resource_consul_cluster_test.go:62: Step 6/6 error: Error running apply: exit status 1
        
        Error: unable to create Consul cluster (test-consul-cluster): [POST /consul/2021-02-04/organizations/{cluster.location.organization_id}/projects/{cluster.location.project_id}/clusters][403] Create default  &{Code:7 Details:[] Error:billing authorization check failed Message:billing authorization check failed}
...

→ make testacc TESTARGS='-run=TestAccConsulSnapshot'
TF_ACC=1 go test ./internal/... -v -run=TestAccConsulSnapshot -timeout 210m
?       github.com/hashicorp/terraform-provider-hcp/internal/clients    [no test files]
PASS
ok      github.com/hashicorp/terraform-provider-hcp/internal/consul     0.140s [no tests to run]
PASS
ok      github.com/hashicorp/terraform-provider-hcp/internal/input      0.131s [no tests to run]
=== RUN   TestAccConsulSnapshot
--- PASS: TestAccConsulSnapshot (909.93s)
PASS
ok      github.com/hashicorp/terraform-provider-hcp/internal/provider   910.251s
```

Everything passes except the update where I hit a payments/no-credit-card err

### E2E Testing

I created a cluster, terminated the dataplane deployment, and confirmed that `cluster_state` was stored after `tf refresh`:

```
      "type": "hcp_consul_cluster",
      "instances": [{
          "status": "tainted",
...
            "cluster_id": "consul-quickstart-1655065377418",
            "cluster_state": "FAILED",
```

On `tf apply` it was replaced:

```
-/+ resource "hcp_consul_cluster" "main" {
      ~ auto_hvn_to_hvn_peering       = false -> (known after apply)
      ~ cloud_provider                = "aws" -> (known after apply)
      ~ cluster_state                 = "FAILED" -> (known after apply)
```


jjtimmons:
this handling of state behavior applies to many other resources as well. Example of another issue but for HVNs:
https://github.com/hashicorp/terraform-provider-hcp/issues/222

In all the places that we see a failed state, we remove the resource from state where we should store the state in TF state so it's deleted on the next `terraform apply/delete`: https://github.com/hashicorp/terraform-provider-hcp/search?q=failed+to+provision

internal/provider/resource_consul_cluster.go
```diff
@@ -382,12 +387,20 @@ func resourceConsulClusterCreate(ctx context.Context, d *schema.ResourceData, me
 		return diag.Errorf("unable to retrieve Consul cluster (%s): %v", payload.Cluster.ID, err)
 	}
 
+	if err := setConsulClusterResourceData(d, cluster); err != nil {
```
mkeeler, line 390: When the create fails, where would the tf have failed previously? Was it in attempting to get the client config files?

I had assumed the WaitForOperation call back on line 379 still prior to this call.

internal/provider/resource_consul_cluster_test.go
```diff
@@ -23,7 +23,7 @@ var updatedConsulCluster = `
 resource "hcp_consul_cluster" "test" {
 	cluster_id = "test-consul-cluster"
 	hvn_id     = hcp_hvn.test.hvn_id
-	tier       = "standard"
+	tier       = "development"
 	size	   = "medium"
```
mkeeler, line 27: With a dev tier shouldn't size be xsmall?

jjtimmons: Makes sense, this was a symptom of me being impatient with the ~35 minute acc testing. I'll fix and run those again

internal/provider/resource_consul_cluster.go
jjtimmons: you're right and yes, this was failing up above in the Wait. Where I saw this was an issue was in `resourceConsulClusterRead` (the `GetConsulClusterByID` call succeeds while `GetConsulClientConfigFiles` fails). I then applied the same pattern everywhere we make these two API calls (storing results of first)

docs/resources/consul_snapshot.md
```diff
@@ -42,6 +42,7 @@ resource "hcp_consul_snapshot" "example" {
 - `restored_at` (String) Timestamp of when the snapshot was restored. If the snapshot has not been restored, this field will be blank.
 - `size` (Number) The size of the snapshot in bytes.
 - `snapshot_id` (String) The ID of the Consul snapshot
+- `snapshot_state` (String) The state of an HCP Consul snapshot.
```
mkeeler, line 45: nit: `snapshot` in `snapshot_state` is redundant. Then again, there is `snapshot_id` so its consistent there.

If I were writing this from scratch I would name things as followed:

`id` -> `cluster_id`
`snapshot_id` -> `id`
`snapshot_state` -> `state`

For now though, I think I would prefer naming this `state` instead but feel free to ignore this comment and continue with the current name if you have strong feelings otherwise.

docs/resources/consul_cluster.md
```diff
@@ -51,6 +51,7 @@ resource "hcp_consul_cluster" "example" {
 ### Read-Only
 
 - `cloud_provider` (String) The provider where the HCP Consul cluster is located.
+- `cluster_state` (String) The state of the HCP Consul cluster.
```
mkeeler, line 54: nit: similar to my comment about `snapshot_state` it feels like `cluster` in `cluster_state` is redundant. It might be preferable to call it just `state`. Then again there are a lot more fields here and its hard to see whether one name or the other is clearly better.

mkeeler (approved):

internal/provider/resource_consul_cluster.go
jjtimmons: Yes, we'll still clean it up on the next `terraform delete` or `terraform apply`. A PR from last year addressed this by storing the ID of the resource before polling: https://github.com/hashicorp/terraform-provider-hcp/pull/59

Current behavior:
- `tf create` fail during poll: bail, return diag.Error, we store nothing but ID
- `tf plan`, read: GET with ID, see it's FAILED, purge from state
- `tf apply`: create a _new_ cluster. Don't delete prior cluster. Run into a duplicate cluster/hvn ID issue

Behavior after this change:
- `tf create` fail during poll: bail, return diag.Error, we store nothing but ID
- `tf plan`, read: GET with ID, set state
- `tf apply`, old cluster is tainted, delete, create a _new_ cluster

docs/resources/consul_snapshot.md
jjtimmons: using just `state` matches our model: https://github.com/hashicorp/cloud-consul-service/blob/de4939ba1d7f47c9f002ec74dc34228fe992fa7e/proto-public/hashicorp/cloud/consul_20210204/consul.proto#L157

bcmdarroch (commented):
Thanks for this fix! I think this is a solid approach, but I wanted to discuss an alternative that doesn't involve adding a `state` field.

internal/provider/resource_consul_cluster.go
```diff
@@ -150,76 +150,81 @@ func resourceConsulCluster() *schema.Resource {
...
-			"organization_id": {
-				Description: "The ID of the organization this HCP Consul cluster is located in.",
+			"cloud_provider": {
+				Description: "The provider where the HCP Consul cluster is located.",
 				Type:        schema.TypeString,
 				Computed:    true,
 			},
-			"project_id": {
-				Description: "The ID of the project this HCP Consul cluster is located in.",
+			"cluster_state": {
```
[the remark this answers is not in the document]
jjtimmons: that's the main change. But there are two other behavior changes. Highlights from description:

- add `state`
- remove DELETED Consul clusters from state (not FAILED)
- set Consul client config resource properties after setting the rest of the cluster  resource properties (ie set state after each API call, rather than at the end of calling all of them)

internal/provider/resource_consul_cluster_test.go
```diff
@@ -186,7 +189,7 @@ func TestAccConsulCluster(t *testing.T) {
 				Config: testConfig(setTestAccConsulClusterConfig(updatedConsulCluster)),
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckConsulClusterExists(resourceName),
-					resource.TestCheckResourceAttr(resourceName, "size", "MEDIUM"),
+					resource.TestCheckResourceAttr(resourceName, "size", "small"),
```
bcmdarroch, line 192: This causes an error in the test: `error="Check failed: Check 2/2 error: hcp_consul_cluster.test: Attribute 'size' expected \"small\", got \"SMALL\""`
```suggestion
					resource.TestCheckResourceAttr(resourceName, "size", "SMALL"),
```

internal/provider/resource_consul_cluster.go
jjtimmons: I realized and messed up and replied out of thread: https://github.com/hashicorp/terraform-provider-hcp/pull/326#issuecomment-1155493682

PR #339, opened 2022-06-28
hcp_packer_image channel attribute support

lbajolet-hashicorp:

Add a new channel attribute to the `hcp_packer_image` data source, so a user can fetch the latest image directly, without having to specify a companion `hcp_packer_iteration` data source to fetch the build ID from.

### :ship: Release Note

```release-note
* datasource/hcp_packer_image: allow `channel` attribute to get an image [GH-339]
```

```
$ make testacc TESTARGS='-run=TestAcc_dataSourcePackerImage_chan'
./scripts/gofmtcheck.sh: line 5: gofmt: command not found
TF_ACC=1 go test ./internal/... -v -run=TestAcc_dataSourcePackerImage_chan -timeout 210m
?   	github.com/hashicorp/terraform-provider-hcp/internal/clients	[no test files]
PASS
ok  	github.com/hashicorp/terraform-provider-hcp/internal/consul	(cached) [no tests to run]
PASS
ok  	github.com/hashicorp/terraform-provider-hcp/internal/input	(cached) [no tests to run]
=== RUN   TestAcc_dataSourcePackerImage_channelAndIterationIDReject
--- PASS: TestAcc_dataSourcePackerImage_channelAndIterationIDReject (0.08s)
=== RUN   TestAcc_dataSourcePackerImage_channelAccept
--- PASS: TestAcc_dataSourcePackerImage_channelAccept (3.94s)
PASS
ok  	github.com/hashicorp/terraform-provider-hcp/internal/provider	4.026s
```

### DIFF
--- a/docs/data-sources/packer_image.md
+++ b/docs/data-sources/packer_image.md
@@ -2,12 +2,12 @@
 page_title: "Data Source hcp_packer_image - terraform-provider-hcp"
 subcategory: ""
 description: |-
-  The Packer Image data source iteration gets the most recent iteration (or build) of an image, given an iteration id.
+  The Packer Image data source iteration gets the most recent iteration (or build) of an image, given an iteration id or a channel.
 ---
 
 # hcp_packer_image (Data Source)
 
-The Packer Image data source iteration gets the most recent iteration (or build) of an image, given an iteration id.
+The Packer Image data source iteration gets the most recent iteration (or build) of an image, given an iteration id or a channel.
 
 ## Example Usage
 
@@ -38,11 +38,12 @@ output "packer-registry-ubuntu" {
 
 - `bucket_name` (String) The slug of the HCP Packer Registry image bucket to pull from.
 - `cloud_provider` (String) Name of the cloud provider this image is stored-in.
-- `iteration_id` (String) HCP ID of this image.
 - `region` (String) Region this image is stored in, if any.
 
 ### Optional
 
+- `channel` (String) The channel that points to the version of the image being retrieved. Either this or `iteration_id` must be specified. Note: will incur a billable request
+- `iteration_id` (String) The iteration from which to get the image. Either this or `channel` must be specified.
 - `timeouts` (Block, Optional) (see [below for nested schema](#nestedblock--timeouts))
 
 ### Read-Only
--- a/docs/data-sources/packer_image.md
+++ b/docs/data-sources/packer_image.md
@@ -11,6 +11,27 @@ The Packer Image data source iteration gets the most recent iteration (or build)
 
 ## Example Usage
 
+### Single image sourcing
+
+```terraform
+data "hcp_packer_image" "baz" {
+  bucket_name    = "hardened-ubuntu-16-04"
+  cloud_provider = "aws"
+  channel        = "production"
+  region         = "us-east-1"
+}
+
+output "packer-registry-ubuntu-east-1" {
+  value = data.hcp_packer_image.baz.cloud_image_id
+}
+```
+
+~> **Note:** The `channel` attribute in this data source may incur a billable request to HCP Packer. This attribute is intended for convenience when using a single image. When sourcing multiple images from a single iteration, the `hcp_packer_iteration` data source is the alternative for querying a channel just once.
+
+~> **Note:** This data source only returns the first found image's metadata filtered by the given schema values, from the returned list of images associated with the specified iteration. Therefore, if multiple images exist in the same region, it will only pick one of them. If that's the case, you may consider separating your builds into different buckets.
+
+### Multiple image sourcing from a single iteration
+
 ```terraform
 data "hcp_packer_iteration" "hardened-source" {
   bucket_name = "hardened-ubuntu-16-04"
@@ -24,8 +45,19 @@ data "hcp_packer_image" "foo" {
   region         = "us-east-1"
 }
 
-output "packer-registry-ubuntu" {
+data "hcp_packer_image" "bar" {
+  bucket_name    = "hardened-ubuntu-16-04"
+  cloud_provider = "aws"
+  iteration_id   = data.hcp_packer_iteration.hardened-source.ulid
+  region         = "us-west-1"
+}
+
+output "packer-registry-ubuntu-east-1" {
   value = data.hcp_packer_image.foo.cloud_image_id
+}
+
+output "packer-registry-ubuntu-west-1" {
+  value = data.hcp_packer_image.bar.cloud_image_id
 }
 ```
 
--- a/internal/provider/data_source_packer_image.go
+++ b/internal/provider/data_source_packer_image.go
@@ -5,6 +5,7 @@ import (
 	"log"
 	"time"
 
+	packermodels "github.com/hashicorp/hcp-sdk-go/clients/cloud-packer-service/stable/2021-04-30/models"
 	sharedmodels "github.com/hashicorp/hcp-sdk-go/clients/cloud-shared/v1/models"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
@@ -15,7 +16,7 @@ var defaultPackerTimeout = time.Minute
 
 func dataSourcePackerImage() *schema.Resource {
 	return &schema.Resource{
-		Description: "The Packer Image data source iteration gets the most recent iteration (or build) of an image, given an iteration id.",
+		Description: "The Packer Image data source iteration gets the most recent iteration (or build) of an image, given an iteration id or a channel.",
 		ReadContext: dataSourcePackerImageRead,
 		Timeouts: &schema.ResourceTimeout{
 			Default: &defaultPackerTimeout,
@@ -33,15 +34,23 @@ func dataSourcePackerImage() *schema.Resource {
 				Type:        schema.TypeString,
 				Required:    true,
 			},
-			"iteration_id": {
-				Description: "HCP ID of this image.",
-				Type:        schema.TypeString,
-				Required:    true,
-			},
 			"region": {
 				Description: "Region this image is stored in, if any.",
 				Type:        schema.TypeString,
 				Required:    true,
+			},
+			// Optional inputs
+			"iteration_id": {
+				Description:  "The iteration from which to get the image. Either this or `channel` must be specified.",
+				Type:         schema.TypeString,
+				Optional:     true,
+				Computed:     true,
+				ExactlyOneOf: []string{"channel"},
+			},
+			"channel": {
+				Description: "The channel that points to the version of the image being retrieved. Either this or `iteration_id` must be specified. Note: will incur a billable request",
+				Type:        schema.TypeString,
+				Optional:    true,
 			},
 			// computed outputs
 			"organization_id": {
@@ -95,9 +104,10 @@ func dataSourcePackerImage() *schema.Resource {
 
 func dataSourcePackerImageRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	bucketName := d.Get("bucket_name").(string)
-	iterationID := d.Get("iteration_id").(string)
 	cloudProvider := d.Get("cloud_provider").(string)
 	region := d.Get("region").(string)
+	channelSlug := d.Get("channel").(string)
+	iterationID := d.Get("iteration_id").(string)
 	client := meta.(*clients.Client)
 
 	loc := &sharedmodels.HashicorpCloudLocationLocation{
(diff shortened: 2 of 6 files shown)


internal/provider/data_source_packer_image_iteration_test.go
```diff
@@ -251,6 +251,35 @@ func getIterationIDFromFingerPrint(t *testing.T, bucketSlug string, fingerprint
 	return ok.Payload.Iteration.ID, nil
 }
 
+func getBuildIDFromIteration(t *testing.T, bucketSlug, iterID, cloudProvider string) (string, error) {
```
JenGoldstrich, line 254: Where is this function called from?

lbajolet-hashicorp: It's called in the channelAccept test: https://github.com/hashicorp/terraform-provider-hcp/pull/339/files#diff-812fcf1f8adbdbf65c25344beec1814ebf9448de4982f3e52479c3700880fdd5R202

internal/provider/data_source_packer_image_test.go
```diff
@@ -134,3 +155,62 @@ func TestAcc_dataSourcePackerImage_revokedIteration(t *testing.T) {
...
+
+func TestAcc_dataSourcePackerImage_channelAndIterationIDReject(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		ProviderFactories: providerFactories,
+		Steps: []resource.TestStep{
+			// basically just testing that we don't pass validation here
+			{
+				PlanOnly:    true,
+				Config:      testConfig(testAccPackerImageBothChanAndIter),
+				ExpectError: regexp.MustCompile("Error: Invalid combination of arguments"),
```
sylviamoss, line 167: I can get this error when testing this locally

internal/provider/data_source_packer_image.go
```diff
@@ -111,9 +122,49 @@ func dataSourcePackerImageRead(ctx context.Context, d *schema.ResourceData, meta
...
+			client,
+			loc,
+			bucketName,
+			chanSlug)
+		if err != nil {
+			return diag.FromErr(err)
+		}
+	}
+
+	var diags diag.Diagnostics
```
sylviamoss, line 154: This variable is never being set

sylviamoss (commented):
I'm a little concerned about the `optional` fields in the docs that are not really optional and also the fact that `terraform validate` can validate those properly. 

In regards to `hcp_packer_iteration` I think we should keep it valid for now while we observe people migrating to use just `hcp_packer_image`. The iteration data source returns information about the iteration that the image doesn't, I can't tell right now how useful those information are for the user, so I definitely not deprecate right now 🤔

docs/data-sources/packer_image.md
```diff
@@ -38,11 +38,12 @@ output "packer-registry-ubuntu" {
 
 - `bucket_name` (String) The slug of the HCP Packer Registry image bucket to pull from.
 - `cloud_provider` (String) Name of the cloud provider this image is stored-in.
-- `iteration_id` (String) HCP ID of this image.
 - `region` (String) Region this image is stored in, if any.
 
 ### Optional
 
+- `channel` (String) Channel that promotes the latest iteration of the image. Either this or `iteration_id` must be specified.
+- `iteration_id` (String) HCP ID of this image. Either this or `channel' must be specified.
```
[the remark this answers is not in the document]
lbajolet-hashicorp: Not certain this is possible, the CI enforces the generated docs being the same as the ones committed :/

internal/provider/data_source_packer_image.go
lbajolet-hashicorp: True, that's an artifact from when both were allowed so long as there wasn't a conflict.
I'll remove this

internal/provider/data_source_packer_image.go
```diff
@@ -33,16 +35,23 @@ func dataSourcePackerImage() *schema.Resource {
...
 			"region": {
 				Description: "Region this image is stored in, if any.",
 				Type:        schema.TypeString,
 				Required:    true,
 			},
+			// Optional inputs
+			"iteration_id": {
+				Description:  "HCP ID of this image. Either this or `channel' must be specified.",
+				Type:         schema.TypeString,
+				Optional:     true,
```
sylviamoss, line 47: What do you think about making this value `Compute` and `Optional`and setting this as an output of this data source as well? This way, users are able to know what's the iteration being queried

internal/provider/data_source_packer_image_test.go
```diff
@@ -134,3 +155,62 @@ func TestAcc_dataSourcePackerImage_revokedIteration(t *testing.T) {
...
+	resource.Test(t, resource.TestCase{
+		PreCheck:          func() { testAccPreCheck(t, map[string]bool{"aws": false, "azure": false}) },
+		ProviderFactories: providerFactories,
+		Steps: []resource.TestStep{
+			{
+				PlanOnly: true,
+				PreConfig: func() {
+					// CheckDestroy doesn't get called when the test fails and doesn't
+					// produce any tf state. In this case we destroy any existing resource
+					// before creating them.
```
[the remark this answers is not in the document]
lbajolet-hashicorp: You're right, I did paste this from the failing test without too much thought. The other test uses `CheckDestroy` already, so that's not a big problem I imagine.
Though semantically, is this the right way to do this? My understanding was that CheckDestroy was only to test that resources are indeed removed, not to do the action of removing.
I imagine that would be done as an extra step for the teardown?

internal/provider/data_source_packer_image.go
sylviamoss: This actually already set but in the output from the build response, in this case, it makes sense to make this one compute.

internal/provider/data_source_packer_image.go
```diff
@@ -33,16 +35,23 @@ func dataSourcePackerImage() *schema.Resource {
...
+				ExactlyOneOf: []string{"channel"},
+			},
+			"channel": {
+				Description:  "The channel that points to the version of the image being retrieved. Either this or `iteration_id` must be specified.",
```
nywilken, line 51:
```suggestion
				Description:  "The channel that points to the iteration of the image being retrieved.",
```

internal/provider/data_source_packer_image.go
```diff
@@ -111,9 +121,49 @@ func dataSourcePackerImageRead(ctx context.Context, d *schema.ResourceData, meta
...
+			channel.Slug,
+			channel.Iteration.ID,
+			iteration.ID))
+	}
+
+	// Assuming we passed the above check, the rest of the channel is not
+	// used after that,
+	if channel != nil {
+		iteration = channel.Iteration
 	}
```
nywilken, line 167: By using the helper function you can get rid of this as we will already have the iteration information we are looking for.

internal/provider/data_source_packer_image.go
```diff
@@ -33,16 +35,23 @@ func dataSourcePackerImage() *schema.Resource {
...
 				Description: "Region this image is stored in, if any.",
 				Type:        schema.TypeString,
 				Required:    true,
 			},
+			// Optional inputs
+			"iteration_id": {
+				Description:  "HCP ID of this image. Either this or `channel' must be specified.",
+				Type:         schema.TypeString,
+				Optional:     true,
+				ExactlyOneOf: []string{"channel"},
```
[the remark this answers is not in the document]
lbajolet-hashicorp: Not in this case as ConflictsWith would not raise an error if both are missing

docs/data-sources/packer_image.md
```diff
@@ -38,11 +38,12 @@ output "packer-registry-ubuntu" {
 
 - `bucket_name` (String) The slug of the HCP Packer Registry image bucket to pull from.
 - `cloud_provider` (String) Name of the cloud provider this image is stored-in.
-- `iteration_id` (String) HCP ID of this image.
 - `region` (String) Region this image is stored in, if any.
 
 ### Optional
 
+- `channel` (String) Channel that promotes the latest iteration of the image. Either this or `iteration_id` must be specified.
```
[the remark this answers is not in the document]
lbajolet-hashicorp: This was already corrected by @sylviamoss in a previous review, the latest reads as "The channel that points to the version of the image being retrieved.", are you OK with their phrasing?

internal/provider/data_source_packer_image.go
```diff
@@ -33,16 +35,23 @@ func dataSourcePackerImage() *schema.Resource {
...
-				Description: "HCP ID of this image.",
-				Type:        schema.TypeString,
-				Required:    true,
-			},
 			"region": {
 				Description: "Region this image is stored in, if any.",
 				Type:        schema.TypeString,
 				Required:    true,
 			},
+			// Optional inputs
```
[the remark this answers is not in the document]
lbajolet-hashicorp: This comment is just here for the sake of consistency as we already have some at lines 25 (`// Required inputs`), and 54 (`// computed outputs`), I think if we remove this one, the rest should go too for the same consistency reasons, what do you think?

internal/provider/data_source_packer_image.go
```diff
@@ -33,16 +35,23 @@ func dataSourcePackerImage() *schema.Resource {
...
+			"iteration_id": {
+				Description:  "HCP ID of this image. Either this or `channel' must be specified.",
```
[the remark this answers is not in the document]
lbajolet-hashicorp: Same as for the channel, the latest revision has this fixed with @sylviamoss's suggestion, now reads "The iteration from which to get the image.", is this OK with you?

internal/provider/data_source_packer_image.go
lbajolet-hashicorp: I agree this one can be promoted to both `Optional` and `Compute` since it's either copied from the input, or derived from the channel, I'll do this if the Schema lets me

internal/provider/data_source_packer_image.go
```diff
@@ -33,16 +35,24 @@ func dataSourcePackerImage() *schema.Resource {
...
+				Type:         schema.TypeString,
+				Optional:     true,
+				ExactlyOneOf: []string{"channel"},
```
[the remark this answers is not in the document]
lbajolet-hashicorp: The problem is slightly more complex here, it fails at plan/apply time (there's a test for the literals-thing, I'm adding one for the static + reference) but not necessarily at validation-time currently, as it only works when none of the fields are computed.
According to @bflad the terraform-plugin-framework will let us write custom validators that we can use for this kind of use-case, but as it stands we cannot do better I'm afraid.

internal/provider/data_source_packer_image.go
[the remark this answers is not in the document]
nywilken: Got all is good. I was looking at an outdated commit.

internal/provider/data_source_packer_image_test.go
```diff
@@ -134,3 +170,78 @@ func TestAcc_dataSourcePackerImage_revokedIteration(t *testing.T) {
 		},
 	})
 }
+
+func TestAcc_dataSourcePackerImage_channelAndIterationIDReject(t *testing.T) {
+	fingerprint := "rejectIterationAndChannel"
+	configs := []string{
+		testAccPackerImageBothChanAndIter,
+		testAccPackerImageBothChanAndIterRef,
```
sylviamoss, line 178: Nice!

internal/provider/data_source_packer_image_test.go
```diff
@@ -134,3 +170,78 @@ func TestAcc_dataSourcePackerImage_revokedIteration(t *testing.T) {
...
+
+						upsertRegistry(t)
+						upsertBucket(t, acctestArchImageBucket)
+						upsertIteration(t, acctestArchImageBucket, fingerprint)
+						itID, err := getIterationIDFromFingerPrint(t, acctestArchImageBucket, fingerprint)
+						if err != nil {
+							t.Fatal(err.Error())
+						}
+						upsertBuild(t, acctestArchImageBucket, fingerprint, itID)
+						createChannel(t, acctestArchImageBucket, acctestImageChannel, itID)
```
sylviamoss, line 202: Can this be done outside `resource.Test` before the loop? These can be done just once and it will work for both test cases.

sylviamoss: Oh, I think I had a similar problem before! No problem then. Thanks for having a look.

sylviamoss: Oh, I think I had a similar problem before! No problem then. Thanks for having a look.

docs/data-sources/packer_image.md
```diff
@@ -24,25 +26,52 @@ data "hcp_packer_image" "foo" {
...
+}
+
+output "packer-registry-ubuntu-east-1-alt" {
+  value = data.hcp_packer_image.baz.cloud_image_id
+}
 ```
 
 ~> **Note:** This data source only returns the first found image's metadata filtered by the given schema values, from the returned list of images associated with the specified iteration. Therefore, if multiple images exist in the same region, it will only pick one of them. If that's the case, you may consider separating your builds into different buckets.
 
+~> **Note:** The `channel` attribute in this data source will incur a billable request to HCP Packer. This attribute is intended for convenience when using a single image. When sourcing multiple images from a single iteration, the separate data source object for the iteration alternative is preferred for saving costs.
```
sylviamoss, line 60: What do you think? 
```suggestion
~> **Note:** The `channel` attribute in this data source will incur a billable request to HCP Packer. This attribute is intended for convenience when using a single image. When sourcing multiple images from a single iteration, the `hcp_packer_iteration` data source 
is the alternative for querying a channel just once.
```

docs/data-sources/packer_image.md
```diff
@@ -24,25 +26,52 @@ data "hcp_packer_image" "foo" {
...
+data "hcp_packer_image" "baz" {
+  bucket_name    = "hardened-ubuntu-16-04"
+  cloud_provider = "aws"
+  channel        = "production"
+  region         = "us-east-1"
+}
+
+output "packer-registry-ubuntu-east-1-alt" {
+  value = data.hcp_packer_image.baz.cloud_image_id
+}
```
sylviamoss, line 55:
```suggestion

### Alternatively, if sourcing a single image

data "hcp_packer_image" "baz" {
  bucket_name    = "hardened-ubuntu-16-04"
  cloud_provider = "aws"
  channel        = "production"
  region         = "us-east-1"
}

output "packer-registry-ubuntu-east-1-alt" {
  value = data.hcp_packer_image.baz.cloud_image_id
}
```

I'm having a hard time adding the code block because the markdown is getting confused with this comment markdown. I'm suggesting moving the alternative into a separated sub-section and keeping both examples separated because all together looks big and hard to quickly see how it works.

sylviamoss (approved):
I suggested some changes to the doc, otherwise LGTM 👍🏼  
Great work!

PR #368, opened 2022-08-11
Update README.md

leehambley:
Fix resource names in example code.

Easy doc fix. Tested against

```
hcp = {
      source  = "hashicorp/hcp"
      version = "~> 0.39.0"
    }
```


bcmdarroch (approved):
Thanks for the fix! 🎉

Issue #393, opened 2022-09-13
Peering Data Sources prevent Destroys if in a FAILED State


jjtimmons:
### Terraform Version and Provider Version

```
→ terraform --version
Terraform v1.0.11

HCP provider version: latest
```

### Affected Resource(s)

* https://registry.terraform.io/providers/hashicorp/hcp/latest/docs/resources/azure_peering_connection
* https://registry.terraform.io/providers/hashicorp/hcp/latest/docs/resources/aws_network_peering

### Terraform Configuration Files

```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key: https://keybase.io/hashicorp
```

### Debug Output

```
TestE2E_LongLivedAzure 2022-09-13T17:06:24Z logger.go:66: Error: error waiting for peering connection (azure-e2e-c38853c6-peer) to become 'ACTIVE': unexpected state 'FAILED', wanted target 'ACTIVE'. last error: %!s(<nil>)
TestE2E_LongLivedAzure 2022-09-13T17:06:24Z logger.go:66: 
TestE2E_LongLivedAzure 2022-09-13T17:06:24Z logger.go:66:   with module.hcp_peering.data.hcp_azure_peering_connection.peering,
TestE2E_LongLivedAzure 2022-09-13T17:06:24Z logger.go:66:   on .terraform/modules/hcp_peering/main.tf line 117, in data "hcp_azure_peering_connection" "peering":
TestE2E_LongLivedAzure 2022-09-13T17:06:24Z logger.go:66:  117: data "hcp_azure_peering_connection" "peering" {
TestE2E_LongLivedAzure 2022-09-13T17:06:24Z logger.go:66: 
TestE2E_LongLivedAzure 2022-09-13T17:06:24Z retry.go:99: Returning due to fatal error: FatalError{Underlying: error while running command: exit status 1; 
Error: error waiting for peering connection (azure-e2e-c38853c6-peer) to become 'ACTIVE': unexpected state 'FAILED', wanted target 'ACTIVE'. last error: %!s(<nil>)

  with module.hcp_peering.data.hcp_azure_peering_connection.peering,
  on .terraform/modules/hcp_peering/main.tf line 117, in data "hcp_azure_peering_connection" "peering":
 117: data "hcp_azure_peering_connection" "peering" {
}
[...]

PR #394, opened 2022-09-14
Prevent FAILED peerings from failing deletes

jjtimmons:
### :ship: Release Note

```release-note
* {affected resource | docs | all}: {Brief description of change} [GH-nnnn]
```

```

...
```

### DIFF
--- a/internal/provider/data_source_aws_network_peering.go
+++ b/internal/provider/data_source_aws_network_peering.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"log"
 
-	networkmodels "github.com/hashicorp/hcp-sdk-go/clients/cloud-network/preview/2020-09-07/models"
 	sharedmodels "github.com/hashicorp/hcp-sdk-go/clients/cloud-shared/v1/models"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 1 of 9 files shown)


jjtimmons:
closes https://github.com/hashicorp/terraform-provider-hcp/issues/393

internal/provider/data_source_aws_network_peering.go
```diff
@@ -107,31 +108,63 @@ func dataSourceAwsNetworkPeeringRead(ctx context.Context, d *schema.ResourceData
...
+	// Skip waiting.
+	if !waitForActive || peering.State == models.HashicorpCloudNetwork20200907PeeringStateACTIVE {
+		return nil
 	}
-	d.SetId(url)
 
-	return nil
+	// If it's in a state where it could later become ACTIVE, wait.
+	waitState := false
+	for _, state := range clients.WaitForPeeringToBeActiveStates {
```
chapmanc, line 138: Switch statement? This looks awkward.

internal/provider/data_source_aws_network_peering.go
```diff
@@ -107,31 +108,63 @@ func dataSourceAwsNetworkPeeringRead(ctx context.Context, d *schema.ResourceData
...
+		if state == string(peering.State) {
+			waitState = true
+			break
+		}
+	}
+
+	// If it's not in a state that we should wait on, issue a warning and bail.
+	if !waitState {
```
chapmanc, line 146: So if it's in FAILING or DELETING state ?

internal/provider/data_source_aws_network_peering.go
```diff
@@ -107,31 +108,63 @@ func dataSourceAwsNetworkPeeringRead(ctx context.Context, d *schema.ResourceData
...
+		return []diag.Diagnostic{{
+			Severity: diag.Warning,
+			Summary:  fmt.Sprintf("Peering is in an unexpected state, connections may fail: %q", string(peering.State)),
+			Detail:   "Expected a CREATING, PENDING_ACCEPTANCE, ACCEPTED, or ACTIVE state",
+		}}
+	}
+
+	// Store resource data again, updating Peering state.
+	var result []diag.Diagnostic
+	peering, err = clients.WaitForPeeringToBeActive(ctx, client, peering.ID, hvnID, loc, peeringCreateTimeout)
```
chapmanc, line 156: What happens if we transition from PENDING -> FAILED? The previous state will say that it's PENDING and then a destroy will try to read it above (without refresh false) and error out ?

internal/provider/data_source_aws_transit_gateway_attachment.go
```diff
@@ -101,23 +102,56 @@ func dataSourceAwsTransitGatewayAttachmentRead(ctx context.Context, d *schema.Re
...
 	}
 
-	return nil
+	// Skip waiting.
+	if !waitForActive || tgwAtt.State == models.HashicorpCloudNetwork20200907TGWAttachmentStateACTIVE {
+		return nil
+	}
+
+	// If it's in a state where it could later become ACTIVE, wait.
+	waitState := false
```
chapmanc, line 124: Nit: Wait state is a bit confusing of a name. Can we reverse the logic and say terminalState and then define the terminal states?

chapmanc (approved):
I think I am good with the logic but I am curious how you tested this? Did you create a failed peering on purpose and try to resolve it?

internal/provider/data_source_aws_network_peering.go
jjtimmons: I started with that but was then re-defining the states we already list in a slice in the peering clients package, which was more awkward imo

internal/provider/data_source_aws_network_peering.go
jjtimmons: yes

internal/provider/data_source_aws_network_peering.go
jjtimmons: 1. tf apply with PENDING -> FAILED
    - we still error out here in the read
2. tf apply again:
    - if we reached this again, we'll throw a warning because we won't wait. In practice I don't expect this to happen, because we'll (normally) have a resource that will see the `FAILED` state, mark the peering as tainted, and create a new peering in its place
3. tf destroy:
    - if we read this we'll just get the warning

internal/provider/data_source_aws_transit_gateway_attachment.go
jjtimmons: yeah sounds good

internal/provider/data_source_aws_network_peering.go
```diff
@@ -127,25 +129,42 @@ func dataSourceAwsNetworkPeeringRead(ctx context.Context, d *schema.ResourceData
...
 
 	// Skip waiting.
-	if !waitForActive {
+	if !waitForActive || peering.State == models.HashicorpCloudNetwork20200907PeeringStateACTIVE {
 		return nil
 	}
 
 	// If it's in a state where it could later become ACTIVE, wait.
-	for _, state := range clients.WaitForPeeringToBeActivePendingStates {
+	waitState := false
```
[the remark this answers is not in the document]
jjtimmons: I'm a bit torn on DRY'ing this into the Waiters because of the shenanigans I'm trying to do here where:
- if it's a terminal state (FAILED/DELETING) at the _start_ of the Read, throw a Warning
- if it's a terminal state after trying to _wait_ in the Read, throw an Error

Also this is a specific (hack?) for the Peering data resources that I wouldn't want to add to `WaitForPeeringToBeAccepted` for example because we always want it to throw an Error there: https://github.com/hashicorp/terraform-provider-hcp/blob/441597cc026ffc6ebd02bd149a953f77fbfd110c/internal/provider/resource_hvn_peering_connection.go#L156

I'm not opposed necessarily, but because of two reasons above I left it within the data_resource_* Reads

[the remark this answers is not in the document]
bcmdarroch: Hmm I see, that does make sense. In the interest of getting this in, we can leave it like this for now, and revisit whenever we add another peering resource.

bcmdarroch (approved):

PR #425, opened 2022-12-05
Add vnet to resource ref

dane:

There was a line where the VNet resource name was missing. I've addressed the typo so the examples can be copied and run without modification.


bcmdarroch (approved):
🎉 thanks!

PR #427, opened 2022-12-08
Increase `hcp_consul_cluster` delete timeout to 35 min

jjtimmons:

In our E2E testing, we see Consul cluster delete times occasionally a bit above 25 minutes. Those clusters delete successfully, but the TF provider errors out on context cancellation. This increases the timeout from 25 minutes to 35 minutes, matching the Create timeout we already have.

I have not ran acceptance testing because this only increases the delete timeout value to match the existing create timeout, no other behavior changed.

```

...
```

### DIFF
--- a/.changelog/427.txt
+++ b/.changelog/427.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+Increase `hcp_consul_cluster` create timeout to 35 minutes
+```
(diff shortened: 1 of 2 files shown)


clly:
We can definitely set the delete timeout to match the create timeout. There is also a timeout block that can get set on the resource itself to configure the timeout: https://registry.terraform.io/providers/hashicorp/hcp/latest/docs/resources/consul_cluster#timeouts

clly (approved):

bcmdarroch (approved):

PR #454, opened 2023-02-09
Hce 339 feat multi project

JolisaBrownHashiCorp:

This PR includes changes to support adding project ID at the provider level. When no project ID is specified, the provider selects as default the ID of the first attached project. 

```

...
```


internal/provider/provider.go
```diff
@@ -148,14 +173,20 @@ func getProjectFromCredentials(ctx context.Context, client *clients.Client) (*mo
...
+			Detail: `The oldest project has been selected as the default. To configure which project is used as default, 
+			set a project in the HCP provider config block. Resources may also be configured with different projects.`,
+		})
+		return getOldestProject(listProjResp.Payload.Projects), diags
 	}
-
-	project := listProjResp.Payload.Projects[0]
-	return project, nil
+	project = listProjResp.Payload.Projects[0]
+	return project, diags
```
JolisaBrownHashiCorp, line 189: Do we need here to still grab the 0th index of returned projects from the payload if existing projects are only length 1? Or would it make more sense to scrap those two lines and always use the return from our getOldestProject call?

bcmdarroch: I'd say yes, mainly so we only throw the warning when more than one project is found. Maybe for the provider developer's context, throw in a comment about defaulting to the single project if only one is found.

bcmdarroch (approved):
:shipit: 🎉 😅

docs/guides/multi-project-migration-guide.md
```diff
@@ -0,0 +1,93 @@
...
+  cluster_id = "test-cluster"
+  hvn_id     = hcp_hvn.test.hvn_id
+  project_id = "0f8c263e-8eb4-4a7f-a0cc-7e476afb9fd2"
+  tier       = "development"
+}
+```
+
+### Override provider project with resource project
+
+Projects may be set at both the resource and provider level. The resource-configured project is always preferred over the provider-configured project.
```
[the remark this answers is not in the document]
apriofrost: To clarify: this is not a blocker, but a suggested change, to follow the Provider’s config best practice.

tobias-hashicorp (approved):

aoripov (approved):

aoripov (approved):

hanshasselberg (approved):

davidsbond (approved):

uruemu (approved):

JenGoldstrich (approved):

graham1228 (approved):

internal/provider/resource_vault_cluster.go
```diff
@@ -65,6 +66,14 @@ func resourceVaultCluster() *schema.Resource {
 				ValidateDiagFunc: validateSlugID,
 			},
 			// Optional fields
+			"project_id": {
+				Description:  "The ID of the HCP project where the Vault cluster is located.",
```
bosouza, line 70: maybe it'd useful to say that if this parameter is not specified then the oldest project will be targeted

bosouza (approved):

mkam (approved):

internal/provider/resource_vault_cluster.go
dmalch-hashicorp: Thank you for your feedback, I've created a ticket to address it in a separate PR https://hashicorp.atlassian.net/browse/HCPF-644. Since it is a large change, I will discuss it with the team and we will implement it separately.

PR #495, opened 2023-04-06
Implement Project datasource

delores-hashicorp:
Implement project data source

```

...
```


internal/provider/data_source_project.go
```diff
@@ -0,0 +1,102 @@
...
+
+	if err := setProjectResourceData(d, project); err != nil {
+		return diag.FromErr(err)
+	}
+
+	return nil
+}
+
+func setProjectResourceData(d *schema.ResourceData, project *models.HashicorpCloudResourcemanagerProject) error {
+	if err := d.Set("project_id", project.ID); err != nil {
```
bcmdarroch, line 89: This is causing this error: `Error: Invalid address to set: []string{"project_id"}`. Leave this as is, then replace `id` with `project_id` above.

internal/provider/data_source_project.go
```diff
@@ -0,0 +1,102 @@
...
+		Schema: map[string]*schema.Schema{
+			// Required inputs
+			"name": {
+				Description: "The name of the HCP project.",
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+			},
+			// Optional inputs
+			"id": {
```
bcmdarroch, line 35: This should be `project_id`, since Terraform reserves `id` for its own unique identifier. I would expect this to be a required attribute, since you can only fetch a project by ID right?

internal/provider/data_source_project.go
```diff
@@ -0,0 +1,102 @@
...
+		Timeouts: &schema.ResourceTimeout{
+			Default: &projectDefaultTimeout,
+		},
+		Schema: map[string]*schema.Schema{
+			// Required inputs
+			"name": {
+				Description: "The name of the HCP project.",
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
```
bcmdarroch, line 32: Terraform can't destroy/recreate resources, so no need for the ForceNew attribute.
```suggestion

```

internal/provider/data_source_project.go
```diff
@@ -0,0 +1,102 @@
...
+func dataSourceProject() *schema.Resource {
+	return &schema.Resource{
+		Description: "The project data source provides information about an existing HashiCorp project.",
+		ReadContext: dataSourceProjectRead,
+		Timeouts: &schema.ResourceTimeout{
+			Default: &projectDefaultTimeout,
+		},
+		Schema: map[string]*schema.Schema{
+			// Required inputs
+			"name": {
```
bcmdarroch, line 28: I'm guessing `name` should probably be computed, since it can't be used to fetch a project.

delores-hashicorp: `id` is the only field that can be used to fetch a project, should it be a required field then?

internal/provider/data_source_project.go
delores-hashicorp: That's right

internal/provider/data_source_project.go
delores-hashicorp: Thanks!

internal/provider/data_source_project.go
```diff
@@ -0,0 +1,102 @@
...
+		Timeouts: &schema.ResourceTimeout{
+			Default: &projectDefaultTimeout,
+		},
+		Schema: map[string]*schema.Schema{
+			// Required inputs
+			"project_id": {
+				Description:      "The ID of the HCP project.",
+				Type:             schema.TypeString,
+				Required:         true,
+				ValidateDiagFunc: validateSlugID,
```
tobias-hashicorp, line 32: ⚪️ The project ID should be expected to be a UUID not a SlugID. We might however want to skip client side validation here to allow us to easier switch to a different identifier format in the future.

docs/data-sources/project.md
```diff
@@ -0,0 +1,47 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "hcp_project Data Source - terraform-provider-hcp"
+subcategory: ""
+description: |-
+  The project data source provides information about an existing HashiCorp project.
```
tobias-hashicorp, line 6:
```suggestion
  The project data source provides information about an existing project in HCP.
```

docs/data-sources/project.md
```diff
@@ -0,0 +1,47 @@
...
+---
+
+# hcp_project (Data Source)
+
+The project data source provides information about an existing HashiCorp project.
```
tobias-hashicorp, line 11:
```suggestion
The project data source provides information about an existing project in HCP.
```

tobias-hashicorp (approved):

PR #536, opened 2023-06-22
HPR-1170: Add `hcp_packer_bucket_names` data source. 

aidan-mundy:

- Adds `hcp_packer_bucket_names` data source.
- Adds documentation for more automated management of channels.

```
$ make testacc TESTARGS='-run=TestAcc.*PackerBucket.* -test.v'
TF_ACC=1 go test ./internal/... -v -run=TestAcc.*PackerBucket.* -test.v -timeout 360m -parallel=10
PASS
ok      github.com/hashicorp/terraform-provider-hcp/internal/clients    (cached) [no tests to run]
PASS
PASS
=== RUN   TestAcc_dataSourcePackerBucketNames
--- PASS: TestAcc_dataSourcePackerBucketNames (21.94s)
PASS
ok      github.com/hashicorp/terraform-provider-hcp/internal/provider   22.467s
```


docs/guides/packer-channel-management.md
```diff
@@ -49,3 +49,208 @@ resource "hcp_packer_channel_assignment" "alpine_advanced" {
...
+  for_each = data.hcp_packer_bucket_names.all.names
+
+  name        = "prod"
+  bucket_name = each.key
+}
+```
+
+Optionally, you can use Terraform functions/expressions to filter the list of bucket names before providing it to `for_each` if channels should only be created for a subset of buckets.
+
+### Managing channel assignment for automatically created channels
```
nywilken, line 67: There's a lot going on in this guide. If there a preferred option that users should take maybe we should show that. 

Wondering if we should show the base case and have a separate guide for more advance usage. What do you think?

aidan-mundy: We could split assignment into a separate guide but it would end up being something like "Advanced Packer Channel Assignment for Channels Automatically Created From a List of Bucket Names". It really goes hand in hand with the idea of automatically created channels.

If I were to choose one of the approaches as the "preferred" option, I'd go with Example 3, specifically Option 1a. However, I think that Example 3 Option 2 is much more appropriate for large deployments.

Example 3, Option 1b could potentially be removed. I think the usecase is limited, but I included it for a good reason: Since buckets cannot be managed with Terraform, deleting a bucket (through the UI/API) will cause Terraform plans to fail. Resolution requires one of these two options:
- Removing the bucket's override from the config file
  - With Git oriented Terraform workflows (as are common with TFC/E), this change would need to be applied to all in-progress PRs until it is applied in the default branch, otherwise speculative planning will fail on all PRs. 
-

aidan-mundy: If I were to choose one of the approaches as the "preferred" option, I'd go with Example 3, specifically Option 1a. However, I think that Example 3 Option 2 is much more appropriate for large deployments where the number of buckets may become too unwieldy to manually create an override for every bucket. Also, per-bucket overrides increase the chance of merge conflicts.

Example 1 is not as robust of a configuration, but allows an easier on-ramp for situations where some of the channels already exist (and have existing non-zero assignments), where the user is not ready to have Terraform-managed assignments for a substantial portion of the channels, or where the user will spread assignment management across many files/workspaces. I wouldn't consider it "preferred" because it is less robust and doesn't provide a simple path to automatically manage assignments for newly created channels. 

Example 2 is really just a tee-up for Example 3, and isn't necessary if we're trying to cut down on length.

Example 3, Option 1b could be removed. I think the usecase is somewhat limited, but I included it for an important reason: Since buckets cannot be managed with Terraform, deleting a bucket (through the UI/API) will cause Terraform plans to fail when using Example 3, Option 1a. Resolution requires one of these two options:
- Removing the bucket's override from the config file
  - With Git oriented Terraform workflows (as are common with TFC/E), this change would need to be applied to all in-progress PRs until it is applied in the default branch, otherwise speculative planning will fail on all PRs.
    - This is a headache and could cause merge conflicts
- Re-create the bucket (possibly only temporarily until the override can be removed from the default branch)
  - This requires an intervention outside of Terraform, which is far from ideal
  - This cannot be done via the UI, and cannot be done from OSS without also running a build AFAIK
    - Terraform practitioners may not have an environment configured that is capable of creating a build that would be appropriate for the bucket, requiring them to push a dummy build. 
If Terraform was capable of managing buckets, this behavior would not be an issue because a fully-Terraform-managed deployment would simply re-create the deleted bucket as part of the plan/apply process. In that scenario, I'd firmly advise Example 3 Option 1a over Option 1b. 

It's hard for me to pick a base case, because it depends a lot on what kind of user you are and what size your deployment is. There isn't a single right answer to "manage the assignment for channel `x` on every single bucket in the registry", and managing _less_ than every single bucket adds _more_ technical complications.

sylviamoss: I agree this is too advanced and is proposing several ways of using terraform functions with the resource, which might not always be the case and can be worked around by learning how to use those functions. So I would handle those complex cases I think a simple example of promoting an iteration from one channel to the other, using some simple for_each filter should be enough to address our target use cases. 

The advanced use cases are making it hard to find simple information on how to unset a channel, for example. 

Agree that Example 3 can go, maybe entirely. What about adding a warning instead about the situation of channel override? Also if a bucket gets deleted, terraform will fail with drifted resources. We could also warn that.

docs/data-sources/packer_bucket_names.md
```diff
@@ -0,0 +1,40 @@
...
+### Optional
+
+- `project_id` (String) The ID of the HCP project where the HCP Packer registry is located.
+- `timeouts` (Block, Optional) (see [below for nested schema](#nestedblock--timeouts))
+
+### Read-Only
+
+- `id` (String) The ID of this resource.
+- `names` (Set of String) A set of names for all buckets in the HCP Packer registry.
+- `organization_id` (String) The ID of the organization where the HCP Packer registry is located.
```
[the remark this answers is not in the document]
aidan-mundy: `project_id` is an optional input, so it shows up only in the "Optional" category, but can still be read if it is not provided

docs/guides/packer-channel-management.md
aidan-mundy: @sylviamoss's [comment on Example 2](https://github.com/hashicorp/terraform-provider-hcp/pull/536#discussion_r1242490285) caused me to realize that both Example 2 and Example 3 Option 2 didn't make a lick of sense. Not sure what I was thinking with those two. I took swing at a revision to cut down on content, repetition, and `for_each` explainers. 

- Reduced to a single example
  - Removed Example 1 Option 1 (individual resources)
    - Effectively the same as assigning an iteration on a manually created channel, which already has an example
  - Removed Example 1 Option 2 (one resource with `for_each`, individual assignments)
    - Simple enough for a user to achieve this on their own. The question of "automatic adoption for all channels" is more important to focus on. 
  - Removed Example 2 (one resource with `for_each`, single value for all assignments)
    - Didn't make sense
  - Merged Example 3, Options 1a, 1b and 2
    - Trimmed boilerplate
    - Made the filtering from Option 2 make sense
    - Replaced Option 1b with a disclaimer about deleted buckets causing plan failures
- Added a link to code in the repo for an example module 
  - Includes code for some of the customization that was causing verbosity in the previous revision of the guide, and some other creature comforts:
    - Leaving select channels unmanaged
    - Requiring explicit configurations for select channels
    - Ignoring invalid bucket names in the configuration
    - Automatically assigning an iteration fetched from another channel
  - I figure this could've been another guide page, but a user looking to have this level of functionality should be able to understand the module code without a guide, and this way it's less "in your face" for users looking for a basic functionality explainer

sylviamoss (approved):
👍🏼  LGTM!

docs/guides/packer-channel-management.md
```diff
@@ -49,3 +49,71 @@ resource "hcp_packer_channel_assignment" "alpine_advanced" {
...
+    },
+    { # Individual channel assignments
+      "bucket1" : "01H1SF9NWAK8AP25PAWDBGZ1YD"
+      "bucket2" : "01H28NJ7WPCZA0FZZ8G3FGGTAF"
+      "bucket3" : data.hcp_packer_iteration.bucket3_staging.id
+    }
+  )
+
+  bucket_name = each.key
+  # Using a reference for `channel_name` allows it to be generated dynamically
```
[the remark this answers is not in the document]
aidan-mundy: It won't be filtered here, but it will generate an error if a bucket filtered out of `hcp_packer_channel.release` is present in the for_each for the assignment. This shows that the for_each input contains invalid overrides. 

Here, "generated dynamically" means that you could use logic in the for_each of `hcp_packer_channel.release` to not have the same channel names for every bucket. For example, a user could set it up so that buckets with names ending in `-internal` could get the channel name "internal-release", where all other buckets would get the channel name "release".

docs/guides/packer-channel-management.md
```diff
@@ -49,3 +49,71 @@ resource "hcp_packer_channel_assignment" "alpine_advanced" {
...
+  # in the `hcp_packer_channel.release` resource if desired.
+  channel_name = hcp_packer_channel.release[each.key].name
+
+  iteration_id = each.value
+}
+```
+
+You are not required to configure an assignment for every channel at the same time as automatic channel creation. Using Terraform builtin functions/expressions, it is possible to limit which channels should have assignments configured by default.
+
+The default value used in the example is `"none"`, which causes Terraform to set the channel to have no assigned iteration and continue to ensure that an iteration is not assigned elsewhere. Each of the three iteration identifier attributes has a "zero value" for this purpose: `iteration_id` and `iteration_fingerprint` both use `"none"`, and `iteration_version` uses `0`. The default value can also be set to `null` if you want to ensure that every channel has an assignment defined explicitly rather than using a default.
```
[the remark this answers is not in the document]
aidan-mundy: Using a default value of `"none"`, all channel assignments are automatically adopted into Terraform management, but channels without an explicitly defined assignment will use that default value (`"none"`. 

Using a default value of `null`, channel assignments will not be automatically adopted into Terraform management, but Terraform will force the practitioner to explicitly set an assignment for each channel in the `for_each` map. This handles the case where both of these are true:
1. the user wants to mandate that assignments for all automatically created channels must be managed by Terraform
2. it is undesirable for channels to be given a default value

An example of where item 2 could apply: Assignments are being handled by a list of filters, and a bucket appears that does not match any of those filters. The user may not want the assignment to be defaulted to `"none"`, and would instead prefer to define a matching filter (or otherwise resolve the discrepancy) before the apply can succeed. In this case, removing the default entirely would solve the issue of not wanting to apply a default value, but would not allow the user to enforce the mandate in item 1. 

The TL;DR explanation is:
- No default
  - Equivalent to explicit assignment being "optional" (
   - If the user doesn't define an assignment for a bucket, assignment will not be managed by Terraform
- Default of `"none"`
  - Equivalent to explicit assignment being "optional" 
    - If the user doesn't define an assignment for a bucket, assignment is enforced as "unassigned" by Terraform 
- Default of `null`
  - Equivalent to explicit assignment being "required" 
    - If the user doesn't define an assignment for a bucket, throws an error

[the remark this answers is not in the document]
sylviamoss: Ohh got it! I was missing the adoption part when interpreting this one. Thanks for explaining!

PR #558, opened 2023-07-24
Add concurrency group to test workflow

aidan-mundy:

- Add concurrency group to test workflow
  - Prevents multiple acceptance testing jobs from running at the same time, which could cause unexpected outcomes.


sylviamoss (approved):
Good idea!

PR #575, opened 2023-08-07
hcp_vault_cluster resource changes for adding vault plugins

blaumer:

The terraform provider for HCP vault will be updated so new optional stanzas will be added to the 'hcp_vault_cluster' terraform resource, so plugins can be specified to be added when using terraform to provision HCP Vault clusters.

A new 'vault_plugin' resource will be created, which can be specified as a subresource which can be optionally added. This will allow for a list of vault_plugin resources to be specified to indicate the plugins a user wants to install on their cluster.

There will be validation for valid plugin_type. Validation is not done for plugin_name. This is because valid plugin names are stored in Launch darkly, which would require a cloud-sdk dependency in the provider, which I feel like would be discouraged as this is a public repo.

During the acceptance test, I used two different plugin configs. 

config1:  
```
vault_plugin {
		plugin_type = "SECRET"
		plugin_name = "venafi-pki-backend"
	}
```
	
config 2: 
```
	vault_plugin {
		plugin_type = "SECRET"
		plugin_name = "venafi-pki-backend"
	}
	vault_plugin {
		plugin_type = "DATABASE"
		plugin_name = "vault-plugin-database-oracle"
	}
```
	
Update 1: config 1 (one plugin added)
Update 2: config 2 (one new plugin added. idempotent plugin/add api called twice)
Update 3: config 1	(one plugin removed. )

The datadog API endpoints support that the plugins were being added/deleted at the correct intervals. I also confirmed on the database that the plugin_registration table matched the expected state.
Screenshot 2023-08-07 at 3 10 15 PM

Because only 1 customer managed plugin is available right now, new tests must be added in future when more plugins are available.

```
make testacc TESTARGS='-run=TestAccVaultClusterAWS'
TF_ACC=1 go test ./internal/... -v -run=TestAccVaultClusterAWS -timeout 360m -parallel=10
PASS
ok      github.com/hashicorp/terraform-provider-hcp/internal/clients    0.333s [no tests to run]
PASS
ok      github.com/hashicorp/terraform-provider-hcp/internal/consul     0.298s [no tests to run]
PASS
ok      github.com/hashicorp/terraform-provider-hcp/internal/input      0.288s [no tests to run]
=== RUN   TestAccVaultClusterAWS
=== PAUSE TestAccVaultClusterAWS
=== CONT  TestAccVaultClusterAWS
--- PASS: TestAccVaultClusterAWS (3173.02s)
PASS
ok      github.com/hashicorp/terraform-provider-hcp/internal/provider   3173.386s

...
```

### DIFF
--- a/docs/data-sources/vault_cluster.md
+++ b/docs/data-sources/vault_cluster.md
@@ -31,6 +31,7 @@ data "hcp_vault_cluster" "example" {
 If not specified, the project specified in the HCP Provider config block will be used, if configured.
 If a project is not configured in the HCP Provider config block, the oldest project in the organization will be used.
 - `timeouts` (Block, Optional) (see [below for nested schema](#nestedblock--timeouts))
+- `vault_plugin` (Block List) The external plugins to install on the vault cluster (see [below for nested schema](#nestedblock--vault_plugin))
 
 ### Read-Only
 
@@ -61,6 +62,15 @@ If a project is not configured in the HCP Provider config block, the oldest proj
 Optional:
 
 - `default` (String)
+
+
+<a id="nestedblock--vault_plugin"></a>
+### Nested Schema for `vault_plugin`
+
+Required:
+
+- `plugin_name` (String) The name of the plugin
+- `plugin_type` (String) The type of the plugin
 
 
 <a id="nestedblock--audit_log_config"></a>
--- a/docs/resources/vault_cluster.md
+++ b/docs/resources/vault_cluster.md
@@ -61,6 +61,7 @@ If a project is not configured in the HCP Provider config block, the oldest proj
 - `public_endpoint` (Boolean) Denotes that the cluster has a public endpoint. Defaults to false.
 - `tier` (String) Tier of the HCP Vault cluster. Valid options for tiers - `dev`, `starter_small`, `standard_small`, `standard_medium`, `standard_large`, `plus_small`, `plus_medium`, `plus_large`. See [pricing information](https://www.hashicorp.com/products/vault/pricing). Changing a cluster's size or tier is only available to admins. See [Scale a cluster](https://registry.terraform.io/providers/hashicorp/hcp/latest/docs/guides/vault-scaling).
 - `timeouts` (Block, Optional) (see [below for nested schema](#nestedblock--timeouts))
+- `vault_plugin` (Block List) The external plugins that are to be installed on the vault cluster (see [below for nested schema](#nestedblock--vault_plugin))
 
 ### Read-Only
 
@@ -126,6 +127,15 @@ Optional:
 - `default` (String)
 - `delete` (String)
 - `update` (String)
+
+
+<a id="nestedblock--vault_plugin"></a>
+### Nested Schema for `vault_plugin`
+
+Required:
+
+- `plugin_name` (String) The name of the plugin - Valid options for plugin name - 'venafi-pki-backend'
+- `plugin_type` (String) The type of the plugin - Valid options for plugin type - 'SECRET', 'AUTH', 'DATABASE'
 
 -> **Note:** When establishing performance replication links between clusters in different HVNs, an HVN peering connection is required. This can be defined explicitly using an [`hcp_hvn_peering_connection`](hvn_peering_connection.md), or HCP will create the connection automatically (peering connections can be imported after creation using [terraform import](https://www.terraform.io/cli/import)). Note HVN peering [CIDR block requirements](https://cloud.hashicorp.com/docs/hcp/network/routes#cidr-block-requirements).
 
--- a/docs/resources/vault_cluster.md
+++ b/docs/resources/vault_cluster.md
@@ -61,7 +61,7 @@ If a project is not configured in the HCP Provider config block, the oldest proj
 - `public_endpoint` (Boolean) Denotes that the cluster has a public endpoint. Defaults to false.
 - `tier` (String) Tier of the HCP Vault cluster. Valid options for tiers - `dev`, `starter_small`, `standard_small`, `standard_medium`, `standard_large`, `plus_small`, `plus_medium`, `plus_large`. See [pricing information](https://www.hashicorp.com/products/vault/pricing). Changing a cluster's size or tier is only available to admins. See [Scale a cluster](https://registry.terraform.io/providers/hashicorp/hcp/latest/docs/guides/vault-scaling).
 - `timeouts` (Block, Optional) (see [below for nested schema](#nestedblock--timeouts))
-- `vault_plugin` (Block List) The external plugins that are to be installed on the vault cluster (see [below for nested schema](#nestedblock--vault_plugin))
+- `vault_plugin` (Block List) The external plugins to install on the vault cluster (see [below for nested schema](#nestedblock--vault_plugin))
 
 ### Read-Only
 
--- a/internal/provider/resource_vault_cluster.go
+++ b/internal/provider/resource_vault_cluster.go
@@ -277,7 +277,7 @@ If a project is not configured in the HCP Provider config block, the oldest proj
 				},
 			},
 			"vault_plugin": {
-				Description: "The external plugins that are to be installed on the vault cluster",
+				Description: "The external plugins to install on the vault cluster",
 				Type:        schema.TypeList,
 				Optional:    true,
 				Computed:    true,
@@ -333,7 +333,6 @@ If a project is not configured in the HCP Provider config block, the oldest proj
 }
 
 func resourceVaultClusterCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
-
 	client := meta.(*clients.Client)
 
 	clusterID := d.Get("cluster_id").(string)
@@ -361,7 +360,8 @@ func resourceVaultClusterCreate(ctx context.Context, d *schema.ResourceData, met
 	if diagErr != nil {
 		return diagErr
 	}
-	pluginConfig, diagErr := getPluginConfig(d)
+
+	pluginConfig, diagErr := getPluginConfig(d, nil)
 	if diagErr != nil {
 		return diagErr
 	}
@@ -557,7 +557,13 @@ func resourceVaultClusterCreate(ctx context.Context, d *schema.ResourceData, met
 		}
 	}
 
-	if err := setVaultClusterResourceData(d, cluster); err != nil {
+	plugins, err := clients.ListPlugins(ctx, client, loc, clusterID)
+	if err != nil {
+		log.Printf("[ERROR] Vault cluster (%s) failed to list plugins", clusterID)
+		return diag.FromErr(err)
+	}
+
+	if err := setVaultClusterResourceData(d, cluster, plugins.Plugins); err != nil {
 		return diag.FromErr(err)
 	}
 
@@ -595,8 +601,14 @@ func resourceVaultClusterRead(ctx context.Context, d *schema.ResourceData, meta 
 		return nil
 	}
 
+	plugins, err := clients.ListPlugins(ctx, client, loc, clusterID)
+	if err != nil {
+		log.Printf("[ERROR] Vault cluster (%s) failed to list plugins", clusterID)
+		return diag.FromErr(err)
+	}
+
 	// Cluster found, update resource data.
-	if err := setVaultClusterResourceData(d, cluster); err != nil {
+	if err := setVaultClusterResourceData(d, cluster, plugins.Plugins); err != nil {
 		return diag.FromErr(err)
 	}
 
@@ -648,7 +660,14 @@ func resourceVaultClusterUpdate(ctx context.Context, d *schema.ResourceData, met
 		return diagErr
 	}
 
-	newPluginConfig, diagErr := getPluginConfig(d)
+	// get plugins for plugin-name validation in getPluginConfig
+	plugins, err := clients.ListPlugins(ctx, client, loc, clusterID)
+	if err != nil {
+		log.Printf("[ERROR] Vault cluster (%s) failed to list plugins", clusterID)
+		return diag.FromErr(err)
+	}
+
+	newPluginConfig, diagErr := getPluginConfig(d, plugins.Plugins)
 	if diagErr != nil {
 		return diagErr
 	}
@@ -765,7 +784,13 @@ func resourceVaultClusterUpdate(ctx context.Context, d *schema.ResourceData, met
 		}
 	}
 
-	if err := setVaultClusterResourceData(d, cluster); err != nil {
+	plugins, err = clients.ListPlugins(ctx, client, loc, clusterID)
+	if err != nil {
+		log.Printf("[ERROR] Vault cluster (%s) failed to list plugins", clusterID)
+		return diag.FromErr(err)
+	}
+
+	if err := setVaultClusterResourceData(d, cluster, plugins.Plugins); err != nil {
 		return diag.FromErr(err)
 	}
 
@@ -892,8 +917,7 @@ func getClusterTier(d *schema.ResourceData) *string {
 }
 
 // setVaultClusterResourceData sets the KV pairs of the Vault cluster resource schema.
-func setVaultClusterResourceData(d *schema.ResourceData, cluster *vaultmodels.HashicorpCloudVault20201125Cluster) error {
-
+func setVaultClusterResourceData(d *schema.ResourceData, cluster *vaultmodels.HashicorpCloudVault20201125Cluster, plugins []*vaultmodels.HashicorpCloudVault20201125PluginRegistrationStatus) error {
 	if err := d.Set("cluster_id", cluster.ID); err != nil {
 		return err
 	}
@@ -1020,6 +1044,20 @@ func setVaultClusterResourceData(d *schema.ResourceData, cluster *vaultmodels.Ha
 				return err
 			}
 		}
+	}
+
+	var pluginConfig []map[string]any
+	for _, plugin := range plugins {
+		if plugin.IsRegistered {
+			pluginMap := map[string]any{}
+			pluginMap["plugin_name"] = plugin.PluginName
+			pluginMap["plugin_type"] = plugin.PluginType
+			pluginConfig = append(pluginConfig, pluginMap)
+		}
+		if err = d.Set("vault_plugin", pluginConfig); err != nil {
+			return err
+		}
+
 	}
 
 	return nil
@@ -1235,7 +1273,7 @@ func flattenMajorVersionUpgradeConfig(config *vaultmodels.HashicorpCloudVault202
 	return []interface{}{configMap}
 }
 
-func getPluginConfig(d *schema.ResourceData) ([]*vaultmodels.HashicorpCloudVault20201125AddPluginRequest, diag.Diagnostics) {
+func getPluginConfig(d *schema.ResourceData, plugins []*vaultmodels.HashicorpCloudVault20201125PluginRegistrationStatus) ([]*vaultmodels.HashicorpCloudVault20201125AddPluginRequest, diag.Diagnostics) {
 	if !d.HasChange("vault_plugin") {
 		return nil, nil
 	}
@@ -1262,12 +1300,20 @@ func getPluginConfig(d *schema.ResourceData) ([]*vaultmodels.HashicorpCloudVault
 		}
 		pluginName := config["plugin_name"].(string)
 		pluginType := config["plugin_type"].(string)
+
+		if plugins != nil {
+			err := validateVaultPluginName(pluginName, pluginType, plugins)
+			if err != nil {
+				return nil, err
+			}
+		}
+
 		pluginConfigs = append(pluginConfigs, &vaultmodels.HashicorpCloudVault20201125AddPluginRequest{
 			PluginName: pluginName,
 			PluginType: pluginType,
 		})
-
 	}
+
 	return pluginConfigs, nil
 }
 
--- a/internal/provider/resource_vault_cluster.go
+++ b/internal/provider/resource_vault_cluster.go
@@ -276,6 +276,33 @@ If a project is not configured in the HCP Provider config block, the oldest proj
 					},
 				},
 			},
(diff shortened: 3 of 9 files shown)


internal/provider/resource_vault_cluster.go
```diff
@@ -276,6 +276,33 @@ If a project is not configured in the HCP Provider config block, the oldest proj
...
 			},
+			"vault_plugin": {
+				Description: "The external plugins that are to be installed on the vault cluster",
+				Type:        schema.TypeList,
+				Optional:    true,
+				Computed:    true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"plugin_name": {
+							Description: "The name of the plugin - Valid options for plugin name - 'venafi-pki-backend'",
```
helenfufu, line 287: Question re: the comment in PR description about not validating for `plugin_name` because it would depend on a flag - what's the behavior here if we pass an invalid option?

Is there validation in cloud-vault-service API for the valid plugin names based on the LD flag? And then the expected behavior when you pass an invalid one here to just fail and HCP TFP would return a validation error from the API?

hashiblaum: the api request would fail on the  cloud-vault-service API , which would cause the terraform here to fail.

Fast failing would be really nice. Thinking we can use the PluginStatus API on vault-service to get valid plugin names. going to look into this tomorrow 

thanks for calling out on the behavior here

hashiblaum: looking closer it seems tough to make network calls in the validation functions here

helenfufu: ooh yeah FWIW I think it's fine to rely on the API to fail (and just let terraform fail), was just double checking that's actually what happens here. Prob not necessary to add a validation function here to call the API - if it fails at the cloud-vault-service API I think it will be quick enough already 😄 

Relying on cloud-vault-service API to fail is also how some of the validation for observability configs works - e.g. in [this customer escalation](https://hashicorp.atlassian.net/browse/HVT-4116?focusedCommentId=277162), and I _think_ historically it's also been how gating on Azure worked during Azure beta/GA/Plus GA since we only ever added those checks in the API and not here (from looking at commit history).

internal/provider/resource_vault_cluster.go
```diff
@@ -680,6 +723,48 @@ func resourceVaultClusterUpdate(ctx context.Context, d *schema.ResourceData, met
...
+		// add all plugins in new plugin config
+		for _, plugin := range newPluginConfig {
+			_, err := clients.AddPlugin(ctx, client, clusterLocationShared, clusterID, plugin)
+			if err != nil {
+				return diag.Errorf("error adding plugin (%s) to Vault cluster (%s): %v", plugin.PluginName, clusterID, err)
+			}
+		}
+	}
+
 	if err := setVaultClusterResourceData(d, cluster); err != nil {
```
helenfufu, line 768: Do you need to update the `setVaultClusterResourceData` function to set the vault plugin info there as well? https://github.com/hashicorp/terraform-provider-hcp/blob/678ec3bd0dad3914655b6a0881f68a9c9ae88158/internal/provider/resource_vault_cluster.go#L810

I was planning to do this for the proxy changes since we appear to set all the fields there (calling the above function) at the end of create, update, and read.

helenfufu (commented):
Couple tiny comments plus a question about `setVaultClusterResourceData`. Rest of the changes make sense to me and testing looks 👍

docs/resources/vault_cluster.md
```diff
@@ -61,6 +61,7 @@ If a project is not configured in the HCP Provider config block, the oldest proj
 - `public_endpoint` (Boolean) Denotes that the cluster has a public endpoint. Defaults to false.
 - `tier` (String) Tier of the HCP Vault cluster. Valid options for tiers - `dev`, `starter_small`, `standard_small`, `standard_medium`, `standard_large`, `plus_small`, `plus_medium`, `plus_large`. See [pricing information](https://www.hashicorp.com/products/vault/pricing). Changing a cluster's size or tier is only available to admins. See [Scale a cluster](https://registry.terraform.io/providers/hashicorp/hcp/latest/docs/guides/vault-scaling).
 - `timeouts` (Block, Optional) (see [below for nested schema](#nestedblock--timeouts))
+- `vault_plugin` (Block List) The external plugins that are to be installed on the vault cluster (see [below for nested schema](#nestedblock--vault_plugin))
```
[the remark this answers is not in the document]
hashiblaum: yeah can do

thinking `(Block List, No Limit)` - thoughts?

internal/provider/resource_vault_cluster.go
hashiblaum: i honestly didn't pay too much attention to that function and just left it as the last call. I can move it before the plugin stuff, if that makes it more clear

helenfufu: Oh I think it's ok to leave it as the last call, but I was wondering if you need to update the contents of the function itself? Like inside the function, we set each individual key to value of `cluster.Field`, example: https://github.com/hashicorp/terraform-provider-hcp/blob/678ec3bd0dad3914655b6a0881f68a9c9ae88158/internal/provider/resource_vault_cluster.go#L848-L867

docs/resources/vault_cluster.md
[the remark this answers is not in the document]
helenfufu: sounds good!

internal/provider/resource_vault_cluster.go
helenfufu: Did a little more digging. I think in order to set a new field in `setVaultClusterResourceData`, you'd also need to specify the new field in `data_source_vault_cluster` - [this file](https://github.com/hashicorp/terraform-provider-hcp/blob/main/internal/provider/data_source_vault_cluster.go#L18) which corresponds to the data source specification [here](https://registry.terraform.io/providers/hashicorp/hcp/latest/docs/data-sources/vault_cluster). 

In my own [branch](https://github.com/hashicorp/terraform-provider-hcp/compare/hvt-4182-http-proxy-for-hcpv-in-hcp-tfp?expand=1) for proxy changes (feel free to check it out but it's def still a WIP), I originally tried adding my fields in `setVaultClusterResourceData` without adding them to the data source, which failed. After adding to data source, setting the new fields in `setVaultClusterResourceData` worked.

docs/resources/vault_cluster.md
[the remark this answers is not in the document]
hashiblaum: wait, I think this is autogenerated and will eventually be overwritten. unfortunately i think this will have to stay

internal/provider/resource_vault_cluster.go
helenfufu: Ah I see your point now, thanks for clarifying. You can't rely on the create and update APIs for validation in the same way the other fields do.

Simple but maybe hacky idea: could you add a validation function here that just validates based on a hard-coded list (similar to how we check whether enum values are valid)? And every time we add more plugins we would update that?

internal/provider/resource_vault_cluster.go
hashiblaum: good callout, I can try to add it

internal/provider/resource_vault_cluster.go
hashiblaum: I was thinking about that, but would really prefer not to maintain a hard-coded list, when it is available over API. I'll try to dig a bit more to see if the API call is possible

internal/provider/validators.go
```diff
@@ -441,3 +441,22 @@ func validateBoundaryPassword(v interface{}, path cty.Path) diag.Diagnostics {
...
 }
+
+func validateVaultPluginType(v interface{}, path cty.Path) diag.Diagnostics {
+	var diagnostics diag.Diagnostics
+
+	err := vaultmodels.HashicorpCloudVault20201125PluginType(strings.ToUpper(v.(string))).Validate(strfmt.Default)
+	if err != nil {
+		enumList := regexp.MustCompile(`\[.*\]`).FindString(err.Error())
+		expectedEnumList := strings.ToLower(enumList)
+		msg := fmt.Sprintf("expected '%v' to be one of: %v", v, expectedEnumList)
```
helenfufu, line 452: Sorry keep finding little things as I work on my own changes

Nit: this is a pattern that affects all our existing enum validations, but the way we currently validate also surfaces the "invalid" enum as an allowed value in our error message.

Encountered this when I was testing on my own branch, so I got an error message like
```
Error: expected '' to be one of: [http_proxy_option_invalid enabled disabled]

          with hcp_vault_cluster.test,
          on terraform_plugin_test.tf line 15, in resource "hcp_vault_cluster" "test":
          15: 	proxy_endpoint     = ""
```
when I passed the wrong proxy option.

Guessing you would get something similar with
```
expected %v to be one of: [plugin_type_invalid, secret, auth, database]
```

I don't think this is a blocking comment since it already affects so many of our other enums (all the MVU ones), but if you wanted to address this on yours, I ended up doing something like 

<details><summary>this example</summary>

```go
func validateVaultClusterProxyEndpoint(v interface{}, path cty.Path) diag.Diagnostics {
	var diagnostics diag.Diagnostics

	err := vaultmodels.HashicorpCloudVault20201125HTTPProxyOption(strings.ToUpper(v.(string))).Validate(strfmt.Default)
	if err != nil {
		enumList := regexp.MustCompile(`\[(.*)\]`).FindStringSubmatch(err.Error())
		expectedEnumList := strings.ToLower(enumList[1])

		// Remove invalid option from allowed list in error message
		expectedEnumList = strings.ReplaceAll(
			expectedEnumList,
			strings.ToLower(string(vaultmodels.HashicorpCloudVault20201125HTTPProxyOptionHTTPPROXYOPTIONINVALID)),
			"",
		)

		// Format as comma-separated list
		expectedEnumList = strings.Join(strings.Fields(expectedEnumList), ", ")

		msg := fmt.Sprintf("expected '%v' to be one of: %v", v, expectedEnumList)
		diagnostics = append(diagnostics, diag.Diagnostic{
			Severity:      diag.Error,
			Summary:       msg,
			Detail:        msg + " (value is case-insensitive).",
			AttributePath: path,
		})
	}

	return diagnostics
}

```

</details>

which gives an error message like `expected '' to be one of: enabled, disabled` instead of the one above.

helenfufu, line 452: Sorry keep finding little things as I work on my own changes

Nit: this is a pattern that affects all our existing enum validations, but the way we currently validate 
1.  surfaces the "invalid" enum as an allowed value in our error message and 
2. actually doesn't block the invalid enum as an allowed value.

For 1, I encountered this when I was testing on my own branch, so I got an error message like
```
Error: expected '' to be one of: [http_proxy_option_invalid enabled disabled]

          with hcp_vault_cluster.test,
          on terraform_plugin_test.tf line 15, in resource "hcp_vault_cluster" "test":
          15: 	proxy_endpoint     = ""
```
when I passed the wrong proxy option.

Guessing you would get something similar with
```
expected %v to be one of: [plugin_type_invalid, secret, auth, database]
```

I don't think this is a blocking comment since it already affects so many of our other enums (all the MVU ones), but if you wanted to address this on yours, I ended up doing something like 

<details><summary>this example</summary>

```go
func validateVaultClusterProxyEndpoint(v interface{}, path cty.Path) diag.Diagnostics {
	var diagnostics diag.Diagnostics

	err := vaultmodels.HashicorpCloudVault20201125HTTPProxyOption(strings.ToUpper(v.(string))).Validate(strfmt.Default)
	if err != nil {
		enumList := regexp.MustCompile(`\[(.*)\]`).FindStringSubmatch(err.Error())
		expectedEnumList := strings.ToLower(enumList[1])

		// Remove invalid option from allowed list in error message
		expectedEnumList = strings.ReplaceAll(
			expectedEnumList,
			strings.ToLower(string(vaultmodels.HashicorpCloudVault20201125HTTPProxyOptionHTTPPROXYOPTIONINVALID)),
			"",
		)

		// Format as comma-separated list
		expectedEnumList = strings.Join(strings.Fields(expectedEnumList), ", ")

		msg := fmt.Sprintf("expected '%v' to be one of: %v", v, expectedEnumList)
		diagnostics = append(diagnostics, diag.Diagnostic{
			Severity:      diag.Error,
			Summary:       msg,
			Detail:        msg + " (value is case-insensitive).",
			AttributePath: path,
		})
	}

	return diagnostics
}

```

</details>

which gives an error message like `expected '' to be one of: enabled, disabled` instead of the one above.

For 2, I decided not to block the invalid enum in the code above, since clusters are currently creating on the invalid option (since it's not populated), and invalid is actually handled in BE / treated as disabled.

Summary: I think it's worth making the error message less confusing, but actually blocking invalid enums might break existing clusters so I avoided doing that.

internal/provider/resource_vault_cluster.go
hashiblaum: updated

internal/provider/validators.go
hashiblaum: updated to remove invalid from the list

invalid isn't in the valid values in the docs and is denied by API , so i think it is fine as well

internal/provider/resource_vault_cluster.go
```diff
@@ -610,6 +660,18 @@ func resourceVaultClusterUpdate(ctx context.Context, d *schema.ResourceData, met
...
 	}
 
+	// get plugins for plugin-name validation in getPluginConfig
+	plugins, err := clients.ListPlugins(ctx, client, loc, clusterID)
+	if err != nil {
+		log.Printf("[ERROR] Vault cluster (%s) failed to list plugins", clusterID)
+		return diag.FromErr(err)
+	}
+
+	newPluginConfig, diagErr := getPluginConfig(d, plugins.Plugins)
```
himran92, line 670: I was a little unclear in start of why we are passing listed cluster plugins here to ensure that new plugin is part of cluster plugin. But then realized that listed cluster plugin can have all valid plugins which may be not registered by the cluster too (the response isRegistered will tell you which are part of the cluster yet). So this way you ensure that the new plugin is at least a right plugin name. 
If my understanding is correct then can we highlight this point in a comment here + also add a comment on top of ListPlugins of its response having both registered/non-registered thus all plugins.

Thinking how we can have  a similar plugin name validation on create time too. Does vault service has an API to expose the plugin name to UI for opting in/out on create? if yes, can we use that to validate here to make both flows similar?

internal/provider/resource_vault_cluster.go
```diff
@@ -276,6 +276,33 @@ If a project is not configured in the HCP Provider config block, the oldest proj
...
+							Required:    true,
+							DiffSuppressFunc: func(_, old, new string, _ *schema.ResourceData) bool {
+								return strings.EqualFold(old, new)
+							},
+						},
+						"plugin_type": {
+							Description:      "The type of the plugin - Valid options for plugin type - 'SECRET', 'AUTH', 'DATABASE'",
+							Type:             schema.TypeString,
+							Required:         true,
+							ValidateDiagFunc: validateVaultPluginType,
```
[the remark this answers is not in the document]
himran92: I also have a relative question in another comment on `how to do plugin name validation on create` :)

docs/data-sources/vault_cluster.md
```diff
@@ -31,6 +31,7 @@ data "hcp_vault_cluster" "example" {
 If not specified, the project specified in the HCP Provider config block will be used, if configured.
 If a project is not configured in the HCP Provider config block, the oldest project in the organization will be used.
 - `timeouts` (Block, Optional) (see [below for nested schema](#nestedblock--timeouts))
+- `vault_plugin` (Block List) The external plugins to install on the vault cluster (see [below for nested schema](#nestedblock--vault_plugin))
```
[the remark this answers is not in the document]
hashiblaum: I have `vault_plugin` because it is specified as an individual resource.

The implementation would look like this in pseduo-terraform

```
resource `hcp_vault_cluster` {
       vault_plugin {
		plugin_type = "SECRET"
		plugin_name = "plugin1"
	}
	vault_plugin {
		plugin_type = "DATABASE"
		plugin_name = "plugin2"
	}
}
```

since each plugin is specified is a separate `vault_plugin` block, that is why I have singular `vault_plugin`

internal/provider/resource_vault_cluster.go
hashiblaum: yes I can add comment, that would be very helpful

Vault service only has plugin APIs on the cluster level, so a cluster resource must exist on vault-service end first. (which is why we cant do the validation).

The options I see are:
1. Hard-coded validation. (personally not a fan , but can be ok if we aren't adding plugins frequently)
2. new vault-service API to serve plugins from feature flag for the LD rule of that organization.
3. What we are doing now where the terraform will fail when the plugin fails to add. (I believe the cluster will still be running and it should be doable for the user to update their terraform resource)
4. Making new `vault_plugin` resource that is not nested in `hcp_vault_cluster`, so it can fail independently of cluster creation and it requires an `hcp_vault_cluster` resource

internal/provider/resource_vault_cluster.go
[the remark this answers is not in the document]
hashiblaum: can definitely update comments to make this more clear

answered question in other thread : )

internal/provider/resource_vault_cluster_test.go
```diff
@@ -307,6 +308,8 @@ func updateVaultPublicEndpointObservabilityDataAndMVU(t *testing.T, in *inputT)
 			resource.TestCheckResourceAttrSet(in.VaultClusterResourceName, "audit_log_config.0.datadog_api_key"),
 			resource.TestCheckResourceAttr(in.VaultClusterResourceName, "audit_log_config.0.datadog_region", "us1"),
 			resource.TestCheckResourceAttr(in.VaultClusterResourceName, "major_version_upgrade_config.0.upgrade_type", "MANUAL"),
+			resource.TestCheckResourceAttr(in.VaultClusterResourceName, "vault_plugin.0.plugin_name", "venafi-pki-backend"),
+			resource.TestCheckResourceAttr(in.VaultClusterResourceName, "vault_plugin.0.plugin_type", "SECRET"),
```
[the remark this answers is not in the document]
hashiblaum: so that test case isn't included yet because there is only 1 customer managed plugin to work with. I did some hacky things to test 2 of them. Once more plugins are added, we can update the tests.

hafsa had the same confusion, sorry if this wasn't clear in pr description

docs/data-sources/vault_cluster.md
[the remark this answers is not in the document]
himran92: thanks for explaining

docs/resources/vault_cluster.md
```diff
@@ -127,6 +128,15 @@ Optional:
 - `delete` (String)
 - `update` (String)
 
+
+<a id="nestedblock--vault_plugin"></a>
+### Nested Schema for `vault_plugin`
+
+Required:
+
+- `plugin_name` (String) The name of the plugin - Valid options for plugin name - 'venafi-pki-backend'
```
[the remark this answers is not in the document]
himran92: that makes sense, thanks for explaining

internal/provider/resource_vault_cluster_test.go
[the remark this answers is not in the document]
helenfufu: Ohh gotcha. Suggestion: Could we maybe have update 3 use a config with no plugins at all and then assert `resource.TestCheckNoResourceAttr` so then acceptance tests run by anyone will test removal without needing the [extra setup](https://github.com/hashicorp/terraform-provider-hcp/pull/575#discussion_r1291595165) you did?

internal/provider/resource_vault_cluster.go
himran92: > 2. new vault-service API to serve plugins from feature flag for the LD rule of that organization.
I thought this already existed for HCP Portal but realizing may be portal is reading the values directly from FF?

> 3. What we are doing now where the terraform will fail when the plugin fails to add. (I believe the cluster will still be running and it should be doable for the user to update their terraform resource)
Hmm does this cause a concern in terraform state where cluster exists though the plugins given in input of create did not apply.

helenfufu: I feel like 3 (current) is probably alright for now given that's the behavior on the UI - like you have to have created the cluster _first_ before you can add plugins, and if that flow fails, your cluster is still running.

> it should be doable for the user to update their terraform resource

Yup, users should be able to update the state using `terraform refresh` if it fails. (Not sure if you're aware, but if you wanted, you can test this and any real creates/updates in prod with your changes in local binary following [this guide](https://github.com/hashicorp/terraform-provider-hcp/blob/main/contributing/breaking-changes.md#how-to-test-existing-resources).)

I think long-term 2 or 4 would make more sense, 2 being an easier change and 4 being a more difficult one + maybe separating out other resources for consistency too.

himran92: For 4 -> There was a FR in talks in another channel where the user was showing the desire to manage the state of configuration like observability, separate than cluster resource in terraform. So it may be that in some long term future we actually do 4 option for all major configurations. 

For now, agree that this flow seems fine.

helenfufu (approved):
Thanks for addressing all comments - approving since
- we all seem to [agree](https://github.com/hashicorp/terraform-provider-hcp/pull/575#discussion_r1291688373) that the late create validation is acceptable for now and we can revisit down the line
- I just have one remaining [test suggestion](https://github.com/hashicorp/terraform-provider-hcp/pull/575#discussion_r1291657596) that I think is worth doing to make sure removal is tested on every acceptance test without special setup, but imo not blocking

internal/provider/resource_vault_cluster.go
hashiblaum: >> 2. new vault-service API to serve plugins from feature flag for the LD rule of that organization.

>I thought this already existed for HCP Portal but realizing may be portal is reading the values directly from FF?

The HCP portal API does use the FF, but the API also requires a valid cluster_id (which is why it can't be called before create).

going to stick with #3 for now.

hashiblaum:
closing due to making new terraform `hcp_vault_cluster` resource in #579

PR #586, opened 2023-08-30
Add support for elasticsearch observability on vault_cluster resource

jaireddjawed:

This PR adds ElasticSearch as an option for observability providers on HCP Vault clusters. I tested this change by adding a unit test that verifies that all required params that are a part of this provider are defined (located in `internal/provider/resource_vault_cluster_config_test.go`.

```

...
```


internal/provider/resource_vault_cluster.go
```diff
@@ -1031,6 +1110,45 @@ func flattenObservabilityConfig(config *vaultmodels.HashicorpCloudVault20201125O
...
+			if configParam, ok := d.GetOk(propertyName); ok && len(configParam.([]interface{})) > 0 {
+				config := configParam.([]interface{})[0].(map[string]interface{})
+				configMap["cloudwatch_secret_access_key"] = config["cloudwatch_secret_access_key"].(string)
+			}
+		}
+
+		if elasticsearch := config.Elasticsearch; elasticsearch != nil {
+			configMap["elasticsearch_endpoint"] = elasticsearch.Endpoint
+			configMap["elasticsearch_user"] = elasticsearch.User
+
```
jaireddjawed, line 1139: I noticed that there was a field called `dataset` on the `config.ElasticSearch` struct. Does that go here as well? Should it only be set for streaming metrics or audit logs?

internal/provider/data_source_vault_cluster.go
bosouza, line 1: I believe all fields you added in this file should be only `Computed`, no `Optional` or `Sensitive`. Although tbh I'm not entirely sure if we should be adding in here fields that can return from the API as `redacted`, so it's possible that my addition of `cloudwatch_secret_access_key` was incorrect.

internal/provider/resource_vault_cluster.go
```diff
@@ -287,6 +308,27 @@ If a project is not configured in the HCP Provider config block, the oldest proj
 							Type:        schema.TypeString,
 							Computed:    true,
 						},
+						"elasticsearch_endpoint": {
+							Description: "ElasticSearch endpoint for streaming audit logs",
+							Type:        schema.TypeString,
+							Optional:    true,
+						},
+						"elasticsearch_dataset": {
+							Description: "ElasticSearch dataset for streaming metrics",
```
bosouza, line 317:
```suggestion
							Description: "ElasticSearch dataset for streaming audit logs",
```

internal/provider/resource_vault_cluster.go
```diff
@@ -1212,6 +1275,25 @@ func getValidObservabilityConfig(config map[string]interface{}) (*vaultmodels.Ha
...
+		if elasticsearchEndpoint == "" || elasticsearchDataset == "" || elasticsearchUser == "" || elasticsearchPassword == "" {
+			missingParamErr = diag.Errorf("elasticsearch configuration is invalid: configuration information missing")
+		}
+
+		observabilityConfig = &vaultmodels.HashicorpCloudVault20201125ObservabilityConfig{
+			Elasticsearch: &vaultmodels.HashicorpCloudVault20201125Elasticsearch{
+				Endpoint: elasticsearchEndpoint,
+				User:     elasticsearchUser,
+				Password: elasticsearchPassword,
+				Dataset:  elasticsearchDataset,
```
bosouza, line 1292: doesn't this break updates? I'd ask you to manually test the creation of a cluster with elasticserach configured as an observability provider for merics and audit logs, then update the settings and apply the terraform changes. That would allow you to validate that the Create and Update methods work fine with the new provider, as well as fetching info from the server and transforming it to the TF representation

internal/provider/data_source_vault_cluster.go
jaireddjawed: Would the `elasticsearch_password` field be `Computed` and not `Sensitive`?

internal/provider/resource_vault_cluster.go
jaireddjawed: How can I test this update? I attempted to use the `hcloud hcp dev run` command in the root directory of this project to see if I can push the updates to my PRDE, but I get an error stating that I cannot use this command because it requires that the repository has a makefile in its root directory.

bosouza: a TF provider like this one for HCP is not a control-plane service that you'd run in PRDE, it's a module for Terraform that our customers import in their `.tf` files.

To test your changes you'd have to use the Terraform module like a customer would, but consuming the locally built provider instead of allowing Terraform to fetch from the Terraform registry. For instructions on how to build this provider see https://github.com/hashicorp/terraform-provider-hcp/tree/main/contributing#building-the-provider

For instructions on how to use the `vault_cluster` resource see https://registry.terraform.io/providers/hashicorp/hcp/latest/docs/resources/vault_cluster. There are also instructions for authenticating https://registry.terraform.io/providers/hashicorp/hcp/latest/docs/guides/auth

Also, I didn't have success trying to get Terraform to target anything other than the prod env, so I'd suggest to just create the resources in prod

Feel free to reach out if it's still unclear how you can manually verify your changes

internal/provider/data_source_vault_cluster.go
bosouza: whoops, got sidetracked when thinking about the response here. Tbh I'm not 100% sure how this TF data source is supposed to be used, but I think it should expose all fields that can be retrieved by the backend. To me it makes sense that all of them are Computed, maybe it's also ok to set some as sensitive. But maybe we should also omit these credential fields that are redacted from the backend, as the original providers did somewhat consistently.

internal/provider/resource_vault_cluster.go
jaireddjawed: I finished writing the terraform script and it's included here for convienience.

```
terraform {
  required_providers {
    hcp = {
      source  = "localhost/providers/hcp"
      version = "0.0.1"
    }
  }
}

provider "hcp" {
  client_id = "fBBIue3fpg7nrV44b6Ej7rCAvSyPKIQR"
  client_secret = "dToqp6oq5CtmWwOoz78SfdISG95fYgvKppt4jXIi_66ekeJWMcfG4QR2fIq6Mu7B"
  project_id = "1fd2e0dc-6243-4468-9999-13576fe2c548"
}

provider "aws" {
  region="us-west-2"
}

resource "hcp_hvn" "example" {
  hvn_id         = "awshvn"
  cloud_provider = "aws"
  region         = "us-west-2"
  cidr_block     = "172.25.16.0/20"
}

  cluster_id = "vault-cluster"
  hvn_id     = hcp_hvn.example.hvn_id
  tier = "standard_small"
  audit_log_config {
    elasticsearch_user="user"
    elasticsearch_password="password"
    elasticsearch_endpoint="https://elasticsearch"
  }
}
```

I get the following error that I believe is expected because elasticsearch may not be fully implemented as an observability provider as of now?

`Error: elasticsearch configuration is invalid: configuration information missing`.

jaireddjawed: I mistakenly added `elasticsearch_dataset` as a user-validated field. After updating, this works as intended.

bosouza: so my concern here wasn't even the issue you hit and fixed, but instead that by setting `Dataset: elasticsearchDataset` you would cause the Dataset field to be set in an update request, which should cause the server to reject the request because that field cannot be set by the caller. While I think that the right approach would be to not set the field at all if it's working then it's ok

bosouza (approved):
lgtm!

PR #590, opened 2023-09-11
Fix semver validation

riddhi89:
The existing validation prevented from specifying a pre-release version which is also a valid semver.
The PR changes it to utilize the [go-version](https://github.com/hashicorp/go-version) library helper which follows the[ SemVer](https://semver.org/) spec.

### DIFF
--- a/internal/provider/validators.go
+++ b/internal/provider/validators.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/go-openapi/strfmt"
 	"github.com/hashicorp/go-cty/cty"
+	"github.com/hashicorp/go-version"
 	consulmodels "github.com/hashicorp/hcp-sdk-go/clients/cloud-consul-service/stable/2021-02-04/models"
 	vaultmodels "github.com/hashicorp/hcp-sdk-go/clients/cloud-vault-service/stable/2020-11-25/models"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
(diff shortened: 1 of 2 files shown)


mkam (approved):

nywilken (approved):
LGTM

internal/provider/validators.go
```diff
@@ -95,8 +96,7 @@ func validateStringInSlice(valid []string, ignoreCase bool) schema.SchemaValidat
 // validateSemVer ensures a specified string is a SemVer.
 func validateSemVer(v interface{}, path cty.Path) diag.Diagnostics {
 	var diagnostics diag.Diagnostics
-
-	if !regexp.MustCompile(`^v?\d+.\d+.\d+$`).MatchString(v.(string)) {
+	if _, err := version.NewSemver(v.(string)); err != nil {
```
nywilken, line 99: This looks good to me. Given that go-version is used in the provider but not for the validation I want to confirm before merging that the consul resource is okay with a user specifying a pre-release version. In other words, is there a reason why the original implementor chose only to support official release versions?

nywilken: Disregard.

riddhi89: > I want to confirm before merging that the consul resource is okay with a user specifying a pre-release version

Might have been a miss. But we do publicly release pre-release versions of our products (can be found [here](https://releases.hashicorp.com/consul/)). On the Consul side, we optionally enable testing out these pre-release versions on HCP for certain customers to unblock their testing/validation processes.

PR #585, opened 2023-08-30
VAULT-18334: Documentation Subcategories

dhuckins:
### :hammer_and_wrench: Description

Updating docs to use sub-categories

- resources/data-sources that already had templates were updated
- resources/data-sources that did not have a template, had a template created

| Name | Subcategory |
| ------ | ------------ |
| hcp_aws_network_peering | HashiCorp Virtual Networks | 
| hcp_aws_transit_gateway_attachment | HashiCorp Virtual Networks | 
| hcp_azure_peering_connection | HashiCorp Virtual Networks | 
| hcp_boundary_cluster | HCP Boundary | 
| hcp_consul_agent_helm_config | HCP Consul | 
| hcp_consul_agent_kubernetes_secret | HCP Consul | 
| hcp_consul_cluster | HCP Consul | 
| hcp_consul_cluster_root_token | HCP Consul | 
| hcp_consul_snapshot | HCP Consul | 
| hcp_consul_versions | HCP Consul | 
| hcp_hvn | HashiCorp Virtual Networks | 
| hcp_hvn_peering_connection | HashiCorp Virtual Networks | 
| hcp_hvn_route | HashiCorp Virtual Networks | 
| hcp_packer_bucket_names | HCP Packer | 
| hcp_packer_channel | HCP Packer | 
| hcp_packer_channel_assignment | HCP Packer | 
| hcp_packer_image | HCP Packer | 
| hcp_packer_image_iteration | HCP Packer | 
| hcp_packer_iteration | HCP Packer | 
| hcp_packer_run_task | HCP Packer | 
| hcp_vault_cluster | HCP Vault | 
| hcp_vault_cluster_admin_token | HCP Vault | 
| hcp_vault_secrets_app | HCP Vault Secrets |

PR #593, opened 2023-09-13
Create App and Secret Resource in Vault Secrets

AnPucel:
This is adding a resource for managing an app and managing a secret within Vault Secrets. 

```
=== RUN   TestAccVaultSecretsResourceApp
--- PASS: TestAccVaultSecretsResourceApp (10.02s)
=== RUN   TestAccVaultSecretsResourceSecret
--- PASS: TestAccVaultSecretsResourceSecret (10.99s)
```


internal/provider/resource_vault_secrets_app.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func (r *vaultsecretsAppResource) Metadata(_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_vault_secrets_app"
+}
+
+func (r *vaultsecretsAppResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Attributes: map[string]schema.Attribute{
+			// TODO: Add validators
```
AnPucel, line 28: Per the previous PR: we should add a project id at the resource level.

internal/provider/resource_vault_secrets_app.go
```diff
@@ -0,0 +1,169 @@
...
+	"fmt"
+
+	sharedmodels "github.com/hashicorp/hcp-sdk-go/clients/cloud-shared/v1/models"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	clients "github.com/hashicorp/terraform-provider-hcp/internal/clients"
+)
+
+func NewVaultSecretsAppResource() *resourceVaultsecretsApp {
```
austinvalle, line 14: nit: typically, this function returns a `resource.Resource`, as the struct `resourceVaultsecretsApp` should implement that interface

internal/provider/provider.go
```diff
@@ -63,7 +63,14 @@ func (p *ProviderFramework) Schema(ctx context.Context, req provider.SchemaReque
...
 func (p *ProviderFramework) Resources(ctx context.Context) []func() resource.Resource {
-	return []func() resource.Resource{}
+	return []func() resource.Resource{
+		func() resource.Resource {
+			return &resourceVaultsecretsApp{}
+		},
+		func() resource.Resource {
+			return &resourceVaultsecretsSecret{}
+		},
+	}
```
austinvalle, line 73: nit: similar to the comment below, this can just be the list of functions defined in each resource file:

```go
return []func() resource.Resource{
		NewVaultSecretsAppResource,
		NewVaultSecretsSecretResource,
	}
```

You can use any of the utility providers as an example:
- https://github.com/hashicorp/terraform-provider-dns/blob/main/internal/provider/resource_dns_cname_record.go#L29
- https://github.com/hashicorp/terraform-provider-dns/blob/b89204a34cba26de70ecbbe0864283ff17082964/internal/provider/provider_framework.go#L334-L343

internal/provider/resource_vault_secrets_app.go
```diff
@@ -0,0 +1,169 @@
...
+type resourceVaultsecretsApp struct {
+	client *clients.Client
+}
+
+func (r *resourceVaultsecretsApp) Metadata(_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_vault_secrets_app"
+}
+
+func (r *resourceVaultsecretsApp) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
```
austinvalle, line 27: I'd add some descriptions for each attribute and the resource itself, will fill out the docs for you 😃

internal/provider/resource_vault_secrets_app.go
```diff
@@ -0,0 +1,169 @@
...
+		)
+		return
+	}
+	r.client = client
+}
+
+type App struct {
+	ID          types.String `tfsdk:"id"`
+	AppName     string       `tfsdk:"app_name"`
+	Description string       `tfsdk:"description"`
```
austinvalle, line 61: This is a [common gotcha](https://github.com/hashicorp/terraform-provider-cloudinit/pull/103) for the [conversion rules](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/types/object#setting-values) of `tfsdk` tags for Go types. When you run code like:
```go
var plan App
diags := req.Plan.Get(ctx, &plan)
```

If `AppName` or `Description` is unknown, you'll get an error like:
```bash
An unexpected error was encountered trying to build a value...
Received unknown value, however the target type cannot handle unknown values. Use the corresponding types package type or a custom type that handles unknown values.
```

General rule of thumb, if it has a `tfsdk` tag, it's best to use the Framework [`types`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework@v1.4.0/types) package so you can handle unknown and null values without unexpected errors. (like you've done with `ID` 👍🏻 )

internal/provider/resource_vault_secrets_app.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	loc := &sharedmodels.HashicorpCloudLocationLocation{
+		OrganizationID: r.client.Config.OrganizationID,
+		ProjectID:      r.client.Config.ProjectID,
+	}
+
+	res, err := clients.CreateVaultSecretsApp(ctx, r.client, loc, plan.AppName, plan.Description)
+	if err != nil {
+		resp.Diagnostics.AddError(err.Error(), "Cannot create app")
```
austinvalle, line 79: I'd swap these two, as the first string is the summary, second is the detail:

```go
resp.Diagnostics.AddError("Error creating Vault Secrets App", err.Error())
```

internal/provider/resource_vault_secrets_app.go
```diff
@@ -0,0 +1,169 @@
...
+
+	plan.ID = types.StringValue(res.Name)
+	plan.AppName = res.Name
+	plan.Description = res.Description
+	diags = resp.State.Set(ctx, plan)
+	resp.Diagnostics.Append(diags...)
+
+	if resp.Diagnostics.HasError() {
+		return
+	}
```
austinvalle, line 91: nit: you'll commonly see the setting of state at the end of a CRUD function reduced to one line
```go
resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
```

internal/provider/resource_vault_secrets_app.go
AnPucel: Ahh. Good catch. Thanks

internal/provider/provider.go
AnPucel: Good call!

internal/provider/resource_vault_secrets_app.go
AnPucel: ahh! this is good to know. Thanks!

internal/provider/resource_vault_secrets_app.go
AnPucel: Will add this as a followup PR

internal/provider/resource_vault_secrets_app.go
```diff
@@ -0,0 +1,168 @@
...
+	client *clients.Client
+}
+
+func (r *resourceVaultsecretsApp) Metadata(_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_vault_secrets_app"
+}
+
+func (r *resourceVaultsecretsApp) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		MarkdownDescription: "The Vault Secrets app resource manages an application.",
```
averche, line 31: Perhaps would be better to say something like "... manages secrets for an application."

internal/provider/resource_vault_secrets_app.go
```diff
@@ -0,0 +1,168 @@
...
+				Validators: []validator.String{
+					stringvalidator.LengthBetween(3, 36),
+					stringvalidator.RegexMatches(
+						regexp.MustCompile(`^[-\da-zA-Z]{3,36}$`),
+						"must contain only letters, numbers or hyphens",
+					),
+				},
+			},
+			"description": schema.StringAttribute{
+				Description: "The Vault Secrets app name description",
```
averche, line 49:
```suggestion
				Description: "The Vault Secrets app description",
```

internal/provider/resource_vault_secrets_app.go
```diff
@@ -0,0 +1,168 @@
...
+		resp.Diagnostics.AddError(
+			"Unexpected Data Source Configure Type",
+			fmt.Sprintf("Expected *clients.Client, got: %T. Please report this issue to the provider developers.", req.ProviderData),
+		)
+		return
+	}
+	r.client = client
+}
+
+type App struct {
```
averche, line 71: This name might be too generic. Would something like `VaultSecretsApp` be better instead?

internal/provider/resource_vault_secrets_secret.go
```diff
@@ -0,0 +1,176 @@
...
+
+func NewVaultSecretsSecretResource() resource.Resource {
+	return &resourceVaultsecretsSecret{}
+}
+
+type resourceVaultsecretsSecret struct {
+	client *clients.Client
+}
+
+type Secret struct {
```
averche, line 25: Similar to above, something like `VaultSecretsSecret` might be better.

internal/provider/resource_vault_secrets_secret.go
```diff
@@ -0,0 +1,176 @@
...
+				Validators: []validator.String{
+					stringvalidator.LengthBetween(3, 36),
+					stringvalidator.RegexMatches(
+						regexp.MustCompile(`^[-\da-zA-Z]{3,36}$`),
+						"must contain only letters, numbers or hyphens",
+					),
+				},
+			},
+			"secret_name": schema.StringAttribute{
+				Description: "The name of the secret",
```
averche, line 56: It would probably be good to add validators here as well.

internal/provider/resource_vault_secrets_secret.go
```diff
@@ -0,0 +1,176 @@
...
+		ProjectID:      r.client.Config.ProjectID,
+	}
+
+	res, err := clients.CreateVaultSecretsAppSecret(ctx, r.client, loc, plan.AppName.String(), plan.SecretName.String(), plan.SecretValue.String())
+	if err != nil {
+		resp.Diagnostics.AddError("Error creating Vault Secrets Secret", err.Error())
+		return
+	}
+
+	// TODO: add more to plan here?
```
averche, line 102: Is this something to do in this PR?

internal/provider/provider.go
```diff
@@ -63,14 +63,15 @@ func (p *ProviderFramework) Schema(ctx context.Context, req provider.SchemaReque
 }
 
 func (p *ProviderFramework) Resources(ctx context.Context) []func() resource.Resource {
-	return []func() resource.Resource{}
+	return []func() resource.Resource{
+		NewVaultSecretsAppResource,
+		NewVaultSecretsSecretResource,
+	}
```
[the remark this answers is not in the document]
AnPucel: This is because we have a separate package for resources using the older framework. They can be found under the `providersSDKv2` folder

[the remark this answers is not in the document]
averche: Perhaps add a comment about this for future readers :)

internal/provider/resource_vault_secrets_app.go
AnPucel: This particular resource only allows you to manage an application. And this gets generated in specific docs for this specific resource.

internal/provider/resource_vault_secrets_secret.go
AnPucel: whoops. This is old

internal/provider/resource_vault_secrets_secret.go
AnPucel: done!

internal/provider/provider.go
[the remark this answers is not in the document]
AnPucel: Sure! There's one in the main file. But happy to add it here as well.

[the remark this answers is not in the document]
AnPucel: Okay. I added a note toward the top of this file.

internal/provider/resource_vault_secrets_secret.go
AnPucel: Yeah. Good call. I'm not sure what is allowed for secret names. I'll have to double check.

templates/resources/vault_secrets_secret.md.tmpl
```diff
@@ -0,0 +1,18 @@
+---
+page_title: "{{.Type}} {{.Name}} - {{.ProviderName}}"
+subcategory: ""
```
dhuckins, line 3:
```suggestion
subcategory: "HCP Vault Secrets"
```

support for #585

dhuckins (approved):

averche (approved):
LGTM but I don't have a lot of context around this repo.

internal/provider/resource_vault_secrets_secret.go
```diff
@@ -0,0 +1,194 @@
...
+					stringvalidator.RegexMatches(
+						regexp.MustCompile(`^[_\da-zA-Z]{3,36}$`),
+						"must contain only letters, numbers or underscores",
+					),
+				},
+			},
+			"secret_value": schema.StringAttribute{
+				Description: "The value of the secret",
+				Required:    true,
+			},
```
aidan-mundy, line 69: This should probably be a sensitive value
```suggestion
			"secret_value": schema.StringAttribute{
				Description: "The value of the secret",
				Required:    true,
				Sensitive:   true,
			},
```

templates/resources/vault_secrets_secret.md.tmpl
```diff
@@ -0,0 +1,18 @@
+---
+page_title: "{{.Type}} {{.Name}} - {{.ProviderName}}"
+subcategory: "HCP Vault Secrets"
+description: |-
+{{ .Description | plainmarkdown | trimspace | prefixlines "  " }}
+---
+
+# {{.Name}} ({{.Type}})
+
+-> **Note:** Please treat your state file as sensitive when using this resource.
```
aidan-mundy, line 10: This is standard practice for all Terraform state files when using resources that have attributes that are marked sensitive. I don't think the disclaimer is necessary, but if still desired, I'd recommend copying the one used by the `vault_cluster_admin_token` resource (includes a link to best practices).

aidan-mundy (changes requested):

templates/resources/vault_secrets_secret.md.tmpl
AnPucel: This is a decision that was made from the Vault Secrets team. I think we're just being extra cautious with folks who choose to use this particular resource.

aidan-mundy (approved):

PR #601, opened 2023-09-20
Push automatic prereleases back by 2 hours

aidan-mundy:

- Delay automatic weekly prereleases by 2 hour 
  - Will be weekly on Wednesdays at 02:00 UTC (previously 00:00 UTC)

The release process currently takes ~4 hours (the vast majority from running acceptance tests). If code is merged to `main` during the release, the release can still succeed, but leaves the `main` branch in an incorrect state that requires manual intervention. This failure mode occurred during last week's release due to a merge from a US West coast engineer just after the start of the prerelease.

Automatic prereleases currently start at 00:00 UTC, which is 17:00 PDT / 16:00 PST (at or just before the end of normal working hours). Delaying by 2 hours provides a worst case start time of 18:00 PST (6 PM west coast US), and a worst case end time of ~08:15 CEST (8:15 AM central Europe).

Not a "fix" per se, but it should work as an easy solution to minimize accidents during normal working hours.

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -1,9 +1,9 @@
 name: Prerelease
 
 on:
-  # Run Every Wednesday at 12:00 AM UTC
+  # Run Every Wednesday at 01:00 AM UTC
   schedule:
-    - cron: '0 0 * * 3'
+    - cron: '0 2 * * 3'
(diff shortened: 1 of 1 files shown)


sylviamoss (approved):
Makes sense!

PR #616, opened 2023-09-28
Fix Packer Docs

aidan-mundy:

- Fixes problems missed/introduced by #585 
  - Improper header formatting (header displays as part of page text and categorization is ignored)
  - Templates not using `import.sh` in the `examples` folder
  - Product-specific guides missing a product category


sylviamoss (approved):
I went to see the current docs and looks terrible 👀 ! Nice catch!

PR #617, opened 2023-09-28
Fix Vault Docs

aidan-mundy:

- Fixes problems missed/introduced by #585 
  - Improper header formatting (header displays as part of page text and categorization is ignored)
  - Product-specific guides missing a product category
  - Template missing for `hcp_vault_plugin` resource and datasource


templates/guides/vault-scaling.md.tmpl
```diff
@@ -1,5 +1,5 @@
 ---
-subcategory: ""
+subcategory: "HCP Vault"
```
AnPucel, line 2: Does this put the guides under the `HCP Vault` subecategory? I think we'd still want the guides to be top level

aidan-mundy: This would put this specific guide under that category, in the same way that resources and data sources work. I have left guides that are cross-product or non-product-specific at the root level, but I figure product-specific guides should go alongside the product's resources/data sources

AnPucel: Gotcha. That seems reasonable.

AnPucel (approved):

PR #626, opened 2023-10-09
Adding TF Guides for Vault Secrets

AnPucel:

Adding guides for Vault Secrets

```

...
```


templates/guides/vault-secrets-data-sources.md.tmpl
```diff
@@ -0,0 +1,17 @@
+---
+subcategory: ""
```
dhuckins, line 2: are these supposed to have a subcategory now? or just not the guides?

cc @aidan-mundy

dhuckins (approved):

templates/guides/vault-secrets-data-sources.md.tmpl
AnPucel: I think the guides need to be at the top level, otherwise it will duplicate the folders w/ resource and data source docs in them.

PR #639, opened 2023-10-17
Add support for newrelic observability on vault_cluster resource

jaireddjawed:

```

...
```

### DIFF
--- a/internal/providersdkv2/data_source_vault_cluster.go
+++ b/internal/providersdkv2/data_source_vault_cluster.go
@@ -202,6 +202,21 @@ If a project is not configured in the HCP Provider config block, the oldest proj
 							Type:        schema.TypeString,
 							Computed:    true,
 						},
+						"newrelic_account_id": {
+							Description: "NewRelic Account ID for streaming metrics",
+							Type:        schema.TypeString,
+							Computed:    true,
+						},
+						"newrelic_license_key": {
+							Description: "NewRelic license key for streaming metrics",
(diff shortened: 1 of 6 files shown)


internal/providersdkv2/data_source_vault_cluster.go
```diff
@@ -276,6 +291,21 @@ If a project is not configured in the HCP Provider config block, the oldest proj
...
+							Computed:    true,
+						},
+						"newrelic_license_key": {
+							Description: "NewRelic license key for streaming audit logs",
+							Type:        schema.TypeString,
+							Computed:    true,
+						},
+						"newrelic_region": {
+							Description: "NewRelic region for streaming audit logs",
+							Type:        schema.TypeString,
```
bosouza, line 306: we don't do any client-side validation on these and have them all as strings, which I think is fine even tho on the frontend we have enums. But can we highlight in the description that it needs to be either EU or US?

bosouza (commented):
Were you able to verify the create and update changes against the portal?

internal/providersdkv2/data_source_vault_cluster.go
jaireddjawed: Will make this change.

jaireddjawed:
> Were you able to verify the create and update changes against the portal?

Yes. I have updated the PR description showing my manual test of the update.

bosouza (approved):
lgtm!

PR #644, opened 2023-10-19
Deprecate support for Terraform < 1.1.5 (protocol v5) in advance of v6 support

aidan-mundy:

- Fixes incorrect changelog entry so deprecation will actually appear in generated changelog.


.github/workflows/changelog-check.yml
```diff
@@ -38,6 +38,9 @@ jobs:
             # "Enhancement is not a valid type of changelog entry, but it's a common mistake.
             echo "::error::Found invalid type (enhancement) in changelog - did you mean improvement?"
             exit 1
+          elif grep -q ':deprecate$' $changelog_files; then
```
dadgar, line 41: Seems like we should have a final case that checks for all known valid strings and errors if it isn't one of them.

dadgar (approved):

PR #650, opened 2023-10-20
Revert "Add support for newrelic observability on vault_cluster resource"

jaireddjawed:
The original PR that adds support for New Relic as an observability provider was merged before the provider was released. This PR reverts the original PR so that these changes aren't in `main` while the provider is not yet available


bosouza (approved):
not sure if that's how we're supposed to revert PRs in here but some change-log issues are certainly better than adding support to the unreleased provider

PR #652, opened 2023-10-24
Delete .changelog/650.txt

aidan-mundy:

Changelog not needed. The removed code was never released in a version of the provider.


bosouza (approved):
Thx Aidan, sorry I didn't notice this no-changelog tag before

PR #660, opened 2023-10-30
Add support for http observability on vault_cluster resource

jaireddjawed:
This PR adds custom HTTP endpoints as an option for observability providers on HCP Vault clusters. I tested this change by adding a unit test that verifies that all required params that are a part of this provider are defined (located in `internal/provider/resource_vault_cluster_config_test.go` and performed a manual test of these changes.

#### Requirements
* Terraform v1.6.2

#### Steps
* Ran `make dev` in the main directory of this repository
* Created the following terraform script in another folder

```
terraform {
  required_providers {
    hcp = {
      source  = "localhost/providers/hcp"
      version = "0.0.1"
    }
  }
}

provider "hcp" {
  client_id = "CLIENT_ID"
  client_secret = "CLIENT_SECRET"
  project_id = "PROJECT_ID"
}

provider "aws" {
  region="us-west-2"
}

// Create a HashiCorp Virtual Network (HVN):
resource "hcp_hvn" "terraform-hvn" {
  hvn_id         = "test-hvn"
  cloud_provider = "aws"
  region         = "us-west-2"
  cidr_block     = "172.25.16.0/20"
}

// Create HCPV cluster on AWS:
resource "hcp_vault_cluster" "tf-vault-cluster" {
  depends_on      = [hcp_hvn.terraform-hvn]
  hvn_id          = hcp_hvn.terraform-hvn.hvn_id
  cluster_id      = "test-vault-cluster"
  tier            = "standard_small"
  public_endpoint = true
  proxy_endpoint = "ENABLED"
  lifecycle {
    prevent_destroy = true
  }
  audit_log_config {
    http_uri="https://webhook.site/3eceb6c7-1007-4ce8-900a-fc41d1a400c9"
    http_codec="JSON"
    http_method="POST"
    http_headers = {
      "key1": "value1",
      "key2": "value2",
      "key3": "value3"
    }
    http_bearer_token = "111234567"
  }
}
```

* Ran the script above by running `terraform init` and `terraform apply` in the directory of the terraform script
* Verified that the script worked by looking at the created vault cluster within the portal and viewing that its audit logs provider is the http endpoint specified in the terraform script.

Screenshot 2023-11-01 at 11 35 08 AM

* Checked to see if audit log http requests are the type of request specified in the terraform script (`POST`), that the headers and bearer token are seen in the http request

Screenshot 2023-11-01 at 11 38 45 AM

### Verified that http observability information appears in `/etc/vector.d/vector.json`

Screenshot 2023-11-01 at 1 33 03 PM

### Check to see if errors are valid
* Verified that an error occurs when `http_basic_user` is given, but `http_basic_password` is not specified
```
audit_log_config {
    http_uri="https://webhook.site/3eceb6c7-1007-4ce8-900a-fc41d1a400c9"
    http_codec="JSON"
    http_method="POST"
    http_headers = {
      "key1": "value1",
      "key2": "value2",
      "key3": "value3"
    }
    http_basic_user = "user"
}
```
Screenshot 2023-11-01 at 11 36 46 AM

* Verified that an error occurs when `http_bearer_token` and `http_basic_user` are specified
```
audit_log_config {
    http_uri="https://webhook.site/3eceb6c7-1007-4ce8-900a-fc41d1a400c9"
    http_codec="JSON"
    http_method="POST"
    http_headers = {
      "key1": "value1",
      "key2": "value2",
      "key3": "value3"
    }
    http_bearer_token = "1231223"
    http_basic_user = "user"
  }
```
Screenshot 2023-11-01 at 11 42 13 AM

```

...
```


internal/providersdkv2/resource_vault_cluster.go
```diff
@@ -1290,6 +1441,54 @@ func getValidObservabilityConfig(config map[string]interface{}) (*vaultmodels.Ha
...
+			// http basic requires both the username and password to be filled
+			if httpBasicUser != "" && httpBasicPassword == "" || httpBasicUser == "" && httpBasicPassword != "" {
+				missingParamErr = diag.Errorf("http configuration is invalid: basic authentication requires username and password")
+			} else {
+				httpBasicAuth = &vaultmodels.HashicorpCloudVault20201125HTTPBasicAuth{
+					User:     httpBasicUser,
+					Password: httpBasicPassword,
+				}
+			}
+		}
```
bosouza, line 1475: nit: I think this could be refactored to make it easier to understand and avoid this case where we're setting `httpBasicAuth` even if nether `httpBasicUser` nor `httpBasicPassword` are set:
* create helper validation function, smt like `validateHttpAuth`, that allows avoiding nested if's by returning an error early
* in the function first check for token and basic at the same time, return an error
* then check for token and just set `httpBearerAuth`, that will be returned later
* then check for either of the basic auth vars, if either are set check that both are set an return an error if not
* return httpBasicAuth and httpBearerToken, which can both be nil if no auth was configured

that's just an idea tho, but I find the nested ifs a bit hard to follow

internal/providersdkv2/resource_vault_cluster.go
```diff
@@ -1290,6 +1441,54 @@ func getValidObservabilityConfig(config map[string]interface{}) (*vaultmodels.Ha
...
+				Compression:   httpCompression,
+				PayloadPrefix: httpPayloadPrefix,
+				PayloadSuffix: httpPayloadSuffix,
+				Method:        httpMethod,
+				URI:           httpURI,
+			},
+		}
+	}
+
 	if missingParamErr != nil {
```
bosouza, line 1492: I wonder if we should change this name now, it's more like `invalidProviderConfigErr` or smt. I really like the usage in this PR tho, it maintains the prioritization of "too many providers" over invalid settings from specific providers

internal/providersdkv2/resource_vault_cluster_config_test.go
bosouza, line 1: we could have more test cases for invalid codecs and with neither basic or bearer auth configured

docs/resources/vault_cluster.md
```diff
@@ -94,6 +94,16 @@ Optional:
 - `grafana_endpoint` (String) Grafana endpoint for streaming audit logs
 - `grafana_password` (String, Sensitive) Grafana password for streaming audit logs
 - `grafana_user` (String) Grafana user for streaming audit logs
+- `http_basic_password` (String, Sensitive) HTTP basic authentication password for streaming audit logs
+- `http_basic_user` (String) HTTP basic authentication username for streaming audit logs
+- `http_bearer_token` (String, Sensitive) HTTP bearer authentication token for streaming audit logs
```
bosouza, line 99: I think we should highlight in the description how these work, like if you have user you need password, and if you have either of them you can't have token

docs/resources/vault_cluster.md
```diff
@@ -94,6 +94,16 @@ Optional:
...
+- `http_codec` (String) HTTP codec for streaming audit logs
```
bosouza, line 100: I think we should list the acceptable values, and also validated it

internal/providersdkv2/resource_vault_cluster_config_test.go
jaireddjawed: I believe that authentication was optional, so neither basic nor bearer auth configured would result in no error.

jaireddjawed: I will add a test case for invalid codecs, however.

docs/data-sources/vault_cluster.md
```diff
@@ -81,6 +81,16 @@ Read-Only:
 - `elasticsearch_user` (String) ElasticSearch user for streaming audit logs
 - `grafana_endpoint` (String) Grafana endpoint for streaming audit logs
 - `grafana_user` (String) Grafana user for streaming audit logs
+- `http_basic_password` (String) HTTP basic authentication password for streaming audit logs, one of the two available authentication methods, can be specified only if http_bearer_token is not provided
```
mercedesbh, line 84: I think instead of `can be specified only if http_bearer_token is not provided` we can mention that it should `can be specified only if http_basic_user is also provided`

docs/data-sources/vault_cluster.md
```diff
@@ -81,6 +81,16 @@ Read-Only:
...
+- `http_basic_user` (String) HTTP basic authentication username for streaming audit logs, one of the two available authentication methods, can be specified only if http_bearer_token is not provided.
```
mercedesbh, line 85: Same here

... `can be specified only if http_basic_password` is also provided

docs/data-sources/vault_cluster.md
```diff
@@ -110,4 +120,14 @@ Read-Only:
 - `elasticsearch_user` (String) ElasticSearch user for streaming metrics
 - `grafana_endpoint` (String) Grafana endpoint for streaming metrics
 - `grafana_user` (String) Grafana user for streaming metrics
+- `http_basic_password` (String) HTTP basic authentication password for streaming metrics, one of the two available authentication methods, can be specified only if http_bearer_token is not provided
```
mercedesbh, line 123: Same comments as above for `http_basic_password` and `http_basic_user`

docs/resources/vault_cluster.md
```diff
@@ -94,6 +94,16 @@ Optional:
 - `grafana_endpoint` (String) Grafana endpoint for streaming audit logs
 - `grafana_password` (String, Sensitive) Grafana password for streaming audit logs
 - `grafana_user` (String) Grafana user for streaming audit logs
+- `http_basic_password` (String, Sensitive) HTTP basic authentication password for streaming audit logs, one of the two available authentication methods, can be specified only if http_bearer_token is not provided
```
mercedesbh, line 97: Same comments as above for `http_basic_password` and `http_basic_user`

docs/resources/vault_cluster.md
```diff
@@ -133,6 +143,16 @@ Optional:
 - `grafana_endpoint` (String) Grafana endpoint for streaming metrics
 - `grafana_password` (String, Sensitive) Grafana password for streaming metrics
 - `grafana_user` (String) Grafana user for streaming metrics
+- `http_basic_password` (String, Sensitive) HTTP basic authentication password for streaming metrics
```
mercedesbh, line 146: Same comments as above for http_basic_password and http_basic_user

internal/providersdkv2/resource_vault_cluster.go
```diff
@@ -1191,6 +1332,37 @@ func getObservabilityConfig(propertyName string, d *schema.ResourceData) (*vault
 	return getValidObservabilityConfig(config)
 }
 
+// if http observability information is provided, this function ensures that authentication fields are valid and returns the authentication method used
+func validateHTTPAuth(httpBasicUser string, httpBasicPassword string, httpBearerToken string) (diag.Diagnostics, *vaultmodels.HashicorpCloudVault20201125HTTPBearerAuth, *vaultmodels.HashicorpCloudVault20201125HTTPBasicAuth) {
```
bosouza, line 1336: diagnostics is equivalent to the error, the convention is for it to be the last return. We can also remove the repeated `string` as the arguments all have the same type
```suggestion
func validateHTTPAuth(httpBasicUser, httpBasicPassword, httpBearerToken string) (*vaultmodels.HashicorpCloudVault20201125HTTPBearerAuth, *vaultmodels.HashicorpCloudVault20201125HTTPBasicAuth, diag.Diagnostics) {
```

internal/providersdkv2/resource_vault_cluster.go
```diff
@@ -1191,6 +1332,37 @@ func getObservabilityConfig(propertyName string, d *schema.ResourceData) (*vault
...
+	// only one of basic or bearer authentication should be submitted
+	if httpBearerToken != "" && httpBasicUser != "" || httpBearerToken != "" && httpBasicPassword != "" {
+		httpConfigError = diag.Errorf("http configuration is invalid: either the basic or bearer authentication method can be submitted, but not both")
+	} else if httpBasicUser != "" && httpBasicPassword == "" || httpBasicUser == "" && httpBasicPassword != "" {
+		// http basic requires both the username and password to be filled
+		httpConfigError = diag.Errorf("http configuration is invalid: basic authentication requires username and password")
+	}
+
+	if httpConfigError != nil {
+		return httpConfigError, nil, nil
```
bosouza, line 1348: isn't this equivalent to returning in each line where you're setting `httpConfigError`?

internal/providersdkv2/resource_vault_cluster.go
```diff
@@ -1191,6 +1332,37 @@ func getObservabilityConfig(propertyName string, d *schema.ResourceData) (*vault
 	return getValidObservabilityConfig(config)
 }
 
+// if http observability information is provided, this function ensures that authentication fields are valid and returns the authentication method used
+func validateHTTPAuth(httpBasicUser string, httpBasicPassword string, httpBearerToken string) (diag.Diagnostics, *vaultmodels.HashicorpCloudVault20201125HTTPBearerAuth, *vaultmodels.HashicorpCloudVault20201125HTTPBasicAuth) {
+	var httpConfigError diag.Diagnostics
+
+	// only one of basic or bearer authentication should be submitted
+	if httpBearerToken != "" && httpBasicUser != "" || httpBearerToken != "" && httpBasicPassword != "" {
```
bosouza, line 1340: can't this expression be simplified?
```suggestion
	if httpBearerToken != "" && (httpBasicUser != "" || httpBasicPassword != "") {
```

internal/providersdkv2/resource_vault_cluster.go
```diff
@@ -1191,6 +1332,37 @@ func getObservabilityConfig(propertyName string, d *schema.ResourceData) (*vault
...
+	}
+
+	if httpBearerToken != "" {
+		httpBearerAuth := &vaultmodels.HashicorpCloudVault20201125HTTPBearerAuth{
+			Token: httpBearerToken,
+		}
+		return nil, httpBearerAuth, nil
+	}
+
+	httpBasicAuth := &vaultmodels.HashicorpCloudVault20201125HTTPBasicAuth{
```
bosouza, line 1358: aren't we setting httpBasic now even if both User and Password are empty?

jaireddjawed: Would User and Password not be empty here because httpConfigError is nil and httpBearerToken is empty if the code goes down this path?

internal/providersdkv2/resource_vault_cluster.go
bosouza: you need the parenthesis otherwise the && operator takes precedence

internal/providersdkv2/resource_vault_cluster.go
bosouza: I believe this code should handle the case where neither basic nor bearer auth is configured, in which case the HTTP sink should be configured without any authentication settings whatsoever. However with the current implementation you would be setting basic auth with empty user and empty password if those aren't provided by the customer. In other words, leaving `httpBasicAuth==nil` will cause the Vector requests to contain a `Authentication: Basic Ogo=` when the intent was for there to be no authentication headers at all.

Or are you purposefully trying to ensure either Basic or Bearer auth is always configured? That doesn't make much sense to me

jaireddjawed: That was my mistake. Your comment makes sense to me now.

internal/providersdkv2/resource_vault_cluster.go
```diff
@@ -1239,6 +1379,35 @@ func getObservabilityConfig(propertyName string, d *schema.ResourceData) (*vault
 	return getValidObservabilityConfig(config)
 }
 
+// if http observability information is provided, this function ensures that authentication fields are valid and returns the authentication method used
+func validateHTTPAuth(httpBasicUser, httpBasicPassword, httpBearerToken string) (*vaultmodels.HashicorpCloudVault20201125HTTPBearerAuth, *vaultmodels.HashicorpCloudVault20201125HTTPBasicAuth, diag.Diagnostics) {
+	// only one of basic or bearer authentication should be submitted
+	if httpBearerToken != "" && (httpBasicUser != "" || httpBasicPassword != "") {
+		return nil, nil, diag.Errorf("http configuration is invalid: either the basic or bearer authentication method can be submitted, but not both")
+	} else if httpBasicUser != "" && httpBasicPassword == "" || httpBasicUser == "" && httpBasicPassword != "" {
```
bosouza, line 1387: nit: the return statement in the previous if block makes this `else` redundant

internal/providersdkv2/resource_vault_cluster.go
```diff
@@ -1239,6 +1379,35 @@ func getObservabilityConfig(propertyName string, d *schema.ResourceData) (*vault
...
+	}
+
+	if httpBearerToken != "" {
+		httpBearerAuth := &vaultmodels.HashicorpCloudVault20201125HTTPBearerAuth{
+			Token: httpBearerToken,
+		}
+		return httpBearerAuth, nil, nil
+	}
+
+	if httpBasicUser != "" && httpBasicPassword != "" {
```
bosouza, line 1399: nit: I think the `||` operator would make the intent clearer here, or even just checking for User as we'd be relying on  either both being set or both being empty

bosouza (approved):
lgtm

PR #669, opened 2023-11-08
Revert "Add support for http observability on vault_cluster resource"

jaireddjawed:
Reverts hashicorp/terraform-provider-hcp#660 because there is a bug that makes http_uri required when an audit_log_config is provided.


mercedesbh:
@jaireddjawed Looks like a changelog check failure

himran92:
@jaireddjawed @mercedesbh  just for context: have we clarified with terraform team around any concerns with reverts?

mercedesbh:
@himran92 There's some discussion now in[ this thread](https://hashicorp.slack.com/archives/C05BBTQFCGZ/p1697822256416009).

aidan-mundy:
Please remove the changelog from this PR, it is not needed. You can bypass the changelog check using the pr/no-changelog tag.

There should not be any concerns with reverting since this code has not yet released in a provider version.

aidan-mundy (changes requested):
Please remove the changelog from this PR, it is not needed. Also make sure the old changelog is removed by this revert. 

You can bypass the changelog check using the pr/no-changelog tag.

There should not be any concerns with reverting since this code has not yet released in a provider version.

aidan-mundy (approved):

PR #691, opened 2023-12-08
Provider accepts workload identity token via filepath HCPID2-1948

dadgar:

Support authenticating the provider using workload identity token's passed via file. This avoids the need to create a credential file to simplify the UX in certain scenarios.

I have tested the changes manually by having the provider authenticate using the new configuration block and tested that the provider errors if given conflicting ways to authenticate.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -126,7 +126,7 @@ resource "hcp_vault_cluster" "example" {
 - `client_secret` (String) The OAuth2 Client Secret for API operations.
 - `credential_file` (String) The path to an HCP credential file to use to authenticate the provider to HCP. You can alternatively set the HCP_CRED_FILE environment variable to point at a credential file as well. Using a credential file allows you to authenticate the provider as a service principal via client credentials or dynamically based on Workload Identity Federation.
 - `project_id` (String) The default project in which resources should be created.
-- `workload_identity` (Block List) Allows authenticating the provider by exchanging the token specified in the `token_file` for a HCP service principal using Workload Identity Federation. (see [below for nested schema](#nestedblock--workload_identity))
+- `workload_identity` (Block List) Allows authenticating the provider by exchanging the OAuth 2.0 access token or OpenID Connect token specified in the `token_file` for a HCP service principal using Workload Identity Federation. (see [below for nested schema](#nestedblock--workload_identity))
 
 <a id="nestedblock--workload_identity"></a>
 ### Nested Schema for `workload_identity`
@@ -134,7 +134,7 @@ resource "hcp_vault_cluster" "example" {
 Required:
 
 - `resource_name` (String) The resource_name of the Workload Identity Provider to exchange the token with.
-- `token_file` (String) The path to a file containing an identity token.
+- `token_file` (String) The path to a file containing a JWT token retrieved from an OpenID Connect (OIDC) or OAuth2 provider.
 -> **Note:** See the [authentication guide](guides/auth.md) about a use case when specifying `project_id` is needed.
 
 For more information about HCP, please review our [documentation page](https://cloud.hashicorp.com/docs/hcp).
--- a/docs/index.md
+++ b/docs/index.md
@@ -126,6 +126,15 @@ resource "hcp_vault_cluster" "example" {
 - `client_secret` (String) The OAuth2 Client Secret for API operations.
 - `credential_file` (String) The path to an HCP credential file to use to authenticate the provider to HCP. You can alternatively set the HCP_CRED_FILE environment variable to point at a credential file as well. Using a credential file allows you to authenticate the provider as a service principal via client credentials or dynamically based on Workload Identity Federation.
 - `project_id` (String) The default project in which resources should be created.
+- `workload_identity` (Block List) Allows authenticating the provider by exchanging the token specified in the `token_file` for a HCP service principal using Workload Identity Federation. (see [below for nested schema](#nestedblock--workload_identity))
+
+<a id="nestedblock--workload_identity"></a>
+### Nested Schema for `workload_identity`
+
+Required:
+
+- `resource_name` (String) The resource_name of the Workload Identity Provider to exchange the token with.
+- `token_file` (String) The path to a file containing an identity token.
 -> **Note:** See the [authentication guide](guides/auth.md) about a use case when specifying `project_id` is needed.
 
 For more information about HCP, please review our [documentation page](https://cloud.hashicorp.com/docs/hcp).
--- a/internal/clients/client.go
+++ b/internal/clients/client.go
@@ -74,7 +74,7 @@ type ClientConfig struct {
 	CredentialFile string
 
 	// WorkloadIdentityTokenFile and WorkloadIdentityResourceName can be set to
-	// indicate that authentication should occure by using workload identity
+	// indicate that authentication should occur by using workload identity
 	// federation. WorloadIdentityTokenFile indicates a file containing the
 	// token content and WorkloadIdentityResourceName is the workload identity
 	// provider resource name to authenticate against.
--- a/internal/clients/client.go
+++ b/internal/clients/client.go
@@ -11,6 +11,8 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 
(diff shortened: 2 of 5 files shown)


internal/clients/client.go
```diff
@@ -71,6 +73,14 @@ type ClientConfig struct {
 	ClientSecret   string
 	CredentialFile string
 
+	// WorkloadIdentityTokenFile and WorkloadIdentityResourceName can be set to
+	// indicate that authentication should occure by using workload identity
```
jasonpilz, line 77:
```suggestion
	// indicate that authentication should occur by using workload identity
```

jasonpilz (approved):
LGTM

internal/provider/provider.go
```diff
@@ -135,27 +179,34 @@ func (p *ProviderFramework) Configure(ctx context.Context, req provider.Configur
...
+	if len(data.WorkloadIdentity.Elements()) == 1 {
+		elements := make([]WorkloadIdentityFrameworkModel, 0, 1)
+		resp.Diagnostics.Append(data.WorkloadIdentity.ElementsAs(ctx, &elements, false)...)
+		if resp.Diagnostics.HasError() {
+			return
+		}
+
+		clientConfig.WorloadIdentityTokenFile = elements[0].TokenFile.ValueString()
+		clientConfig.WorkloadIdentityResourceName = elements[0].ResourceName.ValueString()
+	}
```
tobias-hashicorp, line 207: Should there be an error if more than one `WorkloadIdentity` element is provided?

internal/providersdkv2/provider.go
```diff
@@ -111,24 +131,33 @@ func configure(p *schema.Provider) func(context.Context, *schema.ResourceData) (
...
+			CredentialFile: d.Get("credential_file").(string),
+			SourceChannel:  p.UserAgent("terraform-provider-hcp", version.ProviderVersion),
+		}
 
-		clientID := d.Get("client_id").(string)
-		if clientID == "" {
-			clientID = os.Getenv("HCP_CLIENT_ID")
+		// Check the environment for an HCP Service Principal
+		if clientConfig.ClientID == "" {
+			clientConfig.ClientID = os.Getenv("HCP_CLIENT_ID")
```
tobias-hashicorp, line 143: 🤔 I know this wasn't introduced with this PR, but this shouldn't be necessary. The SDK is already [configured with FromEnv()](https://github.com/hashicorp/terraform-provider-hcp/pull/691/files#diff-460f9527216e2f272b53bb012adc914caa31669c76c04e4e76b1f8c4c8ed0f2cR98).

internal/providersdkv2/provider.go
```diff
@@ -111,24 +131,33 @@ func configure(p *schema.Provider) func(context.Context, *schema.ResourceData) (
...
-		})
+		// Read the workload_identity configuration
+		if v, ok := d.GetOk("workload_identity"); ok && len(v.([]interface{})) == 1 && v.([]interface{})[0] != nil {
+			wi := v.([]interface{})[0].(map[string]interface{})
+			if tf, ok := wi["token_file"].(string); ok && tf != "" {
+				clientConfig.WorloadIdentityTokenFile = tf
+			}
+			if rn, ok := wi["resource_name"].(string); ok && rn != "" {
+				clientConfig.WorkloadIdentityResourceName = rn
+			}
```
tobias-hashicorp, line 157: I haven't tested it, but I suspect that this could also be done with `mapstructure`, which already is a dependency of the provider. It would be a bit easier to read.

In `client.go`:
```go
WorloadIdentityTokenFile     string `mapstructure:"token_file"`
WorkloadIdentityResourceName string `mapstructure:"resource_name"`
```

Here:
```go
err := mapstructure.Decode(v.([]interface{})[0], clientConfig)
if err != nil {
  diags = append(diags, diag.Errorf("unable to parse workload identity config: %v", err)...)
  return nil, diags
}
```

chrisarcand (commented):
Leaving the code for the team, as they've got more context in this area, but I've got two overarching thoughts as to the interface:

1) This whole area of authentication has a _lot_ of names: Workload Identity, JWT Authentication, OpenID Connect, etc etc etc. In all other major cases (AWS, Azure, Google, etc) the interfaces and documentation are pretty specific about mentioning OIDC and JWTs even if they call the overall area something different (For example, Google also calls this area [Workload Identity Federation](https://cloud.google.com/iam/docs/workload-identity-federation) as we do for HCP).  AWS has ["specifies the path to a file which contains an OAuth 2.0 access token or OpenID Connect ID token that is provided by the identity provider."](https://registry.terraform.io/providers/hashicorp/aws/latest/docs#assume_role_with_web_identity-configuration-block) Azure goes all in on calling the feature OIDC, with [oidc-prefixed configuration arguments](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs#oidc_request_token).  

    👉🏻 **Wherever possible, I strongly recommend adding hints of "OIDC" and JWTs in documentation and names to do the same thing, even if the main feature in HCP is called Workload Identity Federation.** "a token" should be expanded on as an OpenID Connect token or JWT. "Authenticating with the provider" could be "Authenticating via OpenID Connect". `workload_identity` is a fine attribute name, I just think we need more hints in the docs.

    In other words, because this area has _so many_ different terms and it's confusing for users, I think the general answer to it all is referencing more terms so at least one or two is familiar to people and not wholly service-specific (Someone may get the hint that this is OIDC, having never heard of Workload Identity Federation, etc). 

2) "Resource name" didn't really click with me at first. 👉🏻 **Could we more specifically say "HCP service principal", if that's what they are?** (even if the service principal is itself a representation of authentication attached to a particular resource in HCP). A link to the service principals docs wouldn't hurt either (https://developer.hashicorp.com/hcp/docs/hcp/admin/iam/service-principals)

dadgar (commented):
@chrisarcand Thanks for the review. I left comments inline clarifying that the resource_name is not a service principal but rather a child object called the Workload Identity Provider.

I wanted to touch on your comment on being explicit that the token_file is pointing at a JWT or OIDC token. I actually avoided that on purpose since the workload identity allows for more than just JWTs to be exchanged. Currently we support a signed AWS CallerIdentity request to exchange tokens for AWS workloads as well. Furthermore, there are other IDP protocols such as SAML that we may want to support in the future.

However, upon reflection, I will go with your suggestion to make it more explicit since the AWS support requires more configuration that exists in the credential_file authentication scheme and thus that only leaves the JWT access token which we can document more clearly. If we do extend to SAML, we can revisit the documentation.

I will update the provider to point at documentation for workload identity federation once it exists 😄

docs/index.md
```diff
@@ -126,6 +126,15 @@ resource "hcp_vault_cluster" "example" {
...
 - `credential_file` (String) The path to an HCP credential file to use to authenticate the provider to HCP. You can alternatively set the HCP_CRED_FILE environment variable to point at a credential file as well. Using a credential file allows you to authenticate the provider as a service principal via client credentials or dynamically based on Workload Identity Federation.
 - `project_id` (String) The default project in which resources should be created.
+- `workload_identity` (Block List) Allows authenticating the provider by exchanging the token specified in the `token_file` for a HCP service principal using Workload Identity Federation. (see [below for nested schema](#nestedblock--workload_identity))
+
+<a id="nestedblock--workload_identity"></a>
+### Nested Schema for `workload_identity`
+
+Required:
+
+- `resource_name` (String) The resource_name of the Workload Identity Provider to exchange the token with.
```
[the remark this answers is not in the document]
chrisarcand: Oooooh I see, my bad, that makes sense. Indifferent on the name, then - I'm not sure `provider_` adds anything or not; either way, it'll be good to reference docs to this resource when they exist. 👍🏻

chrisarcand (commented):
From my high level perspective this looks great! Thanks for the further clarification in the docstrings, that reads better for sure. ✅

chrisarcand:
Should have added: Tomorrow I'm happy to build the provider and give it a go for our specific use case, to verify. Not sure what the release schedule is for this provider but we should make sure it works before stamping it as final.

dadgar:
@chrisarcand Awesome, will wait on confirmation from you that it worked before merging.

chrisarcand:
@dadgar Good to go!

PR #698, opened 2023-12-12
Add ability to skip e2e tests in prerelease

jasonpilz:

* This PR adds the ability to skip the E2E tests as a dependency to ship a new release. (In the event of break-glass scenarios)
* This can be useful when releases are desired for documentation-only, for example.


chapmanc (approved):
Approving to unblock @jasonpilz

aidan-mundy (changes requested):
Heyo, we had this functionality in the repo CI for a period of time, and it resulted in two straight months where our releases failed CI but were released anyway. It is far too easy to just push the "bypass the requirements" button.

This functionality is not required to ship a manual release without tests. To release manually, you just need to update the changelog and provider docs like the prerelease would (see the job summary of a successful run for an example diff), post a PR with the changes, merge to `main`, and then add the desired version tag to the commit on `main`. When a version tag is added, the release workflow is triggered and the provider will be released. 

This process is far more intentional and only marginally more time consuming. Since we removed the option to skip e2e tests, we have had zero releases with failing tests, and I can only recall one other instance where we got close to performing a manual release.

jasonpilz:
@aidan-mundy Luckily, the Prerelease finally passed. I will close this PR.

sylviamoss (approved):

manish-hashicorp (approved):
LGTM. Although I wonder if we should follow @aidan-mundy steps to manually release it.

jasonpilz:
@manish-hashicorp The manual steps are essentially what is left in this workflow when the E2E is skipped. 👍🏻

PR #709, opened 2023-12-19
Guides - Page name updates

jasonpilz:

* Currently the nav shows a suffix for each Guide name, ` - HCP Provider`. Since these page names are already nested under `hcp provider`, this is redundant and adds distraction.
* Removes those suffixes and regenerate the docs.
Screenshot 2023-12-19 at 12 10 58 PM


manish-hashicorp:
Can you check why #691 is in this PR? I believe `main` was rebased recently and I am not sure if this will cause any issues.

jasonpilz:
@manish-hashicorp Yeah I'm working on that

PR #723, opened 2024-01-15
[HPR-1157] HCP Webhook resource

sylviamoss:

```sh
$ make testacc TESTARGS='-run=TestAccNotificationsWebhookResource -test.v'
TF_ACC=1 go test ./internal/... -v -run=TestAccNotificationsWebhookResource -test.v -timeout 360m -parallel=10
?       github.com/hashicorp/terraform-provider-hcp/internal/clients/iampolicy  [no test files]
PASS
ok      github.com/hashicorp/terraform-provider-hcp/internal/clients    (cached) [no tests to run]
PASS
PASS
ok      github.com/hashicorp/terraform-provider-hcp/internal/hcpvalidator       (cached) [no tests to run]
PASS
?       github.com/hashicorp/terraform-provider-hcp/internal/provider   [no test files]
?       github.com/hashicorp/terraform-provider-hcp/internal/provider/acctest   [no test files]
?       github.com/hashicorp/terraform-provider-hcp/internal/provider/modifiers [no test files]
PASS
ok      github.com/hashicorp/terraform-provider-hcp/internal/provider/iam       (cached) [no tests to run]
PASS
ok      github.com/hashicorp/terraform-provider-hcp/internal/provider/logstreaming      (cached) [no tests to run]
PASS
ok      github.com/hashicorp/terraform-provider-hcp/internal/provider/resourcemanager   (cached) [no tests to run]
PASS
ok      github.com/hashicorp/terraform-provider-hcp/internal/provider/vaultsecrets      (cached) [no tests to run]
=== RUN   TestAccNotificationsWebhookResource
--- PASS: TestAccNotificationsWebhookResource (18.53s)
PASS
ok      github.com/hashicorp/terraform-provider-hcp/internal/provider/webhook   19.311s
PASS
ok      github.com/hashicorp/terraform-provider-hcp/internal/provider/webhook/validator (cached) [no tests to run]
PASS
ok      github.com/hashicorp/terraform-provider-hcp/internal/providersdkv2      (cached) [no tests to run]

```


docs/resources/webhook.md
```diff
@@ -0,0 +1,97 @@
...
+
+- `resource_id` (String) Refers to the resource the webhook is subscribed to. If not set, the webhook subscribes to the emitted events listed in events for any resource in the webhook's project.
+
+<a id="nestedatt--subscriptions--events"></a>
+### Nested Schema for `subscriptions.events`
+
+Required:
+
+- `actions` (List of String) The list of event actions subscribed for the resource type set as the [source](#source). For example, `["create", "update"]`. When the action is '*', it means that the webhook is subscribed to all event actions for the event source.
+- `source` (String) The resource type of the source of the event. For example, `hashicorp.packer.version`. Event source might not be the same type as the resource that the webhook is subscribed to ([resource_id](#resource_id)) if the event is from a descendant resource. For example, webhooks are subscribed to a `hashicorp.packer.registry` and receive events for descendent resources such as a `hashicorp.packer.version`.
```
dadgar, line 88: Is there any documentation we can link the user to that enumerates the options they can set?

examples/resources/hcp_webhook/import.sh
```diff
@@ -0,0 +1,3 @@
+# HCP Webhook can be imported by specifying the webhook resource name
+terraform import hcp_webhook.example webhook/prokect/840e3701-55b6-4f86-8c17-b1fe397303c5/geo/us/webhook/example-webhook
```
dadgar, line 2:
```suggestion
terraform import hcp_webhook.example webhook/project/840e3701-55b6-4f86-8c17-b1fe397303c5/geo/us/webhook/example-webhook
```

examples/resources/hcp_webhook/resource.tf
```diff
@@ -0,0 +1,20 @@
+resource "hcp_webhook" "example" {
```
dadgar, line 1: Should we put a comment that describes what this web hook does?

` # Notify for all new versions of a Packer image being created in the project.`

docs/resources/webhook.md
```diff
@@ -0,0 +1,97 @@
...
+---
+
+# hcp_webhook (Resource)
+
+The webhook resource manages a HCP webhook, used to notify external systems about a project resource's lifecycle events
+
+## Example Usage
+
+```terraform
+resource "hcp_webhook" "example" {
```
dadgar, line 15: What do you think about making the name of the resource slightly more specific. I was thinking something like `hcp_notifications_webhook`. The thought being that in the future we may have inbound web hooks as well to trigger actions. This change would potentially make it more clear what this resource is for and potentially prevent a name collision in the future.

internal/hcpvalidator/url.go
```diff
@@ -0,0 +1,58 @@
...
+			v.Description(ctx),
+			value,
+		))
+	}
+}
+
+// URL returns an AttributeValidator which ensures that any configured
+// attribute value satisfies requirements of a valid URL.
+// Null (unconfigured) and unknown (known after apply) values are skipped.
+func URL() validator.String {
```
dadgar, line 56: Could we have another function called `HTTPSUrl()` that returns a validator that requires HTTPS? A lot of our endpoints will not accept a http:// URL so would be a nice helper to provide

internal/provider/webhook/resource_webhook.go
```diff
@@ -0,0 +1,530 @@
...
+
+			"description": schema.StringAttribute{
+				Optional:    true,
+				Description: "The webhook's description. Descriptions are useful for helping others understand the purpose of the webhook.",
+			},
+
+			"enabled": schema.BoolAttribute{
+				Optional:            true,
+				Computed:            true,
+				MarkdownDescription: "Indicates if the webhook should receive payloads for the subscribed events.",
```
dadgar, line 100: Should receive -> should send

Right?

internal/provider/webhook/resource_webhook.go
```diff
@@ -0,0 +1,530 @@
...
+				// The next apply the new name will no longer be an update.
+				state.Name = plan.Name
+				resp.Diagnostics.Append(resp.State.Set(ctx, &state)...)
+			}
+		}()
+	}
+
+	var updateMaks []string
+	if !plan.Config.URL.Equal(state.Config.URL) ||
+		!plan.Config.HmacKey.Equal(state.Config.HmacKey) {
```
brandonc, line 444: I couldn't get the provider to detect that I had changed my hmac key

brandonc (commented):
Some smoke testing findings. I like the new schema!

examples/resources/hcp_webhook/import.sh
nywilken, line 2: Is it possible to strip the static values from the required input name and have provider automatically generate the full resource name based off the input? 

For example instead of 
```
terraform import hcp_webhook.example webhook/prokect/840e3701-55b6-4f86-8c17-b1fe397303c5/geo/us/webhook/example-webhook
```

The user could write
```
# Using an explicit project ID, the import ID is:
# {project_id}:{webhook_geo_location}:{webhook_name}
terraform import hcp_webhook.example 840e3701-55b6-4f86-8c17-b1fe397303c5:us:example-webhook
```

examples/resources/hcp_webhook/resource.tf
nywilken: Explaining what the webhook example does is a good call out. What about setting the description attribute for this resource instead of a comment?

docs/resources/webhook.md
nywilken: That's a great question. I believe these will be documented per-product on their respective webhook pages - based on https://developer.hashicorp.com/hcp/docs/packer/webhooks

Linking to the HCP Packer page to start would be helpful but I don't know how fast it will become outdated once other products start to offer webhooks. 

Wondering if there should be a central webhooks page for HCP that gets updated as products enable webhooks, which can be linked to later on. What do you think @dadgar?

docs/resources/webhook.md
sylviamoss: That's a good point. I like your suggestion. I am just going to run it with Jordan because he approved the design and just want to make sure he is aware of the change.

docs/resources/webhook.md
sylviamoss: Wilken is right, these will be documented per-product and I think the best solution would be to have a single page where we would link to those documentation. I think that makes total sense. 

I can work on adding a section to https://developer.hashicorp.com/hcp/docs/packer/webhooks, but before, I would like to confirm if that would address the concerns of this thread.

examples/resources/hcp_webhook/import.sh
sylviamoss: I like the idea, but I wonder which one could give more work to write. The API provides the resource_name, but `{project_id}:{webhook_geo_location}:{webhook_name}` will always have to be built by the user. 

Would it be helpful if, next to the `import` documentation, the `curl` to get the resource_name with the API is provided?

examples/resources/hcp_webhook/resource.tf
sylviamoss: I think I will add both the comment, to describe the current state of this resource, and a better generic description, because tbh "My new webhook!" is not realistic at all.

internal/hcpvalidator/url.go
sylviamoss: Sure! I can add that as well.

internal/provider/webhook/resource_webhook.go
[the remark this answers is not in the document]
sylviamoss: That's interesting and I think that other resources in this provider are also not sensitive to provider-level project id change. I did looked how they handle project_id and didn't see anything different from what I implemented here. Thanks for catching this one. I will have a look at the plan modifier you created.

internal/provider/webhook/resource_webhook.go
sylviamoss: My line of thought for `receive`  is that the product services will emit events that HCP will send the payload via a request made to the webhook, configured with this resource, that will then receive those payloads and act on them. I have been writing the docs (API, developer.hashicorp.com) from this perspective. 

The wording I got from https://registry.terraform.io/providers/integrations/github/latest/docs/resources/repository_webhook#active which is also

internal/provider/webhook/resource_webhook.go
sylviamoss: I did test that but maybe I made a change in between that changed the behavior. I will have a look and an acceptance test for this too.

examples/resources/hcp_webhook/resource.tf
sylviamoss: I will actually only add it to `description` because it became repetitive information when I wrote in both places.

examples/resources/hcp_webhook/import.sh
nywilken: I'm not sure I understand what you mean by "I wonder which one could give more work to write". Are you referring to the effort of the user? 

Generally speaking the input to an import should be something relatively simple for the user to grok from the system under management so the user can quickly construct the required input without prior knowledge about the underlying API for the resources they wish to manage. 

Looking at the channel import it just takes the bucket name and channel name, all of which is accessible via the UI. 

> Would it be helpful if, next to the import documentation, the curl to get the resource_name with the API is provided?

Are you thinking about users who will query the API over users who just want to use Terraform to manage existing webhooks? 

I think for an advance user who is leveraging the API this information would be helpful especially if they want to do some sort of bulk import. But for the average user I don't think that they will access the API for this information. 

As for constructing from the API, by looking at [webhook events page](https://developer.hashicorp.com/hcp/docs/packer/webhooks), it looks like we have an attribute for project_id, resource_name, but not geo location or the simple webhook name. So constructing `{project_id}:{webhook_geo_location}:{webhook_name}` from the API programmatically would take some extra effort. 

I might be over thinking TBH but my first thought was that there are static values in the input that we can probably do away with for the user since that info may not be directly available in the UI. But also use of webhooks is 🧪 so I can't recall what is shown in the UI.

internal/hcpvalidator/resource_type.go
```diff
@@ -0,0 +1,72 @@
...
+
+const (
+	validResourceTypeRegex = `^hashicorp(\.([a-z][a-z-]*)){2}$`
+	invalidResourceTypeErr = "a Resource Type must have format hashicorp.<service>.<resource> and consist only of lowercase alphabetic characters, dashes, and dots"
+)
+
+var (
+	resourceTypeRegex                  = regexp.MustCompile(validResourceTypeRegex)
+	_                 validator.String = resourceTypeValidator{}
+)
```
nywilken, line 22: Looking at the validators in this PR I'm guessing this is convention used in the provider for checking that the type implements the ValidateString interface, correct? 

I found it a bit odd seeing as the type is actually being defined below. It feels like a check that could be added as as simple unit test to check if it implements the interface, especially since the compiler will push back with a type check error if  `resourceTypeValidator.ValidateString` is being passed/called and it no longer implements the interface. 

If this must be within the validator code for convention my nit would be to use two separate var definitions to remove the added whitespace from fmt. 



```suggestion
var _ validator.String = resourceTypeValidator{}
var resourceTypeRegex = regexp.MustCompile(validResourceTypeRegex)
```

internal/provider/webhook/resource_webhook.go
sylviamoss: I tested it again and I couldn't reproduce what you experienced. I added two new acceptance tests to test that it is possible to update from none to an hmac_key and from one hmac_key to another.  Can you try one more time?

examples/resources/hcp_webhook/import.sh
sylviamoss: I am using `resource_name` because it is a real identifier of the resource, and I am trying to keep the resource aligned with the API as possible. I think `{project_id}:{webhook_geo_location}:{webhook_name}` is a new way to identify the webhook, which differs from the identifier pattern that we have been adopting in the platform and that the webhook service is designed on top. 

https://registry.terraform.io/providers/hashicorp/hcp/latest/docs/resources/service_principal and https://registry.terraform.io/providers/hashicorp/hcp/latest/docs/resources/iam_workload_identity_provider#import are examples of resources that similar to this one it is also using the new identifier and also uses the resource_name to import. 

I agree that there is not enough information on the UI to write the resource_name, which is only because this service adopted the multi-region capability of including geography. I prefer to mitigate that by guiding how to get the resource_name of a webhook instead of working around making this identifier different from the pattern that we are trying to follow.  So I am leaning toward keeping the resource_name and improving the documentation around because of the `geo` part.

internal/provider/webhook/resource_webhook.go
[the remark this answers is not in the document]
sylviamoss: @dadgar I want to catch your attention to @brandonc's comment. I couldn't find any resource that is taking care of this, which means none of them are considering the provider's default changes and cascading updates to resources.

docs/resources/webhook.md
dadgar: What about having a generic web hooks documentation page that describes the capabilities and have links to each product page that integrates with web hooks for more details on events. That way it is scalable.

Something like: 
```
# Webhooks
How they work

## Setting them up examples
...

## Webhook Event Documentation

Packer Webhooks -> Link to existing page
HVS Webhooks -> Link to HVS
...
```

internal/provider/webhook/resource_webhook.go
dadgar: Sounds good, I guess what matters is we are consistent.

internal/hcpvalidator/url.go
dadgar: Does the web hook service allow HTTP endpoints? Should we restrict to HTTPS?

internal/provider/webhook/resource_webhook.go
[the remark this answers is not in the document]
dadgar: @brandonc Good catch, could you file an issue so we eventually fix that. @jasonpilz may be interested in that

docs/resources/webhook.md
nywilken: This is a good option. Especially, if teams already update provider docs as they add platform support or other updates to resources this option would be ideal. 

 I considered it previously for a moment but thought that given the distance between the provider and the product docs themselves that teams not needing to change the provider could forget or not know now to update the docs. In my head the closer it is the the platform docs the more visibility it would get. 

Thinking out loud, users will probably find their way to the docs for their product of choice after clinking into anyone of the web events pages and searching. So maybe the docs becoming stale is not a big issue. A generic web hooks documentation page that describes the capabilities and links to product pages seems like a good tradeoff at the moment.

internal/provider/webhook/resource_webhook.go
```diff
@@ -0,0 +1,530 @@
...
+		newSubscription := webhookSubscription{
+			Events: make([]webhookSubscriptionEvent, 0),
+		}
+
+		if subscription.ResourceID != "" {
+			newSubscription.ResourceID = types.StringValue(subscription.ResourceID)
+		}
+
+		eventsMap := make(map[types.String][]types.String)
+		for _, event := range subscription.Events {
```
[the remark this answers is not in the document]
sylviamoss: Hey @brandonc it took me a bit of work but I was able to implement a quite complex validator. Let me know what you think.

docs/resources/webhook.md
sylviamoss: Would this generic page be a guide?

dadgar: I think it would make more sense to add here so it can be reference for TF, CLI, API. 

https://github.com/hashicorp/hcp-docs

sylviamoss: Got it. I will add a new section to https://developer.hashicorp.com/hcp/docs/hcp/admin/projects/webhooks 
I think it also makes sense to add a section about the terraform resource.

internal/provider/webhook/resource_notifications_webhook.go
```diff
@@ -0,0 +1,545 @@
...
+						MarkdownDescription: `The HTTP or HTTPS destination URL that HCP delivers the event payloads to. 
+The destination must be able to use the HCP webhook 
+[payload](https://developer.hashicorp.com/hcp/docs/hcp/admin/projects/webhooks#webhook-payload).`,
+						Validators: []validator.String{
+							hcpvalidator.URL(),
+						},
+					},
+					"hmac_key": schema.StringAttribute{
+						Optional: true,
+						Description: "The arbitrary secret that HCP uses to sign all its webhook requests. This is a" +
```
brandonc, line 78:
```suggestion
						Description: "The arbitrary secret that HCP uses to sign all its webhook requests. This is a " +
```

internal/provider/webhook/resource_notifications_webhook.go
```diff
@@ -0,0 +1,545 @@
...
+							"write-only field, it is written once and not visible thereafter.",
+						Sensitive: true,
+					},
+				},
+			},
+
+			// Optional fields
+			"project_id": schema.StringAttribute{
+				Description: "The ID of the project to create the webhook under. " +
+					"If unspecified, the webhook will be created in the project the provider is configured with. ",
```
brandonc, line 88:
```suggestion
					"If unspecified, the webhook will be created in the project the provider is configured with.",
```

brandonc (commented):
Looks really good. For some reason I combed for whitespace issues in the string concatenations but everything smoke tested... except I still can't modify the hmac field.

To repro, I applied a simple config like this:

```
resource "hcp_notifications_webhook" "foo" {
  name        = "example-webhook"
  description = "My new webhook!"
  enabled     = false
  config = {
    url = "https://en46gax89xugj.x.pipedream.net"
    hmac = "example"
  }
  subscriptions = [
    {
      events = [
        {
          actions = ["create"]
          source  = "hashicorp.packer.iteration"
        },
      ]
    }
  ]
}
```

and then tried changing "example" to something else, like "example2". The second apply did not detect changes between the state and the config

internal/provider/webhook/resource_webhook.go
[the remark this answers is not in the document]
brandonc: Works great! Regarding the complexity of the validator, I noticed you included a lot of type assertion error checks that would probably already be covered by the basic schema validator. For instance, this code checks to ensure events is indeed a list:

```go
eventsValuable, ok := events.(basetypes.ListValue)
if !ok {
resp.Diagnostics.AddAttributeError(
	req.Path,
	"Invalid subscription events",
	"While performing schema-based validation, an unexpected error occurred. "+
		"The attribute declares a List value validator, however its values do not implement types.ListType interface for custom List types. "+
		"This is likely an issue with terraform-plugin-framework and should be reported to the provider developers.\n\n"+
		fmt.Sprintf("Path: %s\n", req.Path.String())+
		fmt.Sprintf("Element Type: %T\n", EventsType)+
		fmt.Sprintf("Element Value Type: %T\n", events),
)
return
}
```

but if I change events to be not a list, I get the following error from `terraform validate` or any other command that loads config:

```
│ Error: Incorrect attribute value type
│
│   on main.tf line 15, in resource "hcp_notifications_webhook" "foo":
│   15:   subscriptions = [
│   16:     {
│   17:       events = ""
│   18:     }
│   19:   ]
│
│ Inappropriate value for attribute "subscriptions": element 0: attribute "events": list of object required.
```

That being said, this code doesn't hurt your validator and does seem to protect against framework bugs, as described in the error message

sylviamoss:
@brandonc 
I have just noticed that you configured a `hmac` instead of `hmac_key`. Should the provider fail if you configure an attribute that is not present in the schema? 🤔 I would expect yes and would be detected by the framework, but apparently, the resource should implement some source of custom type for that. Is that correct?

internal/provider/webhook/resource_webhook.go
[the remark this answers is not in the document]
sylviamoss: Thanks for having a lot. It seems that I overcomplicated with those checks, but as it doesn't hurt I will leave it. If it is bothering the readability too much I will remove it whenever that happens.

internal/hcpvalidator/url.go
```diff
@@ -0,0 +1,58 @@
...
+const (
+	invalidURLErr = "must be a valid URL"
+)
+
+var (
+	_ validator.String = urlValidator{}
+)
+
+// urlValidator validates that a string Attribute's value is a valid URL.
+type urlValidator struct {
```
dadgar, line 23: Seems like you could store a boolean here `httpsOnly bool` and then use the same validator for both url and HTTPSUrl.

With a separate constructor:

```
func HTTPSURL() validator.String {
   return urlValidator{
      httpsOnly: true,
    }
 }
```

docs/resources/notifications_webhook.md
```diff
@@ -0,0 +1,98 @@
...
+
+### Required
+
+- `config` (Attributes) The webhook configuration used to deliver event payloads. (see [below for nested schema](#nestedatt--config))
+- `name` (String) The webhook's name.
+
+### Optional
+
+- `description` (String) The webhook's description. Descriptions are useful for helping others understand the purpose of the webhook.
+- `enabled` (Boolean) Indicates if the webhook should receive payloads for the subscribed events.
```
dadgar, line 48: Should this say it defaults to true

docs/resources/notifications_webhook.md
```diff
@@ -0,0 +1,98 @@
...
+
+The webhook resource manages a HCP webhook, used to notify external systems about a project resource's lifecycle events
+
+## Example Usage
+
+```terraform
+resource "hcp_notifications_webhook" "example" {
+  name        = "example-webhook"
+  description = "Notify for all of the events for all Packer artifact versions existing in the project."
+  enabled     = true
```
dadgar, line 18: This is the default, remove?

dadgar (approved):

nywilken (approved):
I have an open question about adding a validator for catching invalid event actions. But this otherwise looks good. Approving in advance in case the validator was intentionally skipped.

delores-hashicorp (approved):

internal/provider/webhook/resource_notifications_webhook.go
```diff
@@ -0,0 +1,545 @@
...
+								Attributes: map[string]schema.Attribute{
+									"actions": schema.ListAttribute{
+										ElementType: types.StringType,
+										Required:    true,
+										Description: "The list of event actions subscribed for the resource type set as the [source](#source). " +
+											"For example, `[\"create\", \"update\"]`. " +
+											"When the action is '*', it means that the webhook is subscribed to all event actions for the event source. ",
+										Validators: []validator.List{
+											listvalidator.UniqueValues(),
+										},
```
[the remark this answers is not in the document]
sylviamoss: We can't add a validation because these values are not controlled by the webhook service but by the product service. If we add a validation we limit the resource to existing integrations. New events created by product services that have individual action, like packer has `revoke`, won't be compatible with the resource until a new version is released with the change.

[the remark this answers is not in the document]
nywilken: Ah fair point. You can mark this comment as resolved then.

PR #743, opened 2024-01-30
Revert "Update Prerelease run condition"

sylviamoss:
We need this change to release.
Reverts hashicorp/terraform-provider-hcp#733

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -28,6 +28,7 @@ jobs:
 
   prerelease:
     name: Prerelease
+    if: always()
     needs: [run-tests]
     runs-on: ubuntu-latest
     steps:


nywilken (approved):

anushkatayal (approved):

manish-hashicorp:
@sylviamoss Out of curiosity, was this required when there is option to [skip test](https://github.com/hashicorp/terraform-provider-hcp/blob/f10318cac4eb25f113a33b35f47a7caf37cb84a8/.github/workflows/prerelease.yml#L23)? Did that not help?

sylviamoss:
@manish-hashicorp Yes, because the prerelease job contains a ` needs: [run-tests]` so if tests don't run then it will not run, but with `always()` it will force it to run anyways, then it will always run even when we skip tests. So we need to change the logic for what we are trying to achieve with the option of skipping the tests. 
This was just a quick solution because I needed to release it yesterday.

manish-hashicorp:
Thanks for the info @sylviamoss.

PR #749, opened 2024-02-06
Update example for hcp_packer_channel_assignment

danbarr:

Docs-only change:
Updates the `hcp_packer_channel_assignment` resource example to the current syntax (removing/replacing deprecated attributes).


nywilken (approved):
Good Catch!!!

manish-hashicorp (approved):

PR #776, opened 2024-03-05
Hcp waypoint tfc config update

catsby:
Minor refactoring of the Waypoint TFC Config resource to fix a bug with updating. There is duplicated code but it enables updates and added a regression test
```sh
$ ➜ make testacc TESTARGS="-run=TestAccWaypointTfcConfig_basic"

...
```

### DIFF
--- a/internal/provider/waypoint/resource_waypoint_tfc_config.go
+++ b/internal/provider/waypoint/resource_waypoint_tfc_config.go
@@ -103,8 +103,18 @@ func (r *TfcConfigResource) Create(ctx context.Context, req resource.CreateReque
 		return
 	}
 
-	var err error
-	plan, err = r.upsert(ctx, plan)
+	projectID := r.client.Config.ProjectID
+	if !plan.ProjectID.IsUnknown() {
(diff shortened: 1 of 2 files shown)


internal/provider/waypoint/resource_waypoint_tfc_config.go
```diff
@@ -113,6 +144,16 @@ func (r *TfcConfigResource) Create(ctx context.Context, req resource.CreateReque
 		return
 	}
 
+	// Generate the unique ID for the resource
+	uID := fmt.Sprintf("/project/%s/%s/%s",
```
HenryEstberg, line 148: This could probably be a helper function in addition to the `loc` and namespace stuff like you had mentioned.

catsby: Done, and simplified to not include the tfc org name in https://github.com/hashicorp/terraform-provider-hcp/pull/776/commits/81e37c2e5dea366ee3c56838fe8142e17e43ca8f

HenryEstberg (approved):
LGTM

PR #778, opened 2024-03-06
New Resource: Waypoint Application Template

catsby:
Adds new resource `hcp_waypoint_application_template`. 
Example config:

```
resource "hcp_waypoint_application_template" "template" {
  name = "some test app template"
  summary = "some summary for fun"
  terraform_no_code_module = {
    source  = "some source"
    version = "some version"
  }
  terraform_cloud_workspace_details = {
    name                 = "some name"
    terraform_project_id = "some id"
  }
}
```

(Note that this is a draft while I figure out a test failure)

Tests:
```sh
$ make testacc TESTARGS='-run=TestAccWaypoint_Application_Template_basic'

...
```


internal/provider/provider.go
```diff
@@ -164,6 +165,8 @@ func (p *ProviderFramework) DataSources(ctx context.Context) []func() datasource
 		vaultsecrets.NewVaultSecretsSecretDataSource,
 		// IAM
 		iam.NewServicePrincipalDataSource,
+		// // Waypoint
+		// waypoint.NewApplicationTemplateDataSource,
```
catsby, line 169: I forgot to reintroduce this, I'll update the PR later after the test issue is resolved

internal/provider/waypoint/resource_waypoint_application_template.go
```diff
@@ -0,0 +1,497 @@
...
+// ApplicationTemplateResourceModel describes the resource data model.
+type ApplicationTemplateResourceModel struct {
+	ID        types.String `tfsdk:"id"`
+	Name      types.String `tfsdk:"name"`
+	ProjectID types.String `tfsdk:"project_id"`
+	OrgID     types.String `tfsdk:"organization_id"`
+	Summary   types.String `tfsdk:"summary"`
+	Labels    types.List   `tfsdk:"labels"`
+	// Description            types.String `tfsdk:"description"`
+	// ReadmeMarkdownTemplate types.String `tfsdk:"readme_markdown_template"`
```
catsby, line 46: Also forgot to support these; will update PR

internal/provider/waypoint/data_source_waypoint_application_template.go
```diff
@@ -0,0 +1,227 @@
...
+				Computed:    true,
+				Optional:    true,
+			},
+			"organization_id": schema.StringAttribute{
+				Description: "The ID of the HCP organization where the Waypoint Application Template is located.",
+				Computed:    true,
+			},
+			"project_id": schema.StringAttribute{
+				Description: "The ID of the HCP project where the Waypoint Application Template is located.",
+				Optional:    true,
```
briancain, line 78: I feel like this might be required? I'm not sure though, but what happens when you try to use an application template resource without an HCP project ID?

HenryEstberg: I think the validator will reject the application template and return an error if there is no project ID

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/hashicorp/terraform-provider-hcp
 
-go 1.21
+go 1.22
+
+toolchain go1.22.1
```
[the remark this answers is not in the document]
catsby: I'm not sure how that happened. I reverted it in 4ac1dc0 as I don't believe it was intentional or necessary. Also TIL about the `toolchain` directive in `go.mod` files:

- https://go.dev/doc/toolchain

internal/provider/waypoint/data_source_waypoint_application_template.go
catsby: What @HenryEstberg said; optional here means the user can optionally provide the `project_id` if they want to create this template for a different project than the one their provider is initially configured to work with. A default is obtained from the client ID/Secret pair. In the future, RBAC may prevent this from working in some cases, but for now it's recommended by the HCP provider team(?) to allow for an optional `project_id`

internal/provider/waypoint/data_source_waypoint_application_template.go
```diff
@@ -0,0 +1,227 @@
...
+				Computed:    true,
+			},
+			"summary": schema.StringAttribute{
+				Description: "A brief description of the template, up to 110 characters",
+				Computed:    true,
```
HenryEstberg, line 83: Should this be required instead of computed? It is required by Waypoint in the validator.

HenryEstberg (approved):
I left one comment about summary being computed vs required that you should look at before merging but everything else looks good.

internal/provider/waypoint/data_source_waypoint_application_template.go
HenryEstberg: For anyone else reading, `summary` should actually be computed, but so should be everything else here except for `name`, `project_id`, and `id`. So a small change was required but not the one indicated in the comment above.

briancain (approved):
🚀  Let's goooo

catsby:
@hashicorp/cloud-foundations please take a look when you can, the Waypoint team has approved this

dmalch-hashicorp (approved):

PR #784, opened 2024-03-13
Bump hcp-sdk-go to 0.87.0

catsby:
Follow up to https://github.com/hashicorp/terraform-provider-hcp/pull/767 where the sdk goes from 0.85->0.86, here we bump to the latest of 0.87.0 for hcp-sdk-go. This will unblock the waypoint team.


catsby:
@dmalch-hashicorp thank you again!

Issue #677, opened 2023-11-24
Circular Dependency project_id


jamesgeddes:
### Terraform Version and Provider Version

```
Terraform version:  1.6.0
HCP provider version: 0.77.0
```

### Affected Resource(s)

* hcp_project
* hcp_vault_secrets_app

### Debug Output

The provider is working as currently intended.

### Expected Behavior
When using the HashiCorp Cloud Platform (HCP) Terraform provider, it is anticipated that one should be able to dynamically create an HCP project using the hcp_project resource and then use the output of this resource (specifically the project_id) within the same Terraform configuration to configure other HCP resources, like hcp_vault_secrets_app. This is based on the expectation that Terraform should handle the flow of data between resources and allow the use of outputs from one resource as inputs to others, or even as part of the provider configuration, within the same execution plan.

### Actual Behavior
In practice, when attempting to use the hcp_project resource to create an HCP project and then reference its project_id in subsequent resources or as part of the provider configuration, a circular dependency error is encountered. The HCP Terraform provider requires the project_id to be set at the provider level. However, this project_id is only known after the hcp_project resource is created, leading to a situation where the provider configuration is dependent on a resource that, in turn, depends on the provider. This results in Terraform being unable to resolve the order of creation and configuration, causing a cycle error.

### Important ~~[Factoids](https://www.dictionary.com/browse/factoid)~~ Facts

- Provider Design: The design of the HCP provider necessitates that the project_id be known and set at the provider level, which is not feasible when the project ID is supposed to be dynamically generated within the same Terraform run.
- Use Case Scenario: This issue particularly impacts scenarios where infrastructure management is intended to be dynamic and automated, without manual intervention or a two-phase approach.
- Workarounds Limitations: Potential workarounds, like splitting the Terraform configuration into two separate phases or predefining the project ID, may not align with all use cases, especially those seeking full automation and integration within a single Terraform execution.
- Impact on Infrastructure as Code Practices: This limitation can impact common Infrastructure as Code practices, where dynamic resource provisioning and referencing are key components of automation.

 

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

jamesgeddes:
[...]

PR #793, opened 2024-03-20
Handle multiple organizations for credentials

dmalch-hashicorp:

Updated the provider's logic to handle zero or multiple organizations associated with configured credentials. Added diagnostics to inform the user when no organizations or multiple organizations are found. Now, it prompts an error, requiring users to specify a particular organization in the HCP provider config block.

```sh

...
```

### DIFF
--- a/internal/provider/project_helpers.go
+++ b/internal/provider/project_helpers.go
@@ -12,6 +12,7 @@ import (
 	"github.com/hashicorp/hcp-sdk-go/clients/cloud-resource-manager/stable/2019-12-10/client/project_service"
 	"github.com/hashicorp/hcp-sdk-go/clients/cloud-resource-manager/stable/2019-12-10/models"
 	diagnostic "github.com/hashicorp/terraform-plugin-framework/diag"
+
 	"github.com/hashicorp/terraform-provider-hcp/internal/clients"
 )
 
@@ -30,11 +31,15 @@ func getProjectFromCredentialsFramework(ctx context.Context, client *clients.Cli
 		return nil, diags
 	}
 	orgLen := len(listOrgResp.Payload.Organizations)
-	if orgLen != 1 {
-		diags.AddError(fmt.Sprintf("unexpected number of organizations: expected 1, actual: %v", orgLen), "")
+	if orgLen == 0 {
+		diags.AddError("The configured credentials do not have access to any organization.", "Please assign at least one organization to the configured credentials to use this provider.")
 		return nil, diags
 	}
 	orgID := listOrgResp.Payload.Organizations[0].ID
+	if orgLen > 1 {
+		diags.AddWarning("There is more than one organization associated with the configured credentials.", "The oldest organization is selected as the default. To switch to a specific project within an organization, configure that in the HCP provider config block.")
+		orgID = getOldestOrganization(listOrgResp.Payload.Organizations).ID
+	}
 
 	// Get the project using the organization ID.
 	listProjParams := project_service.NewProjectServiceListParams()
@@ -56,6 +61,21 @@ func getProjectFromCredentialsFramework(ctx context.Context, client *clients.Cli
 	}
 	project = listProjResp.Payload.Projects[0]
 	return project, diags
+}
+
+// getOldestOrganization retrieves the oldest organization from a list based on its created_at time.
+func getOldestOrganization(organizations []*models.HashicorpCloudResourcemanagerOrganization) (oldestOrg *models.HashicorpCloudResourcemanagerOrganization) {
+	oldestTime := time.Now()
+
+	for _, org := range organizations {
+		orgTime := time.Time(org.CreatedAt)
+		if orgTime.Before(oldestTime) {
+			oldestOrg = org
+			oldestTime = orgTime
+		}
+	}
+
+	return oldestOrg
 }
 
 // getOldestProject retrieves the oldest project from a list based on its created_at time.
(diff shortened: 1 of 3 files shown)


internal/provider/project_helpers.go
```diff
@@ -30,11 +31,15 @@ func getProjectFromCredentialsFramework(ctx context.Context, client *clients.Cli
...
-	if orgLen != 1 {
-		diags.AddError(fmt.Sprintf("unexpected number of organizations: expected 1, actual: %v", orgLen), "")
+	if orgLen == 0 {
+		diags.AddError("The configured credentials do not have access to any organization.", "Please assign at least one organization to the configured credentials to use this provider.")
 		return nil, diags
 	}
 	orgID := listOrgResp.Payload.Organizations[0].ID
+	if orgLen > 1 {
+		diags.AddWarning("There is more than one organization associated with the configured credentials.", "The oldest organization is selected as the default. To switch to a specific project within an organization, configure that in the HCP provider config block.")
+		orgID = getOldestOrganization(listOrgResp.Payload.Organizations).ID
```
manish-hashicorp, line 41: I wonder if we should fail if creds can access multiple organizations instead of choosing the oldest one. There is no reason for us to believe that user intends to use the old one. Since we don't provide org id or name anywhere in tf plan I am worried that user may ignore this warning and end up creating resources in org that they didn't intend to. The oldest project was chosen when multi-project was added because it was likely the project user had been using but even there I feel it should've actually failed instead of guessing the project.

@binarymatt Would you still consider this a bug if the error message was clear - as in asking the user to select a project since provider cannot decipher that from credential?

manish-hashicorp (commented):
Let me know your thoughts on this.

internal/provider/project_helpers.go
dmalch-hashicorp: I agree it's important to avoid users unintentionally creating resources in the wrong organization. Failing credentials that access multiple organizations is a good option to consider. 

However, there are also some potential downsides to consider, failing credentials could disrupt the user's workflow, requiring them to take additional steps to specify the organization.

Let's weigh the pros and cons of failing credentials vs. alternative solutions.

binarymatt:
@manish-hashicorp I think an error message would be a good approach.

internal/provider/project_helpers.go
manish-hashicorp: > failing credentials could disrupt the user's workflow, requiring them to take additional steps to specify the organization

That is in fact the intention. The user has to clarify which organization the provider should operate on. Unfortunately, user cannot configure organization in the provider block which I think it should since there are now resources that are getting created in the org - like projects and groups.

dmalch-hashicorp: The code was modified to not default to the oldest organization when multiple organizations are associated with the configured credentials. Now, it instead prompts an error, requiring users to specify a particular organization in the HCP provider config block.

internal/providersdkv2/provider.go
```diff
@@ -210,10 +211,23 @@ func getProjectFromCredentials(ctx context.Context, client *clients.Client) (pro
 		return nil, diags
 	}
 	orgLen := len(listOrgResp.Payload.Organizations)
-	if orgLen != 1 {
-		diags = append(diags, diag.Errorf("unexpected number of organizations: expected 1, actual: %v", orgLen)...)
+	if orgLen == 0 {
+		diags = append(diags, diag.Diagnostic{
+			Severity: diag.Error,
+			Summary:  "The configured credentials do not have access to any organization.",
+			Detail:   "Please assign at least one organization to the configured credentials to use this provider.",
```
manish-hashicorp, line 218: Unfortunately, there is no place to configure the organization id. https://github.com/hashicorp/terraform-provider-hcp/blob/44d4356be6c4daf126b975ab2246b2990b85b8a2/internal/providersdkv2/provider.go#L62-L115 It only takes project id. We will need to update it to take org id and run through different workflows (with user creds, sp creds with diff level access) to see it works. This should fix #677.

dmalch-hashicorp: I see, I was confused because there was a field in the config that made me think that it was possible https://github.com/hashicorp/terraform-provider-hcp/blob/62e7c8949bc35095b0d99af8f0dd72fdaa7f2e3c/internal/clients/client.go#L101

internal/providersdkv2/provider.go
```diff
@@ -210,10 +211,23 @@ func getProjectFromCredentials(ctx context.Context, client *clients.Client) (pro
...
+		})
+		return nil, diags
+	}
+	if orgLen > 1 {
+		diags = append(diags, diag.Diagnostic{
+			Severity: diag.Error,
+			Summary:  "There is more than one organization associated with the configured credentials.",
+			Detail:   "Please configure a specific project in the HCP provider config block",
```
delores-hashicorp, line 226:
```suggestion
			Detail:   "Please configure a specific organization in the HCP provider config block",
```

internal/providersdkv2/provider.go
dmalch-hashicorp: @manish-hashicorp Do you have any recommendations on how to rephrase this error?

manish-hashicorp: Not really until we add support for configuring organization which would be the real fix. Looks like that config is set to be the parent of configured or fetched project id.

binarymatt (approved):

delores-hashicorp (approved):

manish-hashicorp (approved):
Thanks for changing it to project.

PR #813, opened 2024-04-15
HCPF-1554

delores-hashicorp:

```sh

...
```


examples/resources/hcp_vault_secrets_app_iam_policy/resource.tf
```diff
@@ -0,0 +1,23 @@
...
+}
+
+
+resource "hcp_vault_secrets_app" "example" {
+  app_name    = "example-app-name"
+  description = "My new app!"
+}
+
+resource "hcp_vault_secrets_app_iam_policy" "example" {
+  resource_name = "secrets/project/41d107a7-eea6-4b5e-8481-508ab29e2b07/app/example-app-name"
```
manish-hashicorp, line 21: Since the resource is created above can we refer that here?
```terraform
resource "hcp_vault_secrets_app_iam_policy" "example" {
  resource_id = hcp_vault_secrets_app.example.resource_id
  policy_data = data.hcp_iam_policy.example.policy_data
}
```

examples/resources/hcp_vault_secrets_app_iam_binding/resource.tf
```diff
@@ -0,0 +1,14 @@
...
+  name = "example-sp"
+}
+
+resource "hcp_vault_secrets_app" "example" {
+  app_name    = "example-app-name"
+  description = "My new app!"
+}
+
+resource "hcp_vault_secrets_app_iam_binding" "example" {
+  resource_name = "secrets/project/41d107a7-eea6-4b5e-8481-508ab29e2b07/app/example-app-name"
```
manish-hashicorp, line 11: Similarly here too consider taking the resource id or name from the example app above.

internal/provider/vaultsecrets/resource_vault_secrets_app_iam_policy.go
```diff
@@ -0,0 +1,120 @@
...
+	return schema.Schema{
+		MarkdownDescription: d,
+		Attributes: map[string]schema.Attribute{
+			"resource_name": schema.StringAttribute{
+				Required:    true,
+				Description: "The app's resource name in the format secrets/project/<project ID>/app/<app ID>.",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
```
manish-hashicorp, line 41: Are we not taking `resource_id`? Hmm I was thinking we can take both and let users choose one of them. Using `resource_id` will be good because it is provided in all the other resources and globally unique. My [earlier comment](https://hashicorp.slack.com/archives/C04RWBAC30W/p1712960510269219?thread_ts=1712951592.116149&cid=C04RWBAC30W) was only about ensuring it works with `resource_name`. It did not imply that `resource_id` should not be there.

manish-hashicorp (commented):
Thanks. However, can you add `resource_id` attr?

pmmukh (commented):
no real feedback on the core logic, but the examples and tests all make sense to me!

examples/resources/hcp_vault_secrets_app_iam_binding/resource.tf
pmmukh: is this blocked on us on the HVS side adding resource id/name as responses in our APIs and tf resource outputs?

docs/resources/vault_secrets_app_iam_binding.md
```diff
@@ -0,0 +1,19 @@
...
+subcategory: "Cloud Platform"
+description: |-
+{{ .Description | plainmarkdown | trimspace | prefixlines "  " }}
+---
+
+# {{.Name}} ({{.Type}})
+
+{{ .Description | trimspace }}
+
+~> **Note:** `hcp_vault_secrets_app_iam_binding` can only be used if the app has an existing policy and cannot be used in conjunction with
```
dadgar, line 12: This seems like a bug? From a user perspective it should just work to add the first binding.

internal/provider/vaultsecrets/resource_vault_secrets_app_iam_policy.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"github.com/hashicorp/terraform-provider-hcp/internal/clients"
+	"github.com/hashicorp/terraform-provider-hcp/internal/clients/iampolicy"
+)
+
+// vaultSecretsAppIAMSchema is the schema for the vault secret resource IAM resources
+// (policy/binding). It will be merged with the base policy.
+func vaultSecretsAppIAMSchema(binding bool) schema.Schema {
+	// Determine the description based on if it is for the policy or binding
+	d := "Sets the vault secrets app IAM policy and replaces any existing policy."
```
dadgar, line 27:
```suggestion
	d := "Sets the Vault Secrets App IAM policy and replaces any existing policy."
```

internal/provider/vaultsecrets/resource_vault_secrets_app_iam_policy.go
```diff
@@ -0,0 +1,120 @@
...
+	if binding {
+		d = "Updates the vault secrets app IAM policy to bind a role to a new member. Existing bindings are preserved."
```
dadgar, line 29:
```suggestion
		d = "Updates the Vault Secrets App IAM policy to bind a role to a new principal. Existing bindings are preserved."
```

docs/resources/vault_secrets_app_iam_binding.md
dadgar: This is blocking from my perspective

examples/resources/hcp_vault_secrets_app_iam_policy/resource.tf
delores-hashicorp: The resource does not currently return resource id or resource name. I also went with resource name because of the conversation [here](https://hashicorp.slack.com/archives/C04RWBAC30W/p1712951592116149)

examples/resources/hcp_vault_secrets_app_iam_binding/resource.tf
delores-hashicorp: @pmmukh. Yes, it is. I created a ticket for this https://hashicorp.atlassian.net/browse/VAULT-26121

internal/provider/vaultsecrets/resource_vault_secrets_app_iam_policy.go
delores-hashicorp: The IAMPolicy APIs require either a resource ID or a resource name but not both. I initially wanted to make both optional for the TFP resource but [GetMutexKey](https://github.com/hashicorp/terraform-provider-hcp/blob/a6a588164a5608633b13413620ed76a89e8bfac6/internal/clients/iampolicy/iam.go#L31) requires a globally unique identifier, that is why I decided to make one of the fields required and drop the other. An option would have been to make both optional and call the resource manager to get the resource ID and use that as the mutex key but the resource manager does not have any API to get a resource by resource name or resource ID

examples/resources/hcp_vault_secrets_app_iam_policy/resource.tf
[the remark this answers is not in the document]
delores-hashicorp: Yes but vault secret resources do not currently return resource id or resource name

internal/provider/vaultsecrets/resource_vault_secrets_app_iam_policy.go
manish-hashicorp: > An option would have been to make both optional and call the resource manager to get the resource ID and use that as the mutex key but the resource manager does not have any API to get a resource by resource name or resource ID

That is what I was hoping. Would it be possible to do that? It can be done in subsequent PR.

examples/resources/hcp_vault_secrets_app_iam_binding/resource.tf
manish-hashicorp: Until HVS returns it can we construct the resource name in the provider?

internal/provider/vaultsecrets/resource_vault_secrets_app_iam_policy.go
delores-hashicorp: To do that, we will need to create a new API in the resource manager because we don't have an API that can get a resource's ID

manish-hashicorp: Based on this thread adding `resource_id` is increasing scope. How about we just add `resource_name` by constructing it in the provider and add the id later?

docs/resources/vault_secrets_app_iam_binding.md
delores-hashicorp: This has been fixed

aoripov (approved):
Changes are not related to HVD, so I'm approving without a review.
More details in this thread: https://hashicorp.slack.com/archives/C010V5LSZM3/p1713169986594289

docs/resources/vault_secrets_app_iam_binding.md
dadgar: Should the text be updated then?

internal/clients/iampolicy/resource_iam_policy.go
```diff
@@ -225,8 +225,10 @@ func setIamPolicyData(ctx context.Context, in, out TerraformResourceData, update
 		if diags.HasError() {
 			return diags
 		}
-
-		p.Etag = existingPolicy.Etag
+		// Resources might not have an existing policy
+		if existingPolicy != nil {
```
dadgar, line 229: Us this still needed?

internal/provider/vaultsecrets/resource_vault_secrets_app.go
```diff
@@ -58,6 +58,10 @@ func (r *resourceVaultsecretsApp) Schema(_ context.Context, _ resource.SchemaReq
 				Description: "The ID of the HCP organization where the project the HCP Vault Secrets app is located.",
 				Computed:    true,
 			},
+			"resource_name": schema.StringAttribute{
+				Computed:    true,
+				Description: "The app's resource name in the format secrets/project/<project ID>/app/<app ID>.",
```
dadgar, line 63:
```suggestion
				Description: "The app's resource name in the format secrets/project/<project ID>/app/<app Name>.",
```

dadgar (changes requested):

dmalch-hashicorp (approved):

internal/provider/vaultsecrets/resource_vault_secrets_app.go
```diff
@@ -184,3 +192,7 @@ func (r *resourceVaultsecretsApp) Delete(ctx context.Context, req resource.Delet
 		return
 	}
 }
+
+func generateResourceName(projectID, appName string) string {
```
pmmukh, line 196: could we import the resource-manager library here and use the helper func provided there for this generaiton?

dadgar (approved):

pmmukh (approved):
ran through local tests on the binding and policy resources, working great!! thank you!!

internal/provider/vaultsecrets/resource_vault_secrets_app.go
pmmukh: yeah that's fair!

PR #815, opened 2024-04-18
New Resource: Waypoint Action

HenryEstberg:
### :hammer_and_wrench: This adds a new resource `hcp_waypoint_action`.

Example config:
```
resource "hcp_waypoint_action" "test" {
  name = "some_test_action"
  description = "some description for fun"
  request = {
    custom = {
      method = "GET"
      url= "https://www.example.com"
      body = "some body"
      headers = {
        Content-Type = "application/json"
      }
    }
  }
}
```

Acceptance testing:

```sh
$ make testacc TESTARGS='-run=TestAccWaypoint_Action_basic'

...
```


internal/provider/waypoint/resource_waypoint_action_config.go
```diff
@@ -0,0 +1,696 @@
...
+	}
+	if actionCfg.Name != "" {
+		data.Name = types.StringValue(actionCfg.Name)
+	}
+	if actionCfg.Description != "" {
+		data.Description = types.StringValue(actionCfg.Description)
+	}
+	if actionCfg.ActionURL != "" {
+		data.ActionURL = types.StringValue(actionCfg.ActionURL)
+	}
```
catsby, line 418: I think we should always set these, to detect drift

internal/provider/waypoint/resource_waypoint_action_config.go
```diff
@@ -0,0 +1,696 @@
...
+	// These are the updated values
+	if !plan.Name.IsUnknown() {
+		modelBody.ActionConfig.Name = plan.Name.ValueString()
+	}
+	if !plan.Description.IsUnknown() {
+		modelBody.ActionConfig.Description = plan.Description.ValueString()
+	}
+	if !plan.ActionURL.IsUnknown() {
+		modelBody.ActionConfig.ActionURL = plan.ActionURL.ValueString()
+	}
```
catsby, line 520: We should always read these and send them for update, I think? I just tried this branch out and confirmed that if you create a config with a description, and then update the tf configuration to remove the description, Terraform reports that it will remove the value but after successfully applying, the value remains unchanged. Every `terraform plan` afterwards shows a perpetual diff, because we never send the updated `""` value to the API

internal/provider/waypoint/data_source_waypoint_action_config.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+	if actionCfg.Name != "" {
+		data.Name = types.StringValue(actionCfg.Name)
+	}
+	if actionCfg.Description != "" {
+		data.Description = types.StringValue(actionCfg.Description)
+	}
+	if actionCfg.ActionURL != "" {
+		data.ActionURL = types.StringValue(actionCfg.ActionURL)
+	}
```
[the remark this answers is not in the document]
HenryEstberg: The problem that I've encountered with this is that in some cases we do not set these, so the provider has them in state as `null` but the API sends back the empty string `""`. In this case, we really do want to ignore the empty string because if we don't, we get an error like this: `When applying changes to hcp_waypoint_action_config.test, provider
        "provider[\"registry.terraform.io/hashicorp/hcp\"]" produced an unexpected
        new value: .action_url: was null, but now cty.StringVal("")`

internal/provider/waypoint/data_source_waypoint_action_config.go
```diff
@@ -0,0 +1,229 @@
...
+	// In the future, expand this to accommodate other types of requests
+	if actionCfg.Request.Custom != nil {
+		data.Request.Custom = &customRequest{}
+		if actionCfg.Request.Custom.Method != nil {
+			methodString, err := ConvertMethodToStringType(*actionCfg.Request.Custom.Method)
+			if err != nil {
+				resp.Diagnostics.AddError(
+					"Unexpected HTTP Method",
+					"Expected GET, POST, PUT, DELETE, or PATCH. Please report this issue to the provider developers.",
+				)
```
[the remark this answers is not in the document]
HenryEstberg: Yes, probably better to error out here than to let the process continue.

internal/provider/waypoint/data_source_waypoint_action_config.go
```diff
@@ -0,0 +1,229 @@
...
+							},
+						},
+					},
+					/*"github": schema.ListNestedAttribute{
+						Description: "GitHub mode is configured to do various operations on GitHub Repositories.",
+						Optional:    true,
+					},
+					"agent": schema.ListNestedAttribute{
+						Optional: true,
+					},*/
```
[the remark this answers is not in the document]
HenryEstberg: Not really, turns out we will not be adding those in the near future.

internal/provider/waypoint/resource_waypoint_action_config.go
```diff
@@ -0,0 +1,696 @@
...
+		methodString = types.StringValue("PATCH")
+	case waypoint_models.HashicorpCloudWaypointActionConfigFlavorCustomMethodDELETE:
+		methodString = types.StringValue("DELETE")
+	default:
+		return methodString, fmt.Errorf("unknown method")
+	}
+	return methodString, nil
+}
+
+func ConvertMethodToEnumType(method string) (waypoint_models.HashicorpCloudWaypointActionConfigFlavorCustomMethod, error) {
```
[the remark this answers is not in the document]
HenryEstberg: Good catch, I don't see these being used elsewhere.

internal/provider/waypoint/resource_waypoint_action_config.go
```diff
@@ -0,0 +1,696 @@
...
+
+	if !plan.Description.IsUnknown() {
+		modelBody.ActionConfig.Description = plan.Description.ValueString()
+	}
+
+	if !plan.ActionURL.IsUnknown() {
+		modelBody.ActionConfig.ActionURL = plan.ActionURL.ValueString()
+	}
+
+	var diags diag.Diagnostics
```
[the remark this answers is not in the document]
HenryEstberg: This is actually necessary, when trying to use the assignment `:=` in those locations the compiler complains so without the initial declaration it will not work.

internal/provider/waypoint/resource_waypoint_action.go
```diff
@@ -0,0 +1,629 @@
...
+				Description: "The ID of the HCP organization where the Action is located.",
+				Computed:    true,
+			},
+			// An Action description must be fewer than 125 characters if set.
+			"description": schema.StringAttribute{
+				Description: "A description of the Action.",
+				Optional:    true,
+			},
+			"namespace_id": schema.StringAttribute{
+				Description: "Internal Namespace ID",
```
briancain, line 129: This could use a better description...I'm also not entirely sure where a user finds this value in HCP.

internal/provider/waypoint/resource_waypoint_action.go
```diff
@@ -0,0 +1,629 @@
...
+				Computed:    true,
+			},
+			// Action URL is required if the action is custom
+			"action_url": schema.StringAttribute{
+				Description: "The URL to trigger an action on. Only used in custom mode",
+				Optional:    true,
+			},
```
briancain, line 136: Part of me is thinking maybe we don't allow users to set this at all, does it need to be present in the schema because it exists under the Action config, or is it just a field that's user-set-able? If it's the second one, maybe we remove it from the provider.

internal/provider/waypoint/resource_waypoint_action.go
```diff
@@ -0,0 +1,629 @@
...
+				Description: "The kind of HTTP request this should trigger.",
+				Required:    true,
+				Attributes: map[string]schema.Attribute{
+					"custom": schema.SingleNestedAttribute{
+						Description: "Custom mode allows users to define the HTTP method, the request body, etc.",
+						Optional:    true,
+						Attributes: map[string]schema.Attribute{
+							"method": schema.StringAttribute{
+								Description: "The HTTP method to use for the request.",
+								Required:    true,
```
briancain, line 147: `schema.StringAttribute` might have a way to say like "always upper-case this value" or something?

internal/provider/waypoint/resource_waypoint_action.go
briancain: It also looks like we don't use the namespace_id field anyway for our namespace, but use Org and Project IDs to look it up. So maybe this field goes away?

internal/provider/waypoint/resource_waypoint_action.go
```diff
@@ -0,0 +1,629 @@
...
+
+	var diags diag.Diagnostics
+
+	// This is a proxy for the request type, as Custom.Method is required for Custom requests
+	if !plan.Request.Custom.Method.IsUnknown() && !plan.Request.Custom.Method.IsNull() {
+		modelBody.ActionConfig.Request.Custom = &waypoint_models.HashicorpCloudWaypointActionConfigFlavorCustom{}
+
+		method, err := convertMethodToEnumType(plan.Request.Custom.Method.ValueString())
+		if err != nil {
+			resp.Diagnostics.AddError(
```
briancain, line 244: Should we include the `err` too here?

internal/provider/waypoint/resource_waypoint_action.go
```diff
@@ -0,0 +1,629 @@
...
+		return
+	}
+
+	modelBody := &waypoint_models.HashicorpCloudWaypointWaypointServiceCreateActionConfigBody{
+		ActionConfig: &waypoint_models.HashicorpCloudWaypointActionConfig{
+			Request: &waypoint_models.HashicorpCloudWaypointActionConfigRequest{},
+		},
+	}
+
+	modelBody.ActionConfig.Name = plan.Name.ValueString()
```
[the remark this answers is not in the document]
HenryEstberg: I like that idea, but I think that change is outside the scope of this PR. Currently the pattern for every Waypoint resource is to keep the two names separate, so I would not want to make that change here without changing the other resources. In addition, from what I have seen the current pattern of keeping the two seperate is also consistent across the rest of the `hcp` provider.

internal/provider/waypoint/resource_waypoint_action.go
HenryEstberg: It was only used to expose the namespace_id to the user, I never intended it for our use. Your comment made me realize that the field is not that useful however so I will just remove it.

internal/provider/waypoint/resource_waypoint_action.go
HenryEstberg: This could certainly be removed, it is present in the provider because the action config endpoint accepts it and I wasn't sure if it was redundant with the URL within the `request`. I will get rid of it in that case, the provider will function fine without it.

internal/provider/waypoint/resource_waypoint_action.go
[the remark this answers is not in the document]
briancain: Got it, makes sense!

internal/provider/waypoint/resource_waypoint_action.go
briancain: Yep, originally before the `Request` struct was present in actions, we just had `action_url`. Now it's kind of turned into a shortcut, where if you set it with no Request struct, we auto-create a simple GET request on that url. But maybe for the TF provider that shortcut might be more confusing 🤔

internal/provider/waypoint/resource_waypoint_action.go
HenryEstberg: I found it! There is a built-in `Validators` field for every attribute!

internal/provider/waypoint/resource_waypoint_action.go
HenryEstberg: The error that is returned is pretty redundant here, it is just a less verbose version of the error that is already added so I would say not in this case.

HenryEstberg:
Hey @manish-hashicorp could we please get an approval on this one? Thanks!

briancain (approved):

internal/provider/waypoint/data_source_waypoint_action.go
```diff
@@ -0,0 +1,207 @@
...
+		return
+	}
+
+	if actionModel.ID != "" {
+		data.ID = types.StringValue(actionModel.ID)
+	}
+	if actionModel.Name != "" {
+		data.Name = types.StringValue(actionModel.Name)
+	}
+	if actionModel.Description != "" {
```
catsby, line 160: I don't think we should do these checks here, instead we should just set the value to what the API gives us. For example (and hypothetical) if the `Description` were to be cleared out, this pattern would prevent the data source from updating it's value in state. As a rule we should just set the state to reflect the API

internal/provider/waypoint/data_source_waypoint_action.go
```diff
@@ -0,0 +1,207 @@
...
+		methodString, err := convertMethodToStringType(*actionModel.Request.Custom.Method)
+		if err != nil {
+			resp.Diagnostics.AddError(
+				"Unexpected HTTP Method",
+				"Expected GET, POST, PUT, DELETE, or PATCH. Please report this issue to the provider developers.",
+			)
+			return
+		} else {
+			data.Request.Custom.Method = methodString
+		}
```
catsby, line 185: style nit: because the above `if er != nil` always ends in `return`, we can omit the `else` here. It makes the code easier to reason about _in general_, even if in this specific case the `else` branch is just 1 line

catsby: Specifically meaning this instead:

```golang
if err != nil {
  // stuff
  return
}
data.Request.Custom.Method = methodString
```

internal/provider/waypoint/data_source_waypoint_action.go
```diff
@@ -0,0 +1,207 @@
...
+			}
+		} else {
+			data.Request.Custom.Headers = types.MapNull(types.StringType)
+		}
+	}
+	if actionModel.Request.Custom.URL != "" {
+		data.Request.Custom.URL = types.StringValue(actionModel.Request.Custom.URL)
+	}
+	if actionModel.Request.Custom.Body != "" {
+		data.Request.Custom.Body = types.StringValue(actionModel.Request.Custom.Body)
```
catsby, line 205: same comment here about always setting the value in state to whatever we get from the API

catsby (changes requested):
I think we need to change the data source to always set the values

internal/provider/waypoint/resource_waypoint_action.go
```diff
@@ -0,0 +1,607 @@
...
+				Attributes: map[string]schema.Attribute{
+					"custom": schema.SingleNestedAttribute{
+						Description: "Custom mode allows users to define the HTTP method, the request body, etc.",
+						Optional:    true,
+						Attributes: map[string]schema.Attribute{
+							"method": schema.StringAttribute{
+								Description: "The HTTP method to use for the request. Must be one of: 'GET', 'POST', 'PUT', 'DELETE', or 'PATCH'.",
+								Required:    true,
+								Validators: []validator.String{
+									stringvalidator.OneOf("GET", "POST", "PUT", "DELETE", "PATCH"),
```
catsby, line 140: nit: If we're requiring/validating here, then the `convertMethodToStringType` and `convertMethodToEnumType` methods seem superfluous to me when we can convert them like so:
```golang
method := waypoint_models.HashicorpCloudWaypointActionConfigFlavorCustomMethod(plan.Request.Custom.Method.ValueString())
modelBody.ActionConfig.Request.Custom.Method = &method
```

later in the read we can just type cast I think

```golang
data.Request.Custom.Method = types.StringValue(string(*actionCfg.Request.Custom.Method))
```

internal/provider/waypoint/resource_waypoint_action.go
```diff
@@ -0,0 +1,607 @@
...
+		return
+	}
+
+	if actionCfg.ID != "" {
+		data.ID = types.StringValue(actionCfg.ID)
+	}
+	if actionCfg.Name != "" {
+		data.Name = types.StringValue(actionCfg.Name)
+	}
+	if actionCfg.Description != "" {
```
catsby, line 356: same note on empty here.. if it gets edited remotely, the local state will not pick up on the change

catsby (commented):
just ~2 more comments

manish-hashicorp:
Will approve after waypoint team approves.

internal/provider/waypoint/resource_waypoint_action_config.go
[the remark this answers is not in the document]
catsby: This file has changed since I commented, it's now `resource_waypoint_action.go`, but the comment remains:

on line 244 and 301 you can use `:=` and remove the `var diag` declaration since there's no need to declare it in the larger scope.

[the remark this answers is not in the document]
catsby: I apologize, I should have first asked if you could share the compiler warning, could you paste that here if it's still complaining? Maybe I'm missing something!

internal/provider/waypoint/data_source_waypoint_action.go
```diff
@@ -0,0 +1,207 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-provider-hcp/internal/clients"
+)
+
+var _ datasource.DataSource = &DataSourceAction{}
+var _ datasource.DataSourceWithConfigValidators = &DataSourceAction{}
+
+func (d DataSourceAction) ConfigValidators(ctx context.Context) []datasource.ConfigValidator {
+	return []datasource.ConfigValidator{
+		datasourcevalidator.Conflicting(
```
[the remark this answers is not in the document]
HenryEstberg: Good catch, thank you for linking the documentation as well!

internal/provider/waypoint/data_source_waypoint_action.go
HenryEstberg: Makes sense to me, I will update how we are handling description those fields here so they can potentially be empty.

catsby (approved):

HenryEstberg:
The Waypoint team has approved now @manish-hashicorp could you please take another look? Thank you!

WenInCode (approved):

PR #816, opened 2024-04-18
Waypoint: Add variable options to template resource

paladin-devops:

This PR updates the `hcp_waypoint_application_template` resource and data source to set and retrieve input variable options for HCP Waypoint templates.

```sh
$ make testacc TESTARGS='-run=TestAccWaypoint_Application_Template_basic'
=== RUN   TestAccWaypoint_Application_Template_basic
--- PASS: TestAccWaypoint_Application_Template_basic (7.19s)
PASS

$ make testacc TESTARGS='-run=TestAccWaypointData_Application_Template_basic'
=== RUN   TestAccWaypointData_Application_Template_basic
--- PASS: TestAccWaypointData_Application_Template_basic (11.34s)
PASS
```


manish-hashicorp (approved):
Rubber stamping. Please get approval from waypoint team before merging.

catsby (commented):
review isn't done but I noticed I had these sitting here while I dug in to another thing..

internal/provider/waypoint/data_source_waypoint_application_template.go
```diff
@@ -123,6 +124,27 @@ func (d *DataSourceApplicationTemplate) Schema(ctx context.Context, req datasour
 					},
 				},
 			},
+			"variable_options": schema.ListNestedAttribute{
+				Optional:    true,
```
HenryEstberg, line 128: All of this should be `computed` for the data source unless it is identifying information used to retrieve the data.

internal/provider/waypoint/resource_waypoint_application_template.go
```diff
@@ -214,6 +242,22 @@ func (r *ApplicationTemplateResource) Create(ctx context.Context, req resource.C
...
+		diags = v.Options.ElementsAs(ctx, &strOpts, false)
+		if diags.HasError() {
+			return
+		}
+
+		varOpts = append(varOpts, &waypoint_models.HashicorpCloudWaypointTFModuleVariable{
+			Name:         v.Name.ValueString(),
+			VariableType: v.VariableType.ValueString(),
+			Options:      strOpts,
+			UserEditable: false,
```
HenryEstberg, line 257: Do we want to allow `UserEditable` to be set?

internal/provider/waypoint/resource_waypoint_application_template.go
```diff
@@ -361,6 +423,28 @@ func (r *ApplicationTemplateResource) Read(ctx context.Context, req resource.Rea
...
 
+	if appTemplate.VariableOptions != nil && len(appTemplate.VariableOptions) > 0 {
+		varOpts := []*tfcVariableOption{}
+		for _, v := range appTemplate.VariableOptions {
+			varOptsState := &tfcVariableOption{
+				Name:         types.StringValue(v.Name),
+				VariableType: types.StringValue(v.VariableType),
+			}
+
+			vOpts, diags := types.ListValueFrom(ctx, types.StringType, v.Options)
```
HenryEstberg, line 434: This is slightly different than what you're doing on line 351, might be worth making a helper function for reading variable options so that it's the same everywhere you do it.

internal/provider/waypoint/resource_waypoint_application_template.go
paladin-devops: Yes we do, I will add that as a `bool` field.

internal/provider/waypoint/resource_waypoint_application_template.go
paladin-devops: I added the helper `readVarOpts` [here](https://github.com/hashicorp/terraform-provider-hcp/blob/454a6b3469c16d5e7865dc3bad838e79a3131f8d/internal/provider/waypoint/resource_waypoint_application_template.go#L367-L393). This is used by both the data source and the resource.

internal/provider/waypoint/data_source_waypoint_application_template.go
```diff
@@ -123,6 +125,27 @@ func (d *DataSourceApplicationTemplate) Schema(ctx context.Context, req datasour
 					},
 				},
 			},
+			"variable_options": schema.ListNestedAttribute{
```
HenryEstberg, line 128: I think it could be potentially useful in the future to have `user_editable` accessible from the data source, if you agree it is probably worth adding in here.

HenryEstberg (commented):
I have one more comment that I'm including but otherwise everything looks good, I did some manual testing as well and it all works!

internal/provider/waypoint/data_source_waypoint_application_template.go
```diff
@@ -123,6 +125,31 @@ func (d *DataSourceApplicationTemplate) Schema(ctx context.Context, req datasour
...
+							Required:    true,
+							Description: "Variable type",
+						},
+						"options": &schema.ListAttribute{
+							ElementType: types.StringType,
+							Required:    true,
+							Description: "List of options",
+						},
+						"user_editable": &schema.BoolAttribute{
+							Required:    true,
```
catsby, line 147: nit/fyi: these `Required` fields don't make sense because this is a data-source. Don't feel like you need to change it though if TF doesn't complain

catsby (approved):

HenryEstberg (approved):

PR #796, opened 2024-03-22
Accept arguments as workflow_call

manish-hashicorp:
### :hammer_and_wrench: Description

Scheduled workflow call is failing because it is not getting "skip-tests" input. Providing a default value in "workflow_call" setting should help with that.

### :building_construction: Acceptance tests

- [ ] Are there any feature flags that are required to use this functionality?
- [ ] Have you added an acceptance test for the functionality being added?
- [ ] Have you run the acceptance tests on this branch?

Output from acceptance testing:

```sh
$ make testacc TESTARGS='-run=TestAccXXX'

...
```

PR #831, opened 2024-05-02
Fix skip-e2e-tests flag for Prerelease scheduled runs

janet:

Fixing the Prerelease Github Action for scheduled runs. 

After this PR: https://github.com/hashicorp/terraform-provider-hcp/pull/796 , we were still seeing the same error message in the runs [Example](https://github.com/hashicorp/terraform-provider-hcp/actions/runs/8904358607):
```
[Prerelease](https://github.com/hashicorp/terraform-provider-hcp/actions/runs/8904358607/workflow)
The template is not valid. .github/workflows/prerelease.yml (Line: 34, Col: 23): Unexpected value ''
```

Apparently scheduled jobs set all inputs as null ([example 1](https://stackoverflow.com/questions/72539900/schedule-trigger-github-action-workflow-with-input-parameters), [example 2](https://github.com/orgs/community/discussions/74698)), so we need to account for that case.

```sh

...
```

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -14,16 +14,6 @@ on:
         description: "Skip the E2E test suite for break-glass scenarios"
         type: boolean
         default: false
-  workflow_call:
-    inputs:
-      release-as-patch:
-        description: "Patch Release (default: Minor)"
-        type: boolean
-        default: false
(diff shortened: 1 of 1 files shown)


manish-hashicorp (approved):
Awesome! Thank you so much for fixing!!

PR #833, opened 2024-05-06
Waypoint application input vars

paladin-devops:

Add support for input vars to Waypoint applications.

On the `waypoint_application` resource, there is a distinction for how an input variable was set. `app_input_vars` are configurable by the `waypoint_application` resource. `template_input_vars` are input variables that are set for an application, but were configured by the template, which was used for the creation of the app. The `waypoint_application` data source does not make this distinction, and all input variables are "filed under" the `input_vars` attribute of the data source.

Until the HCP Waypoint API is updated to return this information (if it ever does), the `waypoint_application` resource assumes that the type of an input variable in the TF configuration for the resource _is_ the type, so long as the name of the variable matches. Arguably, this assumption _should_ be made by the API. If/when this changes, the code to work around this should be removed, and the provider should defer to the type that the API says an input variable is.

Acceptance test output:
```
--- PASS: TestAccWaypoint_Application_DataSource_basic (11.98s)
PASS

Process finished with the exit code 0

--- PASS: TestAccWaypoint_Application_DataSource_WithInputVars (17.98s)
PASS

Process finished with the exit code 0

=== RUN   TestAccWaypoint_ApplicationInputVariables
--- PASS: TestAccWaypoint_ApplicationInputVariables (7.39s)
PASS

Process finished with the exit code 0

--- PASS: TestAccWaypoint_Application_basic (5.42s)
PASS

Process finished with the exit code 0

=== RUN   TestAccWaypoint_ApplicationInputVariables_OnTemplate
--- PASS: TestAccWaypoint_ApplicationInputVariables_OnTemplate (7.71s)
PASS
```

Fixes WAYP-2173.


internal/provider/waypoint/resource_waypoint_application.go
```diff
@@ -50,6 +53,29 @@ type ApplicationResourceModel struct {
...
 	// ActionCfgs types.List `tfsdk:"action_cfgs"`
+
+	InputVars types.Set `tfsdk:"app_input_vars"`
+
+	// NOTE: At the time of writing this comment, TemplateInputVars is the only
+	// struct field that makes ApplicationResourceModel different from ApplicationDataSourceModel.
+	// One might see an opportunity here to use an embedded struct to avoid code duplication;
+	// however, this is not currently possible in the framework. See this issue for more details:
+	// https://github.com/hashicorp/terraform-plugin-framework/issues/242
+	TemplateInputVars types.Set `tfsdk:"template_input_vars"`
```
paladin-devops, line 64: Depending upon how [this issue](https://github.com/hashicorp/terraform-plugin-framework/issues/242) is addressed, we may be able to simplify this struct to embed `ApplicationDataSourceModel` into this one.

briancain (approved):
Did a quick read through the code, and overall looks good! Makes sense to defer action stuff for now...

internal/provider/waypoint/data_source_waypoint_application.go
```diff
@@ -80,6 +80,26 @@ func (d *DataSourceApplication) Schema(ctx context.Context, req datasource.Schem
 				Computed:    true,
 				Description: "Internal Namespace ID.",
 			},
+			"app_input_vars": schema.SetNestedAttribute{
```
catsby, line 83: Should we name them simply `input_variables`? The `app` prefix seems to stutter a bit (`application.app_input..`) and in the Template resource we use the full word `variables` and not just `vars`

internal/provider/waypoint/data_source_waypoint_application.go
```diff
@@ -146,5 +186,30 @@ func (d *DataSourceApplication) Read(ctx context.Context, req datasource.ReadReq
...
 	}
 
+	// A second API call is made to get the input vars set on the application
+	inputVars, err := clients.GetInputVariables(ctx, client, data.Name.ValueString(), loc)
+	if err != nil {
+		resp.Diagnostics.AddError(err.Error(), "Failed to fetch application's input variables.")
+		return
+	}
+
+	if len(inputVars) > 0 {
```
catsby, line 196: Is there a reason we only set these conditionally? Is it not possible for them to be set in a run, then have no vars in a subsequent run? If there is `0` can we just set the attribute to the `null` value for sets? 

In general, conditionally setting values leads to undetected drift, so we should avoid conditionally setting things in state unless we're very positive the scenario is "basically impossible"

internal/provider/waypoint/data_source_waypoint_application_template_test.go
```diff
@@ -52,19 +48,61 @@ func TestAccWaypointData_Application_Template_basic(t *testing.T) {
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttr(resourceName, "name", updatedName),
 					resource.TestCheckResourceAttr(dataSourceName, "name", updatedName),
-					resource.TestCheckResourceAttr(dataSourceName, "variable_options.0.name", "string_variable"),
-					resource.TestCheckResourceAttr(dataSourceName, "variable_options.0.variable_type", "string"),
-					resource.TestCheckResourceAttr(dataSourceName, "variable_options.0.options.#", "1"),
-					resource.TestCheckResourceAttr(dataSourceName, "variable_options.0.options.0", "a"),
```
catsby, line 58: why are these checks being removed?

internal/provider/waypoint/data_source_waypoint_application.go
```diff
@@ -100,8 +120,28 @@ func (d *DataSourceApplication) Configure(ctx context.Context, req datasource.Co
...
+
+	// deferred and probably a list or objects, but may possible be a separate
+	// ActionCfgs types.List `tfsdk:"action_cfgs"`
+
+	InputVars types.Set `tfsdk:"app_input_vars"`
+}
+
 func (d *DataSourceApplication) Read(ctx context.Context, req datasource.ReadRequest, resp *datasource.ReadResponse) {
-	var data ApplicationResourceModel
+	var data ApplicationDataSourceModel
```
catsby, line 144: Why the change here? The underlying data model remains the same; the type that backs the `types.Set` here is the same as in the `resource_application`

catsby: I see the data source has just input vars and the resource now has TemplateInputVars as well. Why the discrepancy?

internal/provider/waypoint/resource_waypoint_application.go
```diff
@@ -146,6 +212,38 @@ func (r *ApplicationResource) Configure(ctx context.Context, req resource.Config
...
 }
 
+type varConverter struct {
+	name         string
+	variableType string
+	value        string
+}
+
+// FromTerraform5Value implements the ValueConverter interface
+func (vc *varConverter) FromTerraform5Value(val tftypes.Value) error {
```
HenryEstberg, line 222:
```suggestion
func (vc *varConverter) FromTerraformValue(val tftypes.Value) error {
```
Is this a typo?

internal/provider/waypoint/resource_waypoint_application.go
```diff
@@ -272,9 +441,77 @@ func (r *ApplicationResource) Read(ctx context.Context, req resource.ReadRequest
...
+// readInputs reads the inputVars for an app into the application resource model.
+func readInputs(
+	ctx context.Context,
+	inputVars []*waypoint_models.HashicorpCloudWaypointInputVariable,
+	plan *ApplicationResourceModel,
+	varTypes map[string]string,
+) diag.Diagnostics {
+	var diags diag.Diagnostics
+
+	if inputVars != nil {
```
catsby, line 468: we should probably omit this check and just always do this in order to detect drift.

HenryEstberg (approved):
LGTM, with maybe one small nonblocking typo

internal/provider/waypoint/resource_waypoint_application.go
```diff
@@ -272,9 +441,77 @@ func (r *ApplicationResource) Read(ctx context.Context, req resource.ReadRequest
 		data.ReadmeMarkdown = types.StringNull()
 	}
 
+	inputVars, err := clients.GetInputVariables(ctx, client, data.Name.ValueString(), loc)
+	if err != nil {
+		resp.Diagnostics.AddError(err.Error(), "Failed to fetch application's input variables.")
+		return
+	}
+
+	resp.Diagnostics.Append(readInputs(ctx, inputVars, data, varTypes)...)
```
catsby, line 450: there is indirection here that I think is unnecessary. Is there a reason to embed the `readInputs` function call inside this `Append`? What about `readInputs` makes it particularly suited to return diagnostics as opposed to simply formatting and returning two `[]*InputVar` and we do the `SetValueFrom` inline with this READ method, and do `resp.Diagnostics.AddError` there?

catsby (commented):
I have some questions on the data model split and naming of things

internal/provider/waypoint/resource_waypoint_application.go
paladin-devops: `FromTerraform5Value` is correct, it implements the [`ValueConverter` interface](https://github.com/hashicorp/terraform-plugin-go/blob/05dc75aefa5b71406022d0ac08eca99f44fbf378/tftypes/value.go#L17-L24), allowing us to use `As` [here](https://github.com/hashicorp/terraform-provider-hcp/pull/833/files#diff-523041116371f0db18a6aa9f5699c5686d44f0bfea408dd44545e0ee5d868a5aR295).

internal/provider/waypoint/data_source_waypoint_application_template_test.go
paladin-devops: This test for the data source used the "basic" acceptance test resource config. That resource's template has a no-code module which I updated in a previous PR to be v0.0.3, which included variable options (to test that feature). We had a discussion where you requested that be tested _not_ in the "basic" acceptance test, so I [rolled that template's no-code module version back to v0.0.2 ](https://github.com/hashicorp/terraform-provider-hcp/pull/833/files#diff-24fe8b2f639120d51e8271a8582eac54f7c2c76be48da056f541f917559b38a2R162)(with _no_ variable options). That precipitated this change, because now there are no variables on the template, or the application that used that template.

internal/provider/waypoint/data_source_waypoint_application.go
paladin-devops: The discrepancy is because of the distinction that we recently discussed regarding splitting variables set on a template by the platform engineer into one "bucket", and variables chosen by the platform engineer to another. The data source cannot make the distinction because the input variables are not in the TF config of the data source (since it's intended to read that info from HCP Waypoint).

internal/provider/waypoint/data_source_waypoint_application.go
paladin-devops: Great catch, thanks for calling this out!

paladin-devops: Added in https://github.com/hashicorp/terraform-provider-hcp/pull/833/commits/cdca91b82746655864d850da19e6cf2a2989ad46.

internal/provider/waypoint/resource_waypoint_application.go
paladin-devops: Returning diagnostics (which is returned by the call to `types.SetValueFrom` within `readInputs`) helps reduce code duplication, because if we returned two `[]*InputVar` we'd need to make the `SetValueFrom` calls wherever `readInputs` is called. It's not a large amount of duplicated code though, so I'm OK with moving the `SetValueFrom` to the calling function side to reduce the unnecessary indirection.

internal/provider/waypoint/resource_waypoint_application.go
paladin-devops: Fixed in https://github.com/hashicorp/terraform-provider-hcp/pull/833/commits/8dad09e6db677d8af2bf3df37dfd88e58698cc9c.

catsby (approved):

internal/provider/waypoint/resource_waypoint_application.go
paladin-devops: FYI @HenryEstberg I ultimately ended up removing this code after chatting with @catsby and pivoting to using `ElementsAs` to do the conversion, over implementing the conversion in the loop later in this file via `As` (which is what used this interface implementation).

jasonpilz (approved):

PR #838, opened 2024-05-09
Add 404 handling from ResourceServiceGetIamPolicy

pmmukh:
### :hammer_and_wrench: Add 404 handling from ResourceServiceGetIamPolicy

- Currently, when we try to create a binding or policy on an HVS App that doesn't have any policy associated with it, the ResourceServiceGetIamPolicy done to fetch the existing call errors out, which bubbles up as an error in the terraform apply. 
- This PR adds handling for the 404 error in this case, to ignore it and allow the subsequent PUT call to proceed
- Also updates the docs to be under HCP Vault Secrets and fixes an example in one of them

```sh

...
```

### DIFF
--- a/.changelog/838.txt
+++ b/.changelog/838.txt
@@ -1,0 +1,4 @@
+```release-note:bug
+Fix first apply scenario for hcp_vault_secrets_app_iam_binding and hcp_vault_secrets_app_iam_policy, update docs
+
+```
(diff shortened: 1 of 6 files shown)


manish-hashicorp:
Looks good!

codergs (approved):
LGTM, I am not sure if Cloud Platform wants to own these docs or not, but I will assume this is the best categorization.

pmmukh:
Yeap so this recategorization and putting the docs into the HVS folder was recommended from platform, and I imagine going forward it will be the service team i.e us, that owns this long term.

manish-hashicorp (approved):
Looks good

PR #839, opened 2024-05-15
fix state import for Waypoint resources

catsby:
Fix an issue with importing existing Waypoint resources. Prior to this commit any `terraform import` for a waypoint resource could fail because the `projectID` value used to look up the resource by ID would be an empty string, causing a failure to look up the existing resource.


HenryEstberg (approved):

paladin-devops (approved):
LGTM

PR #850, opened 2024-05-28
Include rotated secrets for cloud vault secrets VAULT-22307

dhuckins:

Allows users to fetch rotating secrets using the `hcp_vault_secrets_app` and `hcp_vault_secrets_secret` data sources

#### hcp_vault_secrets_app
`hcp_vault_secrets_app`: keys from complex (rotating) secrets are concatenated in the same pattern we've been using
 
Given these secrets 
image
with
```hcl
data "hcp_vault_secrets_app" "example" {
  app_name = "sample-app"
}

output "all-secrets" {
  value = data.hcp_vault_secrets_app.example.secrets
  sensitive = true
}
```
gives (secret values print out to stdout successfully, am not including here)
```
❯ terraform output all-secrets
tomap({
  "READ_ME" = ***
  "Step1" = ***
  "Step2" = ***
  "Step3" = ***
  "mongodb_atlas_password" = ***
  "mongodb_atlas_username" = ***
})
```

#### hcp_vault_secrets_secret
For backwards compatibility in the `hcp_vault_secrets_secret`, any rotating secret will be JSON encoded
using the same secrets setup above in the portal
```hcl
data "hcp_vault_secrets_secret" "rotating" {
  app_name    = "sample-app"
  secret_name = "mongodb_atlas"
}

output "rotating-secret" {
  value = data.hcp_vault_secrets_secret.rotating.secret_value
  sensitive = true
}
``` 
gives
```
❯ terraform output rotating-secret
"{\"password\":\"***\",\"username\":\"***\"}"
```
Because the user may not expect a json payload we emit a warning
image

 

```sh

...
```

### DIFF
--- a/internal/provider/vaultsecrets/data_source_vault_secrets_app.go
+++ b/internal/provider/vaultsecrets/data_source_vault_secrets_app.go
@@ -106,8 +106,20 @@ func (d *DataSourceVaultSecretsApp) Read(ctx context.Context, req datasource.Rea
 
 	openAppSecrets := map[string]string{}
 	for _, appSecret := range appSecrets {
-		secretName := appSecret.Name
-		openAppSecrets[secretName] = appSecret.Version.Value
+		switch {
+		case appSecret.StaticVersion != nil:
+			openAppSecrets[appSecret.Name] = appSecret.StaticVersion.Value
+		case appSecret.RotatingVersion != nil:
+			for name, value := range appSecret.RotatingVersion.Values {
+				openAppSecrets[appSecret.Name+"_"+name] = value
+			}
+		default:
+			resp.Diagnostics.AddError(
+				"Unsupported HCP Secret type",
+				fmt.Sprintf("HCP Secrets secret type %q is not currently supported by terraform-provider-hcp", appSecret.Type),
+			)
+			return
+		}
 	}
 
 	data.ID = data.AppName
--- a/internal/provider/vaultsecrets/data_source_vault_secrets_secret.go
+++ b/internal/provider/vaultsecrets/data_source_vault_secrets_secret.go
@@ -109,8 +109,8 @@ func (d *DataSourceVaultSecretsSecret) Read(ctx context.Context, req datasource.
 		return
 	}
 
-	// NOTE: for backwards compatibility purposes, if the secret is not a static secret (aka a string)
-	// encode the complex secret as a json string
+	// NOTE: for backwards compatibility purposes, if the secret is not a static secret (a string)
+	// encode the complex secret as a JSON string
 	var secretValue string
 	switch {
 	case openSecret.StaticVersion != nil:
--- a/internal/provider/vaultsecrets/data_source_vault_secrets_secret.go
+++ b/internal/provider/vaultsecrets/data_source_vault_secrets_secret.go
@@ -121,6 +121,10 @@ func (d *DataSourceVaultSecretsSecret) Read(ctx context.Context, req datasource.
 			resp.Diagnostics.AddError(err.Error(), "could not encode rotating secret as json")
 			return
 		}
+		resp.Diagnostics.AddWarning(
+			"HCP Vault Secrets mismatched type",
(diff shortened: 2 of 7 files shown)


internal/provider/vaultsecrets/data_source_vault_secrets_app.go
```diff
@@ -106,8 +106,20 @@ func (d *DataSourceVaultSecretsApp) Read(ctx context.Context, req datasource.Rea
 
 	openAppSecrets := map[string]string{}
 	for _, appSecret := range appSecrets {
-		secretName := appSecret.Name
-		openAppSecrets[secretName] = appSecret.Version.Value
+		switch {
```
averche, line 109: Would it be cleaner to switch on secret.Type instead?

internal/provider/vaultsecrets/data_source_vault_secrets_secret.go
```diff
@@ -107,7 +108,31 @@ func (d *DataSourceVaultSecretsSecret) Read(ctx context.Context, req datasource.
 		resp.Diagnostics.AddError(err.Error(), "Unable to open secret")
 		return
 	}
-	secretValue := openSecret.Version.Value
+
+	// NOTE: for backwards compatibility purposes, if the secret is not a static secret (aka a string)
+	// encode the complex secret as a json string
```
averche, line 113:
```suggestion
	// NOTE: for backwards compatibility purposes, if the secret is not a static secret (a string)
	// encode the complex secret as a JSON string
```

internal/provider/vaultsecrets/data_source_vault_secrets_secret.go
```diff
@@ -107,7 +108,31 @@ func (d *DataSourceVaultSecretsSecret) Read(ctx context.Context, req datasource.
...
+	var secretValue string
+	switch {
```
averche, line 115: Same as above, switching on secret.Type might be cleaner

internal/provider/vaultsecrets/data_source_vault_secrets_secret.go
```diff
@@ -107,7 +108,31 @@ func (d *DataSourceVaultSecretsSecret) Read(ctx context.Context, req datasource.
...
+	case openSecret.RotatingVersion != nil:
+		secretData, err := json.Marshal(openSecret.RotatingVersion.Values)
+		if err != nil {
+			resp.Diagnostics.AddError(err.Error(), "could not encode rotating secret as json")
+			return
+		}
+		resp.Diagnostics.AddWarning(
+			"HCP Vault Secrets mismatched type",
+			"Attempted to get a rotating secret in a KV secret data source, encoding the secret values as JSON",
+		)
```
averche, line 127: I'm a bit unsure about this warning. We may want to keep this data source as an equivalent of [vault_generic_secret](https://registry.terraform.io/providers/hashicorp/vault/latest/docs/data-sources/generic_secret) data source. In this case, it should be perfectly valid to use it to fetch a rotating secret.

I think it's OK to leave it here as well.

averche (approved):
A few comments, nothing is blocking, LGTM!

keeefer (approved):

internal/provider/vaultsecrets/data_source_vault_secrets_app.go
dhuckins: cleaner: yes

was concerned about the string type, are we going to switch to enum at some point?

internal/provider/vaultsecrets/data_source_vault_secrets_secret.go
dhuckins: lets keep, was in the original discussion when we planned this out

internal/provider/vaultsecrets/data_source_vault_secrets_app.go
averche: maybe 🤷

jasonpilz (approved):
LGTM

PR #864, opened 2024-06-11
Add input variables support to Waypoint add-ons

paladin-devops:

This PR enables config authors to configure their Waypoint add-ons with input variables.


teresamychu (approved):
LGTM

internal/provider/waypoint/data_source_waypoint_add_on.go
```diff
@@ -269,11 +311,40 @@ func (d *DataSourceAddOn) Read(ctx context.Context, req datasource.ReadRequest,
 		}
 	}
 
-	diags = readOutputs(ctx, addOn, &state)
+	ol := readOutputs(addOn.OutputValues)
+	if len(ol) > 0 {
+		state.OutputValues, diags = types.ListValueFrom(ctx, types.ObjectType{AttrTypes: outputValue{}.attrTypes()}, ol)
+	} else {
+		state.OutputValues = types.ListNull(types.ObjectType{AttrTypes: outputValue{}.attrTypes()})
+	}
```
HenryEstberg, line 319: Do you think you could explain the change here a little bit? It looks like this is to make this more readable but I want to better understand the why of this change.

internal/provider/waypoint/resource_waypoint_add_on_test.go
HenryEstberg, line 1: I like how thorough the test configs that you have added are!

HenryEstberg (approved):
Looks good to me! I have one question that I have included but it is nonblocking.

internal/provider/waypoint/data_source_waypoint_add_on.go
catsby: It simplifies the function to only operate on the common thing, the outputs, so we can reuse the method for both resources and data sources.

internal/provider/waypoint/data_source_waypoint_add_on.go
```diff
@@ -42,6 +42,28 @@ func (d *DataSourceAddOn) Metadata(ctx context.Context, req datasource.MetadataR
...
+	ApplicationID  types.String `tfsdk:"application_id"`
+	DefinitionID   types.String `tfsdk:"definition_id"`
+	OutputValues   types.List   `tfsdk:"output_values"`
+
+	TerraformNoCodeModule types.Object `tfsdk:"terraform_no_code_module"`
+
+	InputVars types.Set `tfsdk:"input_variables"`
+}
+
 //TODO: Make sure this schema is correct (do we want to include count or output values?)
```
catsby, line 67:
```suggestion
```

nit: I think we can remove this line now, yeah?

catsby (approved):

internal/provider/waypoint/data_source_waypoint_add_on.go
paladin-devops: I think so too!

jasonpilz (approved):

PR #869, opened 2024-06-21
Waypoint: Remove module version

paladin-devops:
The version of a no-code module is being removed from Waypoint templates and add-ons. The pinned version for a no-code module in HCP TF will be used at provision time.

### DIFF
--- a/.changelog/869.txt
+++ b/.changelog/869.txt
@@ -1,0 +1,3 @@
+```release-note:breaking-change
+waypoint: Remove version number from templates and add-on definition resources and data sources.
+```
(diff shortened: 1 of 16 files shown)


catsby (approved):
This is probably fine with the current state of the provider. In the future we'd want to add an additional field and deprecate the old one:

- https://developer.hashicorp.com/terraform/plugin/framework/deprecations#provider-attribute-removal

HenryEstberg (commented):
This all looks good, like you mentioned in your corresponding Waypoint PR this will ofc need to wait until that is merged and the SDK is updated as well. Because that SDK update will likely have to be included here, I will hold off on approving officially until then.

HenryEstberg (approved):

jasonpilz (approved):

PR #871, opened 2024-06-25
HCP Waypoint Application resource and data source reads output values

HenryEstberg:
This is a small addition to the existing HCP Waypoint Application resource and data source to allow both to read any output values that may be returned when getting an Application.

```sh
$ make testacc TESTARGS='-run=TestAccWaypoint_Application'
--- PASS: TestAccWaypoint_Application_basic (9.31s)
--- PASS: TestAccWaypoint_Application_DataSource_basic (16.14s)
--- PASS: TestAccWaypoint_Application_DataSource_WithInputVars (17.89s)
...
```

### DIFF
--- a/internal/provider/waypoint/data_source_waypoint_application.go
+++ b/internal/provider/waypoint/data_source_waypoint_application.go
@@ -238,36 +238,19 @@ func (d *DataSourceApplication) Read(ctx context.Context, req datasource.ReadReq
 		data.InputVars = types.SetNull(types.ObjectType{AttrTypes: InputVar{}.attrTypes()})
 	}
 
+	var diags diag.Diagnostics
+
 	// Read the output values from the application and set them in the plan
-	diags := readAppOutputsData(ctx, application, &data)
+	ol := readOutputs(application.OutputValues)
+	if len(ol) > 0 {
+		data.OutputValues, diags = types.ListValueFrom(ctx, types.ObjectType{AttrTypes: outputValue{}.attrTypes()}, ol)
+	} else {
+		data.OutputValues = types.ListNull(types.ObjectType{AttrTypes: outputValue{}.attrTypes()})
+	}
 	resp.Diagnostics.Append(diags...)
 	if resp.Diagnostics.HasError() {
 		return
 	}
 
 	resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)
-}
-
-func readAppOutputsData(ctx context.Context, app *waypoint_models.HashicorpCloudWaypointApplication, plan *ApplicationDataSourceModel) diag.Diagnostics {
-	var diags diag.Diagnostics
-	if app.OutputValues != nil {
-		outputList := make([]*outputValue, len(app.OutputValues))
-		for i, outputVal := range app.OutputValues {
-			output := &outputValue{
-				Name:      types.StringValue(outputVal.Name),
-				Type:      types.StringValue(outputVal.Type),
-				Value:     types.StringValue(outputVal.Value),
-				Sensitive: types.BoolValue(outputVal.Sensitive),
-			}
-			outputList[i] = output
-		}
-		if len(outputList) > 0 {
-			plan.OutputValues, diags = types.ListValueFrom(ctx, types.ObjectType{AttrTypes: outputValue{}.attrTypes()}, outputList)
-		} else {
-			plan.OutputValues = types.ListNull(types.ObjectType{AttrTypes: outputValue{}.attrTypes()})
-		}
-	} else {
-		plan.OutputValues = types.ListNull(types.ObjectType{AttrTypes: outputValue{}.attrTypes()})
-	}
-	return diags
 }
--- a/internal/provider/waypoint/data_source_waypoint_application.go
+++ b/internal/provider/waypoint/data_source_waypoint_application.go
@@ -103,7 +103,7 @@ func (d *DataSourceApplication) Schema(ctx context.Context, req datasource.Schem
 			},
 			"output_values": schema.ListNestedAttribute{
 				Computed: true,
-				Description: "The output values of the Terraform run for the Add-on, Sensitive values have type " +
+				Description: "The output values, stored by HCP Waypoint, of the Terraform run for the Add-on, Sensitive values have type " +
 					"and value omitted.",
 				NestedObject: schema.NestedAttributeObject{
 					Attributes: map[string]schema.Attribute{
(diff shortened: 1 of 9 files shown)


jasonpilz (approved):

internal/provider/waypoint/data_source_waypoint_application.go
```diff
@@ -211,5 +238,36 @@ func (d *DataSourceApplication) Read(ctx context.Context, req datasource.ReadReq
...
+	diags := readAppOutputsData(ctx, application, &data)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
 	resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)
 }
+
+func readAppOutputsData(ctx context.Context, app *waypoint_models.HashicorpCloudWaypointApplication, plan *ApplicationDataSourceModel) diag.Diagnostics {
```
catsby, line 251: `readAppOutputsData` and `readAppOutputs` are identical save for one of their inputs (either a `*ApplicationDataSourceModel` or `*ApplicationResourceModel`, respectively. They also internalize the use of `diag.Diagnostics` and buries whatever happens when reading the outputs. What do you think of combining into a single method that just accepts a slice of `[]*HashicorpCloudWaypointTFOutputValue` and returning a slice of `[]*outputValue`, similar to what was done in https://github.com/hashicorp/terraform-provider-hcp/pull/864/files#diff-a94b22f215a353a3d43c5eb6fb4f8c9bc64d8c08eb9a4e7ef419eefefc7eae2fR853 ?

internal/provider/waypoint/resource_waypoint_application.go
```diff
@@ -187,6 +190,31 @@ func (r *ApplicationResource) Schema(ctx context.Context, req resource.SchemaReq
 					},
 				},
 			},
+			"output_values": schema.ListNestedAttribute{
+				Computed: true,
+				Description: "The output values of the Terraform run for the Add-on, sensitive values have type " +
```
catsby, line 195:
```suggestion
				Description: "The output values of the Terraform run for the Add-on. Sensitive values have type " +
```

internal/provider/waypoint/data_source_waypoint_application.go
```diff
@@ -100,6 +101,31 @@ func (d *DataSourceApplication) Schema(ctx context.Context, req datasource.Schem
 					},
 				},
 			},
+			"output_values": schema.ListNestedAttribute{
+				Computed: true,
+				Description: "The output values of the Terraform run for the Add-on, sensitive values have type " +
```
catsby, line 106:
```suggestion
				Description: "The output values of the Terraform run for the Add-on. Sensitive values have type " +
```

catsby (commented):
I think it would look cleaner to consolidate the `read*` methods and reuse them, otherwise looks good

internal/provider/waypoint/data_source_waypoint_application.go
HenryEstberg: That looks like a good improvement, will do!

internal/provider/waypoint/data_source_waypoint_application.go
```diff
@@ -100,6 +101,31 @@ func (d *DataSourceApplication) Schema(ctx context.Context, req datasource.Schem
...
+				Computed: true,
+				Description: "The output values of the Terraform run for the Add-on, Sensitive values have type " +
+					"and value omitted.",
+				NestedObject: schema.NestedAttributeObject{
+					Attributes: map[string]schema.Attribute{
+						"name": schema.StringAttribute{
+							Computed:    true,
+							Description: "The name of the output value.",
+						},
+						"type": schema.StringAttribute{
```
briancain, line 114: Does terraform not have a more strict schema here for output variable values? Basically someone could type anything here right now and it would accept it (or make a typo).

internal/provider/waypoint/data_source_waypoint_application.go
```diff
@@ -100,6 +101,31 @@ func (d *DataSourceApplication) Schema(ctx context.Context, req datasource.Schem
 					},
 				},
 			},
+			"output_values": schema.ListNestedAttribute{
```
briancain, line 104: Also aren't these typically set using the [output](https://developer.hashicorp.com/terraform/language/values/outputs) resource themselves? What does `output_values` give us that this existing resource doesn't have?

internal/provider/waypoint/resource_waypoint_application.go
```diff
@@ -187,6 +190,31 @@ func (r *ApplicationResource) Schema(ctx context.Context, req resource.SchemaReq
 					},
 				},
 			},
+			"output_values": schema.ListNestedAttribute{
```
briancain, line 193: Same question about why we need this when https://developer.hashicorp.com/terraform/language/values/outputs already exists in Terraform

internal/provider/waypoint/data_source_waypoint_application.go
HenryEstberg: Good question, I agree that this doesn't add functionality that doesn't exist there. The advantage of this is it displays the information about output values that is already present in the API responses anyways and is returned by Waypoint in the UI so that users can always retrieve the information from Waypoint the same way. I can definitely see the argument for how it is redundant in this case.

internal/provider/waypoint/data_source_waypoint_application.go
HenryEstberg: I am a little confused by your question, because the output values are always computed the user shouldn't be typing anything here.

briancain: Ah, I see it's computed, yes nevermind 👍🏻

internal/provider/waypoint/data_source_waypoint_application.go
```diff
@@ -100,6 +101,31 @@ func (d *DataSourceApplication) Schema(ctx context.Context, req datasource.Schem
...
+				Computed: true,
+				Description: "The output values of the Terraform run for the Add-on, Sensitive values have type " +
+					"and value omitted.",
```
briancain, line 107:
```suggestion
				Description: "The output values, stored by HCP Waypoint, of the Terraform run for the Add-on, Sensitive values have type " +
					"and value omitted.",
```

internal/provider/waypoint/resource_waypoint_application.go
```diff
@@ -187,6 +190,31 @@ func (r *ApplicationResource) Schema(ctx context.Context, req resource.SchemaReq
...
+				Computed: true,
+				Description: "The output values of the Terraform run for the Add-on, Sensitive values have type " +
```
briancain, line 195:
```suggestion
				Description: "The output values, stored by HCP Waypoint, of the Terraform run for the Add-on, Sensitive values have type " +
```

briancain (approved):
I recommend making that docs change, because I think it's a pretty critical piece of information that I think will clear up any confusion as to what these outputs are.

PR #894, opened 2024-07-02
HCP Waypoint Readme update to accept unencoded strings

HenryEstberg:
This PR fixes a number of Waypoint provider issues identified in internal testing. The main changes are:

- The `readme_markdown_template` attribute is now able to handle both a base64 encoded string and an unencoded string to better match the underlying behavior of the Waypoint API, which accepts both for now. Eventually the ability to provide a base64 encoded version will be deprecated in the API, this change allows users to use both the new and old behavior until then. This change has been made for both the add-on definition and template resources.
- Various smaller documentation improvements to support the above change and to address smaller pieces of feedback recieved.

```sh
$  make testacc TESTARGS='-run=TestAccWaypoint'
=== RUN   TestAccWaypoint_Action_DataSource_basic
    data_source_waypoint_action_test.go:20: Waypoint Action tests skipped unless env 'HCP_WAYP_ACTION_TEST' set
--- SKIP: TestAccWaypoint_Action_DataSource_basic (0.00s)
=== RUN   TestAccWaypointData_Add_On_Definition_basic
--- PASS: TestAccWaypointData_Add_On_Definition_basic (14.95s)
=== RUN   TestAccWaypointData_Add_On_basic
--- PASS: TestAccWaypointData_Add_On_basic (26.04s)
=== RUN   TestAccWaypoint_AddOn_DataSource_WithInputVars
--- PASS: TestAccWaypoint_AddOn_DataSource_WithInputVars (23.76s)
--- PASS: TestAccWaypoint_Application_DataSource_basic (16.55s)
--- PASS: TestAccWaypoint_Application_DataSource_WithInputVars (18.11s)
=== RUN   TestAccWaypointData_Template_basic
--- PASS: TestAccWaypointData_Template_basic (11.99s)
=== RUN   TestAccWaypoint_Action_basic
    resource_waypoint_action_test.go:25: Waypoint Action tests skipped unless env 'HCP_WAYP_ACTION_TEST' set
--- SKIP: TestAccWaypoint_Action_basic (0.00s)
=== RUN   TestAccWaypoint_Add_On_Definition_basic
--- PASS: TestAccWaypoint_Add_On_Definition_basic (10.11s)
=== RUN   TestAccWaypoint_Add_On_basic
--- PASS: TestAccWaypoint_Add_On_basic (13.37s)
=== RUN   TestAccWaypoint_AddOnInputVariables
--- PASS: TestAccWaypoint_AddOnInputVariables (13.73s)
=== RUN   TestAccWaypoint_AddOnInputVariables_OnDefinition
--- PASS: TestAccWaypoint_AddOnInputVariables_OnDefinition (14.15s)
--- PASS: TestAccWaypoint_Application_basic (8.91s)
=== RUN   TestAccWaypoint_ApplicationInputVariables
--- PASS: TestAccWaypoint_ApplicationInputVariables (9.82s)
=== RUN   TestAccWaypoint_ApplicationInputVariables_OnTemplate
--- PASS: TestAccWaypoint_ApplicationInputVariables_OnTemplate (10.09s)
=== RUN   TestAccWaypoint_Template_basic
--- PASS: TestAccWaypoint_Template_basic (9.63s)
=== RUN   TestAccWaypoint_template_with_variable_options
--- PASS: TestAccWaypoint_template_with_variable_options (6.67s)
...
```

### DIFF
--- a/.changelog/894.txt
+++ b/.changelog/894.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+waypoint: The `readme_markdown_template` attribute for both template and add-on definition resources now accepts unencoded strings as well as base64 encoded strings.
+```
--- a/internal/provider/waypoint/resource_waypoint_add_on_definition.go
+++ b/internal/provider/waypoint/resource_waypoint_add_on_definition.go
@@ -105,7 +105,7 @@ func (r *AddOnDefinitionResource) Schema(ctx context.Context, req resource.Schem
 				},
 			},
 			"readme_markdown_template": schema.StringAttribute{
-				Description: "The markdown template for the Add-on Definition README.",
+				Description: "The markdown template for the Add-on Definition README. Must be base 64 encoded.",
 				Optional:    true,
 			},
 			"terraform_cloud_workspace_details": &schema.SingleNestedAttribute{
@@ -114,17 +114,18 @@ func (r *AddOnDefinitionResource) Schema(ctx context.Context, req resource.Schem
 				Attributes: map[string]schema.Attribute{
 					"name": &schema.StringAttribute{
 						Required:    true,
-						Description: "Name of the Terraform Cloud Workspace",
+						Description: "Name of the Terraform Cloud Project",
 					},
 					"terraform_project_id": &schema.StringAttribute{
 						Required:    true,
-						Description: "Tetraform Cloud Project ID",
+						Description: "Terraform Cloud Project ID",
 					},
 				},
 			},
 			"terraform_no_code_module": &schema.SingleNestedAttribute{
-				Required:    true,
-				Description: "Terraform Cloud no-code Module details.",
+				Required: true,
+				Description: "Terraform Cloud no-code Module details. Refer to " +
+					"https://developer.hashicorp.com/terraform/language/modules/sources for more details.",
 				Attributes: map[string]schema.Attribute{
 					"source": &schema.StringAttribute{
 						Required:    true,
@@ -155,9 +156,11 @@ func (r *AddOnDefinitionResource) Schema(ctx context.Context, req resource.Schem
 							Description: "List of options",
 						},
 						"user_editable": &schema.BoolAttribute{
-							Optional:    true,
-							Computed:    true,
-							Description: "Whether the variable is editable by the user creating an add-on",
+							Optional: true,
+							Computed: true,
+							Description: "Whether the variable is editable by the user creating an " +
+								"add-on. If options are provided, then the user may only use those " +
+								"options, regardless of this setting.",
 						},
 					},
 				},
--- a/internal/provider/waypoint/resource_waypoint_add_on_definition.go
+++ b/internal/provider/waypoint/resource_waypoint_add_on_definition.go
@@ -18,6 +18,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 	"github.com/hashicorp/terraform-provider-hcp/internal/clients"
 )
@@ -46,7 +47,7 @@ type AddOnDefinitionResourceModel struct {
 	Description            types.String `tfsdk:"description"`
 	ReadmeMarkdownTemplate types.String `tfsdk:"readme_markdown_template"`
 
-	TerraformCloudWorkspace  *tfcWorkspace        `tfsdk:"terraform_cloud_workspace_details"`
+	TerraformCloudWorkspace  types.Object         `tfsdk:"terraform_cloud_workspace_details"`
 	TerraformNoCodeModule    *tfcNoCodeModule     `tfsdk:"terraform_no_code_module"`
 	TerraformVariableOptions []*tfcVariableOption `tfsdk:"variable_options"`
 }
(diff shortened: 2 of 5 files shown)


internal/provider/waypoint/resource_waypoint_template.go
```diff
@@ -60,6 +61,13 @@ type tfcWorkspace struct {
 	TerraformProjectID types.String `tfsdk:"terraform_project_id"`
 }
 
+func (t tfcWorkspace) attrTypes() map[string]attr.Type {
+	return map[string]attr.Type{
+		"name":                 types.StringType,
+		"terraform_project_id": types.StringType,
+	}
+}
```
catsby, line 69: It's not clear to me what this is fixing or doing that we weren't doing before, since the field is still a `SingleNestedAttribute`.. what improvements does this make?

catsby (commented):
Seems fine but I'm curious about why the change

internal/provider/waypoint/resource_waypoint_template.go
HenryEstberg: That is there so that you can call `tfcWorkspace.attrTypes()` in places like line 337 of resource_waypoint_add_on_definition.go

HenryEstberg:
Hey @jasonpilz could I get an an approval on this one please? Thank you!

internal/provider/waypoint/resource_waypoint_add_on_definition.go
```diff
@@ -105,26 +106,29 @@ func (r *AddOnDefinitionResource) Schema(ctx context.Context, req resource.Schem
...
 			},
 			"readme_markdown_template": schema.StringAttribute{
-				Description: "The markdown template for the Add-on Definition README.",
+				Description: "The markdown template for the Add-on Definition README. Must be base 64 encoded.",
 				Optional:    true,
 			},
 			"terraform_cloud_workspace_details": &schema.SingleNestedAttribute{
-				Required:    true,
-				Description: "Terraform Cloud Workspace details",
+				Optional: true,
```
briancain, line 113: This is accurate? I don't think this should technically be optional because isn't `terraform_project_id` required?

I also noticed that the server ends up doing nothing with the workspace name field, so we probably shouldn't even be asking for that and instead just ask for `terraform_project_id`?

internal/provider/waypoint/resource_waypoint_add_on_definition.go
```diff
@@ -105,26 +106,29 @@ func (r *AddOnDefinitionResource) Schema(ctx context.Context, req resource.Schem
 				},
 			},
 			"readme_markdown_template": schema.StringAttribute{
-				Description: "The markdown template for the Add-on Definition README.",
+				Description: "The markdown template for the Add-on Definition README. Must be base 64 encoded.",
```
briancain, line 109: We should just base64 encode it for the user. Or at least document that users are expected to set this field using https://developer.hashicorp.com/terraform/language/functions/base64encode for this field. But given it's always required, maybe we should just be base64 encoding it ourselves transparently behind the scenes in the provider.

internal/provider/waypoint/resource_waypoint_add_on_definition.go
```diff
@@ -105,26 +106,29 @@ func (r *AddOnDefinitionResource) Schema(ctx context.Context, req resource.Schem
...
+						Description: "Terraform Cloud Project ID",
 					},
 				},
 			},
 			"terraform_no_code_module": &schema.SingleNestedAttribute{
-				Required:    true,
-				Description: "Terraform Cloud no-code Module details.",
+				Required: true,
+				Description: "Terraform Cloud no-code Module details. Refer to " +
+					"https://developer.hashicorp.com/terraform/language/modules/sources for more details.",
```
briancain, line 131: I like this, but in reality, the Waypoint server actually only supports a source that is fully quantified right? We don't support local sources for example, so linking them to these docs is a bit misleading.

briancain:
Can we get a better PR description here? It says doc updates, but this PR is doing more than updating the docs.

internal/provider/waypoint/resource_waypoint_add_on_definition.go
briancain: Basically the fact that this field was made to be base64 encoded is tech debt that we are planning to remove, so doing this behind the scenes means this providers resource in users module code won't need to be changed in the future when we remove the base64 field.

HenryEstberg: I am going to defer this change to another seperate PR then because for the time being all the Waypoint providers are doing this in the same way, so we should change all of them at once. I am going to write up another Jira ticket to track that.

HenryEstberg: Actually instead of doing that I am just going to make the change here across the different Waypoint resources.

.changelog/894.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+waypoint: TFC workspace details are now replaced with only `terraform_project_id` in the `hcp_wapyoint_add_on_definition` resource. The `readme_markdow_template` attribute for both template and add-on definition resources now accepts unencoded strings as well as base64 encoded strings.
```
jasonpilz, line 2:
```suggestion
waypoint: TFC workspace details are now replaced with only `terraform_project_id` in the `hcp_wapyoint_add_on_definition` resource. The `readme_markdown_template` attribute for both template and add-on definition resources now accepts unencoded strings as well as base64 encoded strings.
```

jasonpilz (approved):
LGTM

internal/provider/waypoint/resource_waypoint_add_on_definition.go
```diff
@@ -109,26 +110,31 @@ func (r *AddOnDefinitionResource) Schema(ctx context.Context, req resource.Schem
 				Optional:    true,
 			},
 			"terraform_cloud_workspace_details": &schema.SingleNestedAttribute{
-				Required:    true,
-				Description: "Terraform Cloud Workspace details",
+				Optional: true,
+				Computed: true,
+				Description: "Terraform Cloud Workspace details. If not provided, defaults to " +
+					"the HCP terraform project of the associated application.",
```
briancain, line 116:
```suggestion
					"the HCP Terraform project of the associated application.",
```

internal/provider/waypoint/resource_waypoint_add_on_definition.go
briancain: We should still fix this to not link to the docs!

internal/provider/waypoint/resource_waypoint_template.go
```diff
@@ -118,7 +126,7 @@ func (r *TemplateResource) Schema(ctx context.Context, req resource.SchemaReques
 			},
 			"readme_markdown_template": schema.StringAttribute{
 				Optional:    true,
-				Description: "Instructions for using the template (markdown format supported",
+				Description: "Instructions for using the template (markdown format supported). Must be base 64 encoded.",
```
briancain, line 129: I don't think this is true anymore?

HenryEstberg: Ah yes good catches, I fixed those in the old commit and missed them when going back through and reverting the project_id changes today, I'll clean those up while I'm shortening the commit history.

briancain (approved):

PR #1052, opened 2024-07-17
Waypoint: Require TF Project ID to be set, instead of a `terraform_cloud_workspace_details` resource

briancain:

Prior to this commit, the template and add-on definition resource required a user to set a full resource for specifying `terraform_cloud_workspace_details`. However, this resource required users to set a `name` which is a value that Waypoint does nothing with. Rather than requiring users to set this field that isn't used in the first place, this pull request instead introduces a `terraform_project_id` field that is required, and will use the add-on or template name for the erroneous field.

Fixes WAYP-2768 WAYP-2769

```sh

...
```

### DIFF
--- a/internal/provider/waypoint/resource_waypoint_add_on_definition.go
+++ b/internal/provider/waypoint/resource_waypoint_add_on_definition.go
@@ -111,7 +111,7 @@ func (r *AddOnDefinitionResource) Schema(ctx context.Context, req resource.Schem
 			},
 			"terraform_project_id": schema.StringAttribute{
 				Required:    true,
-				Description: "The ID of the Terraform Cloud Project to create workspaces in",
+				Description: "The ID of the Terraform Cloud Project to create workspaces in. The ID is found on the Terraform Cloud Project settings page.",
 				PlanModifiers: []planmodifier.String{
 					stringplanmodifier.RequiresReplace(),
 				},
--- a/internal/provider/waypoint/resource_waypoint_add_on_definition.go
+++ b/internal/provider/waypoint/resource_waypoint_add_on_definition.go
@@ -117,8 +117,9 @@ func (r *AddOnDefinitionResource) Schema(ctx context.Context, req resource.Schem
 				},
 			},
 			"terraform_cloud_workspace_details": &schema.SingleNestedAttribute{
-				Optional: true,
-				Computed: true,
+				DeprecationMessage: "terraform_cloud_workspace_details is deprecated, use terraform_project_id instead",
+				Optional:           true,
+				Computed:           true,
 				Description: "Terraform Cloud Workspace details. If not provided, defaults to " +
 					"the HCP Terraform project of the associated application.",
 				Attributes: map[string]schema.Attribute{
--- a/internal/provider/waypoint/resource_waypoint_add_on_definition.go
+++ b/internal/provider/waypoint/resource_waypoint_add_on_definition.go
@@ -11,7 +11,6 @@ import (
 	sharedmodels "github.com/hashicorp/hcp-sdk-go/clients/cloud-shared/v1/models"
 	"github.com/hashicorp/hcp-sdk-go/clients/cloud-waypoint-service/preview/2023-08-18/client/waypoint_service"
 	waypointModels "github.com/hashicorp/hcp-sdk-go/clients/cloud-waypoint-service/preview/2023-08-18/models"
-	waypoint_models "github.com/hashicorp/hcp-sdk-go/clients/cloud-waypoint-service/preview/2023-08-18/models"
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
@@ -47,7 +46,7 @@ type AddOnDefinitionResourceModel struct {
 	Description            types.String `tfsdk:"description"`
 	ReadmeMarkdownTemplate types.String `tfsdk:"readme_markdown_template"`
 
-	TerraformProjectId          types.String         `tfsdk:"terraform_project_id"`
+	TerraformProjectID          types.String         `tfsdk:"terraform_project_id"`
 	TerraformCloudWorkspace     *tfcWorkspace        `tfsdk:"terraform_cloud_workspace_details"`
 	TerraformNoCodeModuleSource types.String         `tfsdk:"terraform_no_code_module_source"`
 	TerraformVariableOptions    []*tfcVariableOption `tfsdk:"variable_options"`
@@ -250,16 +249,16 @@ func (r *AddOnDefinitionResource) Create(ctx context.Context, req resource.Creat
 		})
 	}
 
-	tfProjId := plan.TerraformProjectId.ValueString()
+	tfProjID := plan.TerraformProjectID.ValueString()
 	tfWsName := plan.TerraformCloudWorkspace.Name.ValueString()
 	if tfWsName == "" {
 		// NOTE: this field is optional anyways, so if its unset, lets just use
 		// the template name for it.
 		tfWsName = plan.Name.ValueString()
 	}
-	tfWsDetails := &waypoint_models.HashicorpCloudWaypointTerraformCloudWorkspaceDetails{
+	tfWsDetails := &waypointModels.HashicorpCloudWaypointTerraformCloudWorkspaceDetails{
 		Name:      tfWsName,
-		ProjectID: tfProjId,
+		ProjectID: tfProjID,
 	}
 
 	modelBody := &waypointModels.HashicorpCloudWaypointWaypointServiceCreateAddOnDefinitionBody{
@@ -485,16 +484,16 @@ func (r *AddOnDefinitionResource) Update(ctx context.Context, req resource.Updat
 		})
 	}
 
-	tfProjId := plan.TerraformProjectId.ValueString()
+	tfProjID := plan.TerraformProjectID.ValueString()
 	tfWsName := plan.TerraformCloudWorkspace.Name.ValueString()
 	if tfWsName == "" {
 		// NOTE: this field is optional anyways, so if its unset, lets just use
 		// the template name for it.
 		tfWsName = plan.Name.ValueString()
 	}
-	tfWsDetails := &waypoint_models.HashicorpCloudWaypointTerraformCloudWorkspaceDetails{
+	tfWsDetails := &waypointModels.HashicorpCloudWaypointTerraformCloudWorkspaceDetails{
 		Name:      tfWsName,
-		ProjectID: tfProjId,
+		ProjectID: tfProjID,
 	}
 
 	// TODO: add support for Tags
--- a/internal/provider/waypoint/resource_waypoint_add_on_definition.go
+++ b/internal/provider/waypoint/resource_waypoint_add_on_definition.go
@@ -11,6 +11,7 @@ import (
 	sharedmodels "github.com/hashicorp/hcp-sdk-go/clients/cloud-shared/v1/models"
 	"github.com/hashicorp/hcp-sdk-go/clients/cloud-waypoint-service/preview/2023-08-18/client/waypoint_service"
 	waypointModels "github.com/hashicorp/hcp-sdk-go/clients/cloud-waypoint-service/preview/2023-08-18/models"
+	waypoint_models "github.com/hashicorp/hcp-sdk-go/clients/cloud-waypoint-service/preview/2023-08-18/models"
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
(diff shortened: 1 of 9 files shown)


internal/provider/waypoint/resource_waypoint_add_on_definition.go
```diff
@@ -277,19 +294,6 @@ func (r *AddOnDefinitionResource) Create(ctx context.Context, req resource.Creat
 		modelBody.ReadmeMarkdownTemplate = readmeBytes
 	}
 
-	if !plan.TerraformCloudWorkspace.IsNull() && !plan.TerraformCloudWorkspace.IsUnknown() {
```
briancain, line 280: This was accidentally included in a merged PR so I am removing it so it's back to how it used to function.

internal/provider/waypoint/resource_waypoint_add_on_definition.go
```diff
@@ -333,18 +337,6 @@ func (r *AddOnDefinitionResource) Create(ctx context.Context, req resource.Creat
 	}
 	plan.Labels = labels
 
-	if addOnDefinition.TerraformCloudWorkspaceDetails != nil {
```
briancain, line 336: This was accidentally included in a merged PR so I am removing it so it's back to how it used to function.

internal/provider/waypoint/resource_waypoint_add_on_definition.go
```diff
@@ -425,18 +417,6 @@ func (r *AddOnDefinitionResource) Read(ctx context.Context, req resource.ReadReq
 	}
 	state.Labels = labels
 
-	if definition.TerraformCloudWorkspaceDetails != nil {
```
briancain, line 428: This was accidentally included in a merged PR so I am removing it so it's back to how it used to function.

internal/provider/waypoint/resource_waypoint_add_on_definition.go
```diff
@@ -539,19 +532,6 @@ func (r *AddOnDefinitionResource) Update(ctx context.Context, req resource.Updat
 		modelBody.ReadmeMarkdownTemplate = readmeBytes
 	}
 
-	if !plan.TerraformCloudWorkspace.IsNull() && !plan.TerraformCloudWorkspace.IsUnknown() {
```
briancain, line 542: This was accidentally included in a merged PR so I am removing it so it's back to how it used to function.

internal/provider/waypoint/resource_waypoint_add_on_definition.go
```diff
@@ -596,18 +576,6 @@ func (r *AddOnDefinitionResource) Update(ctx context.Context, req resource.Updat
 	}
 	plan.Labels = labels
 
-	if addOnDefinition.TerraformCloudWorkspaceDetails != nil {
```
briancain, line 599: This was accidentally included in a merged PR so I am removing it so it's back to how it used to function.

internal/provider/waypoint/resource_waypoint_add_on_definition.go
HenryEstberg: I think this one specifically is not quite right, the change in the PR that we should revert is [this](https://github.com/hashicorp/terraform-provider-hcp/pull/894/files#diff-514a4247ea434f616ba275ad89920b100f00d291f761cf4948bf3bab5da58283R336). By deleting the whole block, we can no longer read if the add-on definition's terraform cloud workspace details change, which isn't what we want.

briancain: Gotcha, I'll make sure to update this back to how it was then!

internal/provider/waypoint/resource_waypoint_add_on_definition.go
HenryEstberg: Actually ditto here about not wanting to delete the whole block.

internal/provider/waypoint/resource_waypoint_add_on_definition.go
HenryEstberg: Ditto here as well about not wanting to delete the whole block.

HenryEstberg (approved):
Looks good to me! Please make sure to run the acceptance tests and include the results in the PR description.

internal/provider/waypoint/resource_waypoint_add_on_definition.go
```diff
@@ -252,6 +256,18 @@ func (r *AddOnDefinitionResource) Create(ctx context.Context, req resource.Creat
...
+	if tfWsName == "" {
+		// NOTE: this field is optional anyways, so if its unset, lets just use
+		// the template name for it.
+		tfWsName = plan.Name.ValueString()
+	}
+	tfWsDetails := &waypointModels.HashicorpCloudWaypointTerraformCloudWorkspaceDetails{
+		Name:      tfWsName,
+		ProjectID: tfProjID,
+	}
+
```
catsby, line 270: If the `Name` is optional, what do you think of just doing away with it? Because `terraform_cloud_workspace_details` is currently `Optional`, but if users do specify it they need to specify both `name` and `terraform_project_id`. The code above only uses the `name` if it's specified in `terraform_cloud_workspace_details` (if I read it correctly).

Right now the only way it's used is if it's populated in the block, so the config will look like this:

```hcl
  terraform_project_id = "<ID for project foo>"
  terraform_cloud_workspace_details = {
    name                 = "<Name for project foo>"
    terraform_project_id = "<ID for project foo>"
  }
```

Maybe we should do away with the block entirely and just have this:

```hcl
terraform_project_id = "<ID for project foo>"
```

What do you think?

internal/provider/waypoint/resource_waypoint_add_on_definition.go
```diff
@@ -109,6 +109,10 @@ func (r *AddOnDefinitionResource) Schema(ctx context.Context, req resource.Schem
 				Description: "The markdown template for the Add-on Definition README (markdown format supported).",
 				Optional:    true,
 			},
+			"terraform_project_id": schema.StringAttribute{
+				Required:    true,
+				Description: "The ID of the Terraform Cloud Project to create workspaces in",
```
catsby, line 114: I imagine this isn't something we can change trivially, right? We couldn't just change the project ID value to another project, that wouldn't move the workspace in TFC to the new project, correct? Assuming that's the case, we should mark this as something that needs to trigger a replacement if changed as part of the `schema.StringAttribute`:

```go
  PlanModifiers: []planmodifier.String{
    stringplanmodifier.RequiresReplace(),
  },
```

catsby (approved):
I think we should either remove/deprecate the existing `terraform_cloud_workspace_details` block completely, or just mark the `name` part of it as `Optional` and provide a default (like how this PR does)

internal/provider/waypoint/resource_waypoint_add_on_definition.go
briancain: Hrm, I though I made this change such that `terraform_cloud_workspace_details` is completely optional now to define in a module: https://github.com/hashicorp/terraform-provider-hcp/pull/1052/files#diff-514a4247ea434f616ba275ad89920b100f00d291f761cf4948bf3bab5da58283R116-R117

I wanted to remove this too, but I think the Waypoint API requires that this struct is set in a request. And I didn't want to break that.

briancain: It sounds like we are going to mark `terraform_cloud_workspace_details` as deprecated via `DeprecationMessage`. Then let users just set `terraform_project_id`, and construct this behind the scenes to satisfy the Waypoint API validation.

internal/provider/waypoint/resource_waypoint_add_on_definition.go
briancain: That makes a lot of sense to me!

paladin-devops (approved):
Thanks for simplifying the template's project config setting on this client's side of things.

briancain:
Could someone from @hashicorp/cloud-core-platform-surfaces review this PR? Thanks

briancain:
There are some acceptance test failures but I think those failures are also on `main` and aren't related to this PR.

briancain:
Could one of @meirish @pierluc-codes or @jasonpilz review this? Thanks

internal/provider/waypoint/resource_waypoint_add_on_definition.go
pierluc-codes: Are you sure you are not breaking anything by moving this change?

internal/provider/waypoint/resource_waypoint_add_on_definition.go
```diff
@@ -109,9 +109,17 @@ func (r *AddOnDefinitionResource) Schema(ctx context.Context, req resource.Schem
...
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
 			"terraform_cloud_workspace_details": &schema.SingleNestedAttribute{
-				Optional: true,
-				Computed: true,
+				DeprecationMessage: "terraform_cloud_workspace_details is deprecated, use terraform_project_id instead",
```
pierluc-codes, line 120: That is cool that you are pointing to what to use, but can you do the extra mile for our user by pointing how to get that information?

internal/provider/waypoint/resource_waypoint_add_on_definition.go
briancain: Well, about a month ago this wasn't a breaking change because it was changed by a merged PR after this PR was open. I've had to rebase this PR multiple times since it's been open, which is what my comment indicates. But now, this project has done a release since this PR was open. So it's now probably a breaking change :/

I'm fine to put it back to how it was working previously if that's the intention, it's not a big deal to put it back.

internal/provider/waypoint/resource_waypoint_add_on_definition.go
briancain: We don't do this already for the existing param so I was following what has already been merged: https://registry.terraform.io/providers/hashicorp/hcp/latest/docs/resources/waypoint_tfc_config#optional

I'm fine adding more docs. What is expected here though on this deprecation message? Are you asking about how to get `terraform_project_id` value when you say "do the extra mile for our user by pointing how to get that information?"? Wouldn't that be on that fields docs, and not docs about `terraform_cloud_workspace_details` being deprecated?

internal/provider/waypoint/resource_waypoint_add_on_test.go
```diff
@@ -228,6 +230,7 @@ resource "hcp_waypoint_template" "test" {
   summary = "some summary for fun"
   readme_markdown_template = base64encode("# Some Readme")
   terraform_no_code_module_source = "private/waypoint-tfc-testing/waypoint-template-starter/null"
+	terraform_project_id = "prj-gfVyPJ2q2Aurn25o"
```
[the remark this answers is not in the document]
briancain: This entire file is a mix of tabs and spaces. I agree. Ideally this would happen before these tests were merged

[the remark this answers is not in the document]
briancain: I tried to fix some of them in this PR, which is the changes in this test file you see that aren't related to `terraform_project_id`

internal/provider/waypoint/resource_waypoint_add_on_definition.go
pierluc-codes: Sure.

internal/provider/waypoint/resource_waypoint_add_on_definition.go
pierluc-codes: Yes please :)

internal/provider/waypoint/resource_waypoint_add_on_test.go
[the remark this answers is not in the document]
pierluc-codes: > . Ideally this would happen before these tests were merged

I am not sure what it means since you want to merge in main? I expected the fix to be done in the scope of this PR.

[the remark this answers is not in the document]
briancain: Fixed in ead3fc328468fbbc1ad47abc128c760994222aca

And what I meant was, these tests existed before this PR was open, and it has plenty of mixed spacing and tabs:
* https://github.com/hashicorp/terraform-provider-hcp/pull/1052/files#diff-6535482890df9ededf38feccdf323552d43d8c7f50d4c443d9a45c9e811907e8L255-L354
* https://github.com/hashicorp/terraform-provider-hcp/pull/1052/files#diff-01be39748de68db6001914f74e3b360072065eb46d86f41a0c3fa9c555ebbe6cL283-L295

They make PR linters for things like this so next time it would be nice if that was in place!

internal/provider/waypoint/resource_waypoint_add_on_definition.go
briancain: [52d7f51](https://github.com/hashicorp/terraform-provider-hcp/pull/1052/commits/52d7f512dd4f62f73cc335726b105b97a925a2ce)

pierluc-codes (approved):

PR #1053, opened 2024-07-22
replace TokenFile with direct Token in provider configuration

liamcervante:

For Stacks, we want to support the ability to process the JWT directly instead of loading it via a file. This PR updates the provider configuration so that users can choose to specify the token directly instead of writing it into a file first. The old approach of writing into a file is still available.


internal/provider/provider.go
```diff
@@ -100,8 +102,15 @@ func (p *ProviderFramework) Schema(ctx context.Context, req provider.SchemaReque
 				NestedObject: schema.NestedBlockObject{
 					Attributes: map[string]schema.Attribute{
 						"token_file": schema.StringAttribute{
-							Required:    true,
-							Description: "The path to a file containing a JWT token retrieved from an OpenID Connect (OIDC) or OAuth2 provider.",
+							Optional:    true,
+							Description: "The path to a file containing a JWT token retrieved from an OpenID Connect (OIDC) or OAuth2 provider. Exactly one of `token_file` or `token` must be set.",
+							Validators: []validator.String{
+								stringvalidator.LengthAtLeast(1),
```
dadgar, line 108: Could we use `stringvalidator.ConflictsWith` or something similar to have TF validate only one is set?

liamcervante: Nice, didn't know they had such specific validators. I did end up using `ExactlyOneOf` instead of just `ConflictsWith` - I think that has the same effect but with an additional validation that one must be set.

dadgar (approved):

liamcervante:
I've tested this locally, and all is working 🎉

dhuckins (approved):

docs/index.md
```diff
@@ -134,7 +134,11 @@ resource "hcp_vault_cluster" "example" {
 Required:
 
 - `resource_name` (String) The resource_name of the Workload Identity Provider to exchange the token with.
-- `token_file` (String) The path to a file containing a JWT token retrieved from an OpenID Connect (OIDC) or OAuth2 provider.
+
+Optional:
+
+- `token` (String) The JWT token retrieved from an OpenID Connect (OIDC) or OAuth2 provider. Exactly one of `token_file` or `token` must be set.
```
pierluc-codes, line 140: What is the motivation of doing an exactly one?

internal/clients/client.go
```diff
@@ -122,18 +127,35 @@ func NewClient(config ClientConfig) (*Client, error) {
 		opts = append(opts, hcpConfig.WithClientCredentials(config.ClientID, config.ClientSecret))
 	} else if config.CredentialFile != "" {
 		opts = append(opts, hcpConfig.WithCredentialFilePath(config.CredentialFile))
-	} else if config.WorloadIdentityTokenFile != "" && config.WorkloadIdentityResourceName != "" {
+	} else if (config.WorkloadIdentityToken != "" || config.WorkloadIdentityTokenFile != "") && config.WorkloadIdentityResourceName != "" {
```
pierluc-codes, line 130: This is hard to parse for my out-of-context brain. Is it possible to do it better?

internal/clients/vault_secrets_preview.go
```diff
@@ -101,15 +101,15 @@ func GetRotatingSecretState(ctx context.Context, client *Client, loc *sharedmode
 
 // CreateMongoDBAtlasRotationIntegration NOTE: currently just needed for tests
 func CreateMongoDBAtlasRotationIntegration(ctx context.Context, client *Client, loc *sharedmodels.HashicorpCloudLocationLocation, integrationName, mongodbAtlasPublicKey, mongodbAtlasPrivateKey string) (*secretmodels.Secrets20231128MongoDBAtlasIntegration, error) {
-	body := secret_service.CreateMongoDBAtlasIntegrationBody{
+	body := secretmodels.SecretServiceCreateMongoDBAtlasIntegrationBody{
```
pierluc-codes, line 104: Can you give some context for those changes?

docs/index.md
liamcervante: This is two different ways of specifying the same information. Either the token has been written to disk, in which case you should specify the `token_file` variable, or the token has been loaded dynamically and Terraform just has it, in which case you should specify it directly via `token`.

It doesn't make sense to specify both of these together. Similarly, you must specify the token somehow if you are authenticating via the block, so you must specify at least one.

internal/clients/vault_secrets_preview.go
liamcervante: These are changes forced by the hcp-sdk-go upgrade that also added in the new `token` input when authenticating. The old approach (via the `secret_service` package) has gone and so wasn't compiling after the upgrade. 

There's no functional changes here, the location and names of the various structs have simply been moved around in the referenced library, so the code needs to be updated to point to the new locations.

internal/clients/client.go
liamcervante: I've pulled out the combined or logic into a standalone variable, and added some additional comments explaining the flow. Hopefully, that is better?

internal/clients/client.go
```diff
@@ -116,24 +121,47 @@ type ClientConfig struct {
 
 // NewClient creates a new Client that is capable of making HCP requests
 func NewClient(config ClientConfig) (*Client, error) {
+	hasWorkloadIdentityToken := config.WorkloadIdentityToken != "" || config.WorkloadIdentityTokenFile != ""
```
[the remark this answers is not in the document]
liamcervante: I've added more in 👍

docs/index.md
liamcervante: I've changed this to `AtLeastOneOf`, and the `token` field takes precedence over the `token_file` field.

liamcervante:
> I don't see any added or updated test

This is very difficult to test with an automated process. It's why there weren't any automatic tests for the previous implementation which just supported the `token_file` input.

We'd need to:

1. Generate a public/private key pair.
2. Host and operate an OAuth / OIDC exchange service that publishes the public key.
3. Configure a HCP account that trusts our new OAuth / OIDC service
4. Ask for the OAuth / OIDC service to generate a JWT using the private key
5. Then we can test the provider by inputting that JWT into the token or token_file fields

I've tested this manually by modifying TFC / Atlas / tfc-agent so that I can get a JWT from there which usually isn't possible, and definitely not something we should allow in normal operation. I then built the provider locally, and tested Terraform operations using the JWT I hacked out of TFC with my HCP account set up to trust TFC.

I could add some unit tests for this, but the client actually initiates the token exchange as part of it's initialisation, so we'd need to create some kind of mock or move the token exchange out of initialisation and refactor everything to call the token exchange after.

pierluc-codes:
>  This is very difficult to test with an automated process.

I would be minimally happy with only unit test that validate that the expected output are created from the given input.

liamcervante:
Added unit tests!

pierluc-codes (approved):

PR #1078, opened 2024-08-28
Implemented HVS AWS Integration resource

maxcoulombe:

* Implemented a resource to manage HVS AWS Integrations.
* Adjusted the audience and capabilities on the dynamic secret data source to fit with the HVS SDK update that happened earlier
* Added a decorator pattern so the integration resources for the next providers are easier to implement and maintain

acc-tests


internal/clients/vault_secrets_preview.go
```diff
@@ -157,11 +160,12 @@ func CreateMongoDBAtlasRotatingSecret(
...
 // CreateAwsIntegration NOTE: currently just needed for tests
-func CreateAwsIntegration(ctx context.Context, client *Client, loc *sharedmodels.HashicorpCloudLocationLocation, name, roleArn string) (*secretmodels.Secrets20231128AwsIntegration, error) {
+func CreateAwsIntegration(ctx context.Context, client *Client, loc *sharedmodels.HashicorpCloudLocationLocation, name, roleArn, audience string, capabilities []*secretmodels.Secrets20231128Capability) (*secretmodels.Secrets20231128AwsIntegration, error) {
 	body := secretmodels.SecretServiceCreateAwsIntegrationBody{
-		Name: name,
+		Name:         name,
+		Capabilities: capabilities,
 		FederatedWorkloadIdentity: &secretmodels.Secrets20231128AwsFederatedWorkloadIdentityRequest{
-			Audience: loc.OrganizationID,
+			Audience: audience,
```
maxcoulombe, line 168: Hardcoding the orgID as the audience was a wrong assumption I introduced when implementing the dynamic secret data source.

internal/provider/vaultsecrets/integration_utils.go
```diff
@@ -0,0 +1,109 @@
+package vaultsecrets
```
maxcoulombe, line 1: I tried to group as much provider-agnostic logic as possible here so the integration resources for the next providers are easier to implement.

internal/provider/vaultsecrets/integration_utils.go
```diff
@@ -0,0 +1,109 @@
...
+
+	diags.Append(concreteIntegration.fromModel(ctx, orgID, projID, model)...)
+	if diags.HasError() {
+		return diags
+	}
+
+	diags.Append(state.Set(ctx, &concreteIntegration)...)
+
+	return diags
+}
```
maxcoulombe, line 109: This decorator might need a tad convoluted but it encapsulates a lot of Terraform nuances that are easy to get wrong. It codify most of the gotchas mentioned in that [checklist](https://github.com/hashicorp/terraform-provider-hcp/blob/main/contributing/checklist-resource.md#crud-operations) such as:
* how to handle a create operation when the resource already exists
* how to handle resource that gets deleted out of band
* how to manage delete for a resource that was already deleted
* etc.

examples/resources/hcp_vault_secrets_integration_aws/import.sh
```diff
@@ -0,0 +1,4 @@
+# Vault Secrets AWS Integration can be imported by specifying the name of the integration
+# Note that since the AWS secret access key is never returned on the Vault Secrets API,
+# the next plan or apply will show a diff for that field if using the access keys authentication method.
```
austingebauer, line 3: Good detail to include

austingebauer (approved):
Spent ~20m playing around and trying to break things without success 🎉 I think the new pattern is worth it. Thanks again for walking me through the changes earlier.

jasonpilz (approved):

PR #1097, opened 2024-09-19
Dynamic secret resource implementation

maxcoulombe:

Added the Vault Secrets dynamic secret resource.
Created some interfaces and abstractions so additional dynamic and rotating secrets are easier to implement.

Screenshot from 2024-09-20 14-39-19

...
```

### CODE REVIEW

File: docs/resources/vault_secrets_dynamic_secret.md [hunk 1]
```diff
@@ -0,0 +1,73 @@
+---
```
maxcoulombe, line 1: One of the downside of TF resources with a high level of variability is that the doc grows over time instead of being well scoped to each dynamic/rotating secret types.

internal/provider/vaultsecrets/dynamic_secret_aws.go
```diff
@@ -0,0 +1,99 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
```
maxcoulombe, line 2: The interfaces and abstractions may seem convoluted at first, but the benefit is that new secret types only need to add their new fields to the schema and implement the `dynamicSecret` or future `rotatingSecret` interface.

internal/provider/vaultsecrets/resource_vault_secrets_dynamic_secret.go
```diff
@@ -0,0 +1,239 @@
...
+	// deleteFunc not necessary, all secrets use the same delete request
+	fromModel(s *DynamicSecret, model any) diag.Diagnostics
+}
+
+// dynamicSecretsImpl is a map of all the concrete dynamic secrets implementations by provider
+// so the Terraform resource can look up the correct implementation based on the resource secret_provider field
+var dynamicSecretsImpl = map[Provider]dynamicSecret{
+	ProviderAWS: &awsDynamicSecret{},
+	ProviderGCP: &gcpDynamicSecret{},
+}
```
maxcoulombe, line 37: This map of interface pattern has been used successfully many times in Vault ENT & HVS. It eliminates the proliferation of switch cases so adding new variants does not require modifications in several disjointed locations. You only need to implement the interface and add it to the collection.

internal/provider/vaultsecrets/resource_vault_secrets_dynamic_secret_test.go
```diff
@@ -0,0 +1,71 @@
...
+	"os"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+	"github.com/hashicorp/terraform-provider-hcp/internal/provider/acctest"
+)
+
+func TestAccVaultSecretsResourceDynamicSecret(t *testing.T) {
+	// TODO read env vars for the target provider
```
maxcoulombe, line 14: I'll add the acc tests if this more generic resource pattern is accepted.

internal/provider/vaultsecrets/vault_secrets_utils.go
```diff
@@ -16,8 +16,28 @@ import (
...
 )
 
-// sharedIntegrationAttributes are the attributes shared between all the Vault Secrets integrations
-var sharedIntegrationAttributes = map[string]schema.Attribute{
+type Provider string
+
+const (
+	ProviderAWS Provider = "aws"
+	ProviderGCP Provider = "gcp"
+)
```
maxcoulombe, line 24: As a follow-up we could publish the full list of providers in our public protos (like we do for capabilities) instead of redefining it here. I'm not sure we want to publish a static list with the ecosystem initiative though.

docs/resources/vault_secrets_dynamic_secret.md
```diff
@@ -0,0 +1,73 @@
...
+# hcp_vault_secrets_dynamic_secret (Resource)
+
+The Vault Secrets dynamic secret resource manages a dynamic secret configuration.
+
+## Example Usage
+
+```terraform
+resource "hcp_vault_secrets_dynamic_secret" "example_aws" {
+  app_name         = "my-app-1"
+  secret_provider  = "aws"
```
maxcoulombe, line 18: Provider is a reserved and very meaningful term in TF-land so I went with `secret_provider` to disambiguate. This is the term that was used for the [data sources](https://registry.terraform.io/providers/hashicorp/hcp/latest/docs/data-sources/vault_secrets_rotating_secret).

internal/provider/vaultsecrets/vault_secrets_utils.go
austingebauer: Ya I could see that being useful. HCP CLI repo has also defined these similar to here. Also unsure if that's desired with the ecosystem initiative.

examples/resources/hcp_vault_secrets_dynamic_secret/resource.tf
```diff
@@ -0,0 +1,21 @@
+resource "hcp_vault_secrets_dynamic_secret" "example_aws" {
+  app_name         = "my-app-1"
+  secret_provider  = "aws"
+  name             = "my_aws_1"
+  integration_name = "my-integration-1"
+  default_ttl      = "900s"
+  aws_assume_role = {
+    iam_role_arn = "arn:aws:iam::<account_id>>:role/<role_name>"
```
austingebauer, line 8:
```suggestion
    iam_role_arn = "arn:aws:iam::<account_id>:role/<role_name>"
```

internal/provider/vaultsecrets/vault_secrets_utils.go
```diff
@@ -58,22 +80,63 @@ var sharedIntegrationAttributes = map[string]schema.Attribute{
...
+		Validators: []validator.String{
+			slugValidator,
+		},
+	},
+	"secret_provider": schema.StringAttribute{
+		Description: "The third party platform the dynamic credentials give access to. One of `aws` or `gcp`.",
+		Required:    true,
+		PlanModifiers: []planmodifier.String{
+			stringplanmodifier.RequiresReplace(),
+		}, // TODO: Add validators check its in enum
```
austingebauer, line 100: Still todo? I checked this case and received `unsupported provider, expected one of [gcp aws], got 'gcpp'`.

internal/provider/vaultsecrets/vault_secrets_utils.go
```diff
@@ -58,22 +80,63 @@ var sharedIntegrationAttributes = map[string]schema.Attribute{
...
+		Required:    true,
+		PlanModifiers: []planmodifier.String{
+			stringplanmodifier.RequiresReplace(),
+		},
+		Validators: []validator.String{
+			secretNameValidator,
+		},
+	},
+	"integration_name": schema.StringAttribute{
+		Description: "The Vault Secrets integration name with the capability to rotate the secret.",
```
austingebauer, line 113: nit: mention of rotation capability might be slightly confusing in docs when used for dynamic secret resources?

austingebauer (approved):
LGTM! Tried a bunch of different lifecycle/modifications and couldn't break it 👍

internal/provider/vaultsecrets/vault_secrets_utils.go
maxcoulombe: I'll defer for now.

internal/provider/vaultsecrets/vault_secrets_utils.go
maxcoulombe: Whoops, no its a leftover.

internal/provider/vaultsecrets/vault_secrets_utils.go
maxcoulombe: Good catch, I'll make the text more generic.

jasonpilz (approved):

PR #1105, opened 2024-09-24
Fix doc by using templates

maxcoulombe:

Even if we use the default doc template as is we need to declare .md.tmpl files so the resources are in the correct sub-category.


emoncuso (approved):

austingebauer:
Update: Failure is being worked out by the cloud-core-platform team. Will merge once it is fixed.

PR #1109, opened 2024-10-01
Support pagination for Vault OpenAppSecrets preview API

divyapola5:

- [*] Have you added an acceptance test for the functionality being added?
- [*] Have you run the acceptance tests on this branch?

Screenshot 2024-10-01 at 9 26 42 AM

```sh

...
```


internal/clients/vault_secrets_preview.go
```diff
@@ -61,27 +61,41 @@ func OpenVaultSecretsAppSecrets(ctx context.Context, client *Client, loc *shared
...
+	var result []*secretmodels.Secrets20231128OpenSecret
+	nextPage := ""
 	for attempt := 0; attempt < retryCount; attempt++ {
-		secrets, err = client.VaultSecretsPreview.OpenAppSecrets(params, nil)
-		if err != nil {
-			var serviceErr *secret_service.OpenAppSecretDefault
-			ok := errors.As(err, &serviceErr)
-			if !ok {
+		for {
+			if nextPage != "" {
```
rodrigo-hcp, line 68: is the if statement necessary? We can send an empty pagination token

internal/clients/vault_secrets_preview.go
```diff
@@ -61,27 +61,41 @@ func OpenVaultSecretsAppSecrets(ctx context.Context, client *Client, loc *shared
...
-				continue
+			result = append(result, secrets.GetPayload().Secrets...)
+			pagination := secrets.GetPayload().Pagination
+			if pagination == nil || pagination.NextPageToken == "" {
+				return result, nil
 			}
-			return nil, err
+
+			nextPage = pagination.NextPageToken
+			break
```
rodrigo-hcp, line 90: I believe this `break` shouldn't be here

internal/clients/vault_secrets_preview.go
divyapola5: Yeah, removed the extra variable

internal/clients/vault_secrets_preview.go
divyapola5: Yup, meant to make the retry loop as inner loop but forgot to :) Thanks for pointing it out!

internal/clients/vault_secrets_preview.go
```diff
@@ -61,27 +61,34 @@ func OpenVaultSecretsAppSecrets(ctx context.Context, client *Client, loc *shared
...
+	for {
+		for attempt := 0; attempt < retryCount; attempt++ {
+			secrets, err = client.VaultSecretsPreview.OpenAppSecrets(params, nil)
+			if err != nil {
+				var serviceErr *secret_service.OpenAppSecretDefault
+				ok := errors.As(err, &serviceErr)
+				if !ok {
+					return nil, err
+				}
+				if shouldRetryWithSleep(ctx, serviceErr, attempt, []int{http.StatusTooManyRequests}) {
```
hghaf099, line 75: It seems that if there is an issue and we get `http.StatusTooManyRequests` consistently, we get stuck in the for ever loop. Would you think that the outer for loop should also check for a break condition to avoid getting stuck in a forever loop? Like maximum number of such error, or in a period of time.

internal/clients/vault_secrets_preview.go
```diff
@@ -61,27 +61,34 @@ func OpenVaultSecretsAppSecrets(ctx context.Context, client *Client, loc *shared
...
+				}
 				return nil, err
 			}
-			if shouldRetryWithSleep(ctx, serviceErr, attempt, []int{http.StatusTooManyRequests}) {
-				continue
+			if secrets == nil {
+				return nil, errors.New("unable to get secrets")
 			}
-			return nil, err
+			result = append(result, secrets.GetPayload().Secrets...)
```
hghaf099, line 83: Just curious, could `secrets.GetPayload()` return nil if `secrets` is not nil?

internal/clients/vault_secrets_preview.go
```diff
@@ -61,27 +61,34 @@ func OpenVaultSecretsAppSecrets(ctx context.Context, client *Client, loc *shared
...
+			pagination := secrets.GetPayload().Pagination
+			if pagination == nil || pagination.NextPageToken == "" {
+				return result, nil
+			}
+			params.PaginationNextPageToken = &pagination.NextPageToken
```
hghaf099, line 88: It seems that the intention here is to read all secrets of an app. If that is the case, can we bump the pageSize parameter of the pagination to its max value to have limited rounds of the API call?

hghaf099: If you think it is a good idea, you would also need to change the number secrets generated in your new test

divyapola5: Thank you for the suggestion @hghaf099. 
It is a good idea but the optimization is unhelpful if we increase the max size in future and forgot to update the provider.
More details in slack thread here: https://hashicorp.slack.com/archives/C06P685734L/p1728399960978739

internal/clients/vault_secrets_preview.go
divyapola5: No idea but I don't think it is a problem because this code returned secrets.GetPayload().Secrets even before my change for pagination support and we never had issues before. 
I looked at other usages of GetPayload() in this repo and did not find any checks for if GetPayload() is nil, so I don't think it can be nil if `secrets` is not nil.

divyapola5: No idea but I don't think it is a problem because this code returned secrets.GetPayload().Secrets even before my change for pagination support and we never had issues before. 
I looked at other usages of GetPayload() in this repo and did not find any checks for if GetPayload() is nil, so I don't think it can be nil if `secrets` is not nil.

divyapola5: No idea but I don't think it is a problem because this code returned secrets.GetPayload().Secrets even before my change for pagination support and we never had issues before. 
I looked at other usages of GetPayload() in this repo and did not find any checks for if GetPayload() is nil, so I don't think it can be nil if `secrets` is not nil.

internal/clients/vault_secrets_preview.go
divyapola5: I have updated the code and moved the logic of returning out of inner for loop. Please let me know if this does not look right. Thanks!

hghaf099 (approved):
LGTM! I think there is still room for improving test coverage. But, that should not be gating this PR.

PR #1118, opened 2024-10-15
Fix HVS app resource

maxcoulombe:

The HVS App resource had a couple of issues to resolve:
* Changing the app's name caused an error instead of triggering a recreate operation
* Deleting an app out of band caused an error instead of triggering a recreate operation
* Deleting an app that no longer exists failed instead of being lenient
* Importing an existing app was not supported
* The resource name was generated instead of relying on the API response
* Using an empty description failed with the error:
```
hcp_vault_secrets_app.example_2: Creation complete after 1s [id=my-app-2]
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to hcp_vault_secrets_app.example, provider "provider[\"registry.terraform.io/hashicorp/hcp\"]" produced an unexpected new value: .description: was null, but
│ now cty.StringVal("").
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

Instead of addressing each edge case individually, the resource now uses the decorator & HVS resource interface that were developed for the Integrations, Rotating & Dynamic Secrets resources. All these corner cases (and more!) mentioned above are properly managed and encapsulated in the decorator so new resources just work :tm: .

The diff is a bit hard to follow, consider this a rewrite.

The acceptance tests cover all the problematic cases listed above. This refactor is non-breaking, existing App resources & statefile representations are unchanged.

acc_test_app


raymonstah (approved):

jasonpilz (approved):
Looks good, thanks!

PR #1154, opened 2024-12-13
Fix a panic when creating HVS app or secret

divyapola5:
This change adds in a fix for the [bug](https://hashicorp.atlassian.net/browse/HV-3233). There is a panic when creating an app or a secret and this change addresses the issue.

Screenshot 2024-12-13 at 3 11 05 PM
Screenshot 2024-12-13 at 3 11 24 PM
Screenshot 2024-12-13 at 3 25 21 PM

```sh

...
```


.changelog/1154.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+creating a hvs app or kv secret should not panic
```
rodrigo-hcp, line 2: _nit_: I'd rather just repeat the PR title here

```suggestion
Fix a panic when creating HVS app or secret
```

divyapola5: done

rodrigo-hcp (approved):

maxcoulombe:
The changes look good. I did find another bug that we could fix at the same time or as a follow-up PR - up to you.

If you change the name of a hcp_vault_secrets_secret I'd expect that the secret with the previous name gets deleted and the a secret with the new name gets created. However, the TF provider treats it as an update and creates the new secret while not deleting the previous one. Same if the app name changes, the resource gets duplicated instead of recreated.

I think it can be simply fixed by adding this to the "app_name" and "secret_name" on the hcp_vault_secrets_secret schema.
```
PlanModifiers: []planmodifier.String{
stringplanmodifier.RequiresReplace(),
},
```

Does it make sense?

maxcoulombe (approved):

divyapola5:
> The changes look good. I was able to reproduce the panic then when building this branch it fixed my TF project.
> 
> I did find another bug that we could fix at the same time or as a follow-up PR - up to you.
> 
> If you change the name of a hcp_vault_secrets_secret I'd expect that the secret with the previous name gets deleted and the a secret with the new name gets created. However, the TF provider treats it as an update and creates the new secret while not deleting the previous one. Same if the app name changes, the resource gets duplicated instead of recreated.
> 
> I think it can be simply fixed by adding this to the "app_name" and "secret_name" on the hcp_vault_secrets_secret schema.
> 
> ```
> PlanModifiers: []planmodifier.String{
> stringplanmodifier.RequiresReplace(),
> },
> ```
> 
> Does it make sense?

Ah did not know about this. Sure, will add it as part of this PR.

jasonpilz (approved):

PR #1157, opened 2024-12-18
Add RequiresReplace attribute for secret_name in hvs secret resource

divyapola5:

If you change the name of a hcp_vault_secrets_secret it is expected that the secret with the previous name gets deleted and the a secret with the new name gets created. However, the TF provider treats it as an update and creates the new secret while not deleting the previous one. This change adds a fix for this.

JIRA: https://hashicorp.atlassian.net/browse/HV-3235

```sh

...
```

Local test where I changed the app name and secret name, terraform plan shows 2 to be added and 2 to be destroyed
Screenshot 2024-12-18 at 2 07 17 PM

### DIFF
--- a/.changelog/1157.txt
+++ b/.changelog/1157.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-Fix a bug when updating HVS secret name
+Fix a bug where updating an HVS secret name or app name would not recreate the resource as expected
 ```
--- a/.changelog/1157.txt
+++ b/.changelog/1157.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+Fix a bug when updating HVS secret name
+```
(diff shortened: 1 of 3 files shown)


internal/provider/vaultsecrets/resource_vault_secrets_secret.go
```diff
@@ -76,6 +76,9 @@ func (r *resourceVaultsecretsSecret) Schema(_ context.Context, _ resource.Schema
 						"must contain only ASCII letters, numbers, and underscores; must not start with a number",
 					),
 				},
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
```
maxcoulombe, line 81: We should also add this to the app_name since it is not a modifiable field.

Can we add tests for that scenario? We can either use the TF test framework to verify changing the secret_name causes a recreate operation or add a TestConfigFunc that verifies that the secret with the old name no longer exists after the rename is applied.

divyapola5: Done

.changelog/1157.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+Fix a bug when updating HVS secret name
```
maxcoulombe, line 2:
```suggestion
Fix a bug where updating an HVS secret name would not recreate the resource as expected
```

maxcoulombe (approved):
Thanks!

pierluc-codes (approved):

PR #1162, opened 2025-01-04
Disable starter tier in vault_cluster terraform resource

ashwinuday:

As part of Vault starter SKU EOL enablement we are removing support for starter tier cluster creation. This PR introduces changes to vault_cluster tier validation to treat `starter` clusters as invalid. 

Testing:
Built the provider and tested vault-cluster terraform resource with the provider. 
Screenshot 2025-01-09 at 8 16 12 AM

```sh

...
```


jasonpilz (dismissed):

.changelog/1162.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:feature
+remove 'starter_small' as a valid tier for vault cluster
```
shivalisharma23, line 2:
```suggestion
remove 'starter_small' as a valid tier for hcp vault cluster
```

internal/providersdkv2/validators.go
```diff
@@ -17,6 +17,7 @@ import (
 	vaultmodels "github.com/hashicorp/hcp-sdk-go/clients/cloud-vault-service/stable/2020-11-25/models"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	vaulthelper "github.com/hashicorp/terraform-provider-hcp/internal/helpers"
```
shivalisharma23, line 20: We might need to remove the tier from the test here as well: https://github.com/hashicorp/terraform-provider-hcp/blob/0ff3e21c849d2e0e8af4ba1c1ccf0b912159837e/internal/providersdkv2/validators_test.go#L458

docs/guides/vault-scaling.md
```diff
@@ -2,7 +2,7 @@
 subcategory: ""
 page_title: "Resize or scale a Vault cluster"
 description: |-
-    Change a current HCP Vault cluster in terms of tiers (Dev, Starter, Standard, Plus) or sizes (S, M, L).
+    Change a current HCP Vault cluster in terms of tiers (Dev, Standard, Plus) or sizes (S, M, L).
```
shivalisharma23, line 5: We might need to change the file [vault-scaling.md.tmpl](https://github.com/hashicorp/terraform-provider-hcp/blob/main/templates/guides/vault-scaling.md.tmpl#L5), and then run `go generate`, which will automatically generate this file

ashwinuday: > We might need to change the file [vault-scaling.md.tmpl](https://github.com/hashicorp/terraform-provider-hcp/blob/main/templates/guides/vault-scaling.md.tmpl#L5), and then run `go generate`, which will automatically generate this file

Both md and md.tmpl files are generated from the description. Didn't add the md.tmpl, added it now. Thanks.

internal/providersdkv2/validators.go
ashwinuday: This fetches from API and compares the result, we won't be able to modify it until we change the vault-service API  model itself.

.changelog/1162.txt
ashwinuday: updated

shivalisharma23 (approved):
LGTM!!
Thanks @ashwinuday

pacoguzman:
What should happen with already existing clusters? We cannot update provider version without changing cluster tier

PR #1183, opened 2025-01-22
HCPCP-2202: Consul Acceptance Test Workflow

jasonpilz:
### :hammer_and_wrench: Description

* This PR onboards the `Consul` acceptance tests to an updated github action execution model as described in the linked RFC.

### Tests Migrated
```sh
TestAccConsulCluster
TestAccConsulSnapshot
```

### Supporting ENV Variables
```sh
CONSUL_HCP_ORGANIZATION_ID
CONSUL_HCP_PROJECT_ID
CONSUL_HCP_CLIENT_ID
CONSUL_HCP_CLIENT_SECRET
```

## Related
* https://github.com/hashicorp/terraform-provider-hcp/pull/1169
* https://github.com/hashicorp/terraform-provider-hcp/pull/1171
* RFC [HCP-483](https://go.hashi.co/rfc/hcp-483)
* https://github.com/hashicorp/terraform-provider-hcp/actions/workflows/_testacc_consul.yml
  * ~26 minutes total execution

PR #1191, opened 2025-01-24
HCPCP-2200: IAM Acceptance Tests Workflow

jasonpilz:

* This PR onboards the `IAM` acceptance tests to an updated github action execution model as described in the linked RFC. This includes the `iam` and `resourcemanager` packages.

### Tests Migrated
```sh
TestAccUserPrincipalDataSource
TestAccServicePrincipalDataSource
TestAccServicePrincipalKeyResource
TestAccServicePrincipalResource_ExplicitProject
TestAccServicePrincipalResource_Organization
TestAccServicePrincipalResource_Project
TestAccProjectDataSource
TestAccProjectIamBindingResource
TestAccProjectIamPolicyResource
TestAccProjectResource
TestAccOrganizationDataSource
TestAccOrganizationIamBindingResource
TestAccIAMPolicyDataSource
TestAccIAMPolicyDataSource_Validation
TestAccMultiProject
TestAccMultiProjectResource
TestAccWorkloadIdentityProviderResource
TestAccGroupDataSource
TestAccGroupIamBindingResource
TestAccGroupIamPolicyResource
TestAccGroupMembersResource
TestAccGroupResource
```

⚠️ 3 Tests have been skipped which require more intricate dependency config before they will pass. In an effort to onboard the bulk of the tests we propose to not block on these but follow up in subsequent PR to address these skipped tests:
```
TestAccGroupIamPolicyResource
TestAccGroupIamBindingResource
TestAccGroupMembersResource
```
These will be addressed via https://hashicorp.atlassian.net/browse/HCPCP-2215

### Supporting ENV Variables
```sh
IAM_HCP_ORGANIZATION_ID
IAM_HCP_PROJECT_ID
IAM_HCP_CLIENT_ID
IAM_HCP_CLIENT_SECRET
```
## Workflow
* https://github.com/hashicorp/terraform-provider-hcp/actions/workflows/_testacc_iam.yml
  * ~3 minutes total execution

## Related
* https://github.com/hashicorp/terraform-provider-hcp/pull/1190
* https://github.com/hashicorp/terraform-provider-hcp/pull/1188
* https://github.com/hashicorp/terraform-provider-hcp/pull/1187
* https://github.com/hashicorp/terraform-provider-hcp/pull/1185
* https://github.com/hashicorp/terraform-provider-hcp/pull/1183
* https://github.com/hashicorp/terraform-provider-hcp/pull/1169
* https://github.com/hashicorp/terraform-provider-hcp/pull/1171
* RFC [HCP-483](https://go.hashi.co/rfc/hcp-483)


.github/workflows/testacc.yml
```diff
@@ -22,9 +22,10 @@ jobs:
   # testacc_platform:
   #   name: Platform
   #   uses: ./.github/workflows/_testacc_platform.yml
-  # testacc_iam:
-  #   name: Iam
-  #   uses: ./.github/workflows/_testacc_iam.yml
+  testacc_iam:
+    name: Iam
```
itsjaspermilan, line 26:
```suggestion
    name: IAM
```

internal/provider/iam/data_user_principal_test.go
```diff
@@ -19,23 +19,23 @@ func TestAccUserPrincipalDataSource(t *testing.T) {
 		PreCheck:                 func() { acctest.PreCheck(t) },
 		Steps: []resource.TestStep{
 			{
-				Config: testAccUserPrincipalConfigEmailInput("cloud-experiences-tooling@hashicorp.com"),
+				Config: testAccUserPrincipalConfigEmailInput("hcp-terraform-provider-iam@hashicorp.com"),
```
itsjaspermilan, line 22: Thanks for changing this email! Will it be available in 1Password as well in case things need to be debugged?

internal/provider/iam/resource_group_iam_policy_test.go
```diff
@@ -14,6 +14,8 @@ import (
 )
 
 func TestAccGroupIamPolicyResource(t *testing.T) {
+	t.Skip("TODO Fix this test")
```
itsjaspermilan, line 17: Without adding too much overhead/scope, is it possible to run these using the old method until the new way is completed? I don't imagine we will be introducing a regression anytime soon but just want to be safe with these workflows.

itsjaspermilan (commented):
This is really cool! Do you have a link to the ticket that covers the remaining tests?

internal/provider/iam/resource_group_iam_policy_test.go
itsjaspermilan: If the work to add in these tests using the new methods is already underway then you can disregard this comment 😄

jasonpilz: Unfortunately the old way has been broken for over a year 😢 . This is because the scope of all the tests exceeded 6 hours of execution, which is the most the GH runners will go. Ordinarily I'd agree with you, and I don't like skipping tests, but I simply can't fix every broken test for all products at a first pass. I've done the ones I could, but some are trickier than others.

internal/provider/iam/data_user_principal_test.go
jasonpilz: Yep! The new login creds for this org are in 1Pass. I can also invite you to manage the org if you'd like

internal/provider/iam/resource_group_iam_policy_test.go
itsjaspermilan: By "old way" I was thinking more so of the old GH action that was replaced in this PR. I can understand that having redundant actions might make things confusing though so as long as we have a path forward I think this `skip` makes sense.

jasonpilz: Ah I see, yes sorry IAM is the only tests that actually work I forgot that. I am happy to work with you on getting them going if your willing? Will continue chat via slack.

itsjaspermilan (dismissed):

itsjaspermilan (approved):
🚀

pierluc-codes (approved):
🚀

Issue #1240, opened 2025-03-24
`hcp_waypoint_template` fails to create when setting `variable_options` without `options` sub parameter


drewmullen:
```hcl
resource "hcp_waypoint_template" "new_app" {
  name                            = "${var.app_name}-template"
  summary                         = "Deploys ssm param for a team"
  description                     = <<EOF
this description is very valuable
EOF
  terraform_project_id            = tfe_project.new_app.id
  terraform_no_code_module_source = "private/<redacted org nam>/no-code-ssm/aws"
  terraform_no_code_module_id     = data.tfe_no_code_module.ssm.id
  variable_options = [
    {
      name          = "param_name"
      user_editable = false
      variable_type = "string"
      options       = ["small", "medium", "large"]
    },
    {
      name          = "param_val"
      user_editable = true
      variable_type = "string"
    },
  ]
}

```

```shell
Plan: 1 to add, 0 to change, 0 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

hcp_waypoint_template.new_app: Creating...
╷
│ Error: Missing Resource State After Create
│ 
│   with hcp_waypoint_template.new_app,
│   on main.tf line 26, in resource "hcp_waypoint_template" "new_app":
│   26: resource "hcp_waypoint_template" "new_app" {
│ 
│ The Terraform Provider unexpectedly returned no resource state after having no errors in the resource creation. This is always an issue
│ in the Terraform Provider and should be reported to the provider developers.
│ 
│ The resource may have been successfully created, but Terraform is not tracking it. Applying the configuration again with no other
│ action may result in duplicate resource errors. Import the resource if the resource was actually created and Terraform should be
│ tracking it.
```

drewmullen:
Tests should not use hardcoded values:


https://github.com/hashicorp/terraform-provider-hcp/blob/cdb61f57c59df9555ed11c46f7eb9e10071ef16f/internal/provider/waypoint/resource_waypoint_template_test.go#L203-L219

Tests may also be failing:

```shell
TESTARGS="-run TestAcc_Waypoint_Template_basic" make testacc
==> Checking that code complies with gofmt requirements...
golangci-lint run --config ./golangci-config.yml 
internal/clients/iampolicy/resource_iam_binding.go:107:5: S1009: should omit nil check; len() for nil maps is defined as zero (gosimple)
[...]

PR #1241, opened 2025-03-26
Do not call ElementAs when parsing `tfe_waypoint_template.variable_options.options`

drewmullen:

Closes: #1240 

Instead of adding a new test I just modified the existing test to validate setting no `options` works. I'm not able to run acc tests without access to your env.

```sh
See explanation above
...
```

Some lint checks fail but not related to my files:
```shell
make fmtcheck
internal/clients/iampolicy/resource_iam_binding.go:107:5: S1009: should omit nil check; len() for nil maps is defined as zero (gosimple)
	if resp.Schema.Attributes == nil || len(resp.Schema.Attributes) == 0 {
	   ^
internal/clients/iampolicy/resource_iam_policy.go:114:5: S1009: should omit nil check; len() for nil maps is defined as zero (gosimple)
	if resp.Schema.Attributes == nil || len(resp.Schema.Attributes) == 0 {
	   ^
internal/provider/packer/testutils/testclient/operation.go:48:11: SA1006: printf-style function with dynamic format string and no further arguments should use print-style function instead (staticcheck)
			return fmt.Errorf(msg)
			       ^
internal/provider/packer/testutils/testclient/operation.go:51:11: SA1006: printf-style function with dynamic format string and no further arguments should use print-style function instead (staticcheck)
			return fmt.Errorf(msg)
			       ^
internal/providersdkv2/test_helpers_for_packer_test.go:113:11: SA1006: printf-style function with dynamic format string and no further arguments should use print-style function instead (staticcheck)
			return fmt.Errorf(msg)
```

### DIFF
--- a/.changelog/1241.txt
+++ b/.changelog/1241.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+Fixed a bug when `tfe_waypoint_template.variable_options` did not specify `options` list.
+```
--- a/internal/provider/waypoint/resource_waypoint_template.go
+++ b/internal/provider/waypoint/resource_waypoint_template.go
@@ -286,9 +286,11 @@ func (r *TemplateResource) Create(ctx context.Context, req resource.CreateReques
 	var varOpts []*waypoint_models.HashicorpCloudWaypointTFModuleVariable
 	for _, v := range plan.TerraformVariableOptions {
 		strOpts := []string{}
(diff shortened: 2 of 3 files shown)


drewmullen:
moved PR to [here](https://github.com/hashicorp/terraform-provider-hcp/pull/1242) for permission reasons

PR #1242, opened 2025-03-27
Bug: fix `waypoint_template.variable_options` when no `options` is included 

drewmullen:

Closes: #1240, #1244

Instead of adding a new test I just modified the existing test to validate setting no `options` works. I'm not able to run acc tests without access to your env.

```sh
See explanation above
...
```


HenryEstberg (approved):
Looks good! I did some manual testing by pulling your branch which all looked good. As long as the acceptance tests are passing, this is good to go. Thanks for chasing this down!

PR #1216, opened 2025-02-14
Remove StatusPage

jasonpilz:
### :hammer_and_wrench: Description

* The StatusPage resource is currently undergoing migration (to Incident.io) but the replacement isn't yet ready. Currently the previous checks are failing, so this PR removes the integration / warnings in the meantime

PR #1297, opened 2025-05-23
Restore status page check

alzuri:

Follow up to #1216 that restores status page checking using new API.
Compared to the previous approach this expands the component list under consideration, adds a missing http timeout, co-locates declared incident names alongside impacted components for more context, and shares logic to reduce duplication across the provider implementations.

```sh

...
```
## PCI review checklist

- [ ] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [ ] If applicable, I've worked with GRC to document the impact of any changes to security controls.

  Examples of changes to controls include access controls, encryption, logging, etc.

- [ ] If applicable, I've worked with GRC to ensure compliance due to a significant change to the in-scope PCI environment.

  Examples include changes to operating systems, ports, protocols, services, cryptography-related components, PII processing code, etc.

### DIFF
--- a/internal/statuspage/statuspage.go
+++ b/internal/statuspage/statuspage.go
@@ -1,0 +1,174 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package statuspage
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"slices"
+	"strings"
+	"time"
+
+	frameworkDiag "github.com/hashicorp/terraform-plugin-framework/diag"
+	sdkv2Diag "github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+)
+
+const (
+	statuspageURL = "https://status.hashicorp.com/api/v1/summary"
+	warnSummary   = "You may experience issues using HCP."
+	warnDetailFmt = "HCP is reporting the following:\n\n%s\n\nPlease check https://status.hashicorp.com for more details."
+)
+
+// IDs from the incident.io status page for components relevant to this provider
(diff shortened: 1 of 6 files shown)


internal/statuspage/statuspage.go
```diff
@@ -0,0 +1,174 @@
...
+	if result.errorMessage != "" {
+		diags.AddWarning(warnSummary, result.errorMessage)
+		return diags
+	}
+
+	if !result.operational {
+		diags.AddWarning(warnSummary, fmt.Sprintf(warnDetailFmt, result.statusMessage))
+	}
+
+	return diags
```
huikang, line 149: nit: should we return nil if everything is normal?

internal/statuspage/statuspage.go
```diff
@@ -0,0 +1,174 @@
...
+	"HCP Waypoint":      "01JK8R0BRY0Q21819AYRKH5GZZ",
+}
+
+// These groups contain many regions and have a top-level ID that is not returned in the API response
+var hcpGroupNames = []string{
+	"HCP Consul Dedicated",
+	"HCP Vault Dedicated",
+}
+
+type statuspage struct {
```
huikang, line 42: Can we add some comments with links to the definition of the status page structure?

huikang (approved):
Thanks for the PR. Left some non-blocking comments for considering.

alzuri:
> Would it be possible to add some tests?

added lots in the way that made sense to me (not quite as we discussed...)

also, added provider config and env levers for optionally disabling (I believe precisely as discussed)

pierluc-codes (commented):
good job 👏

pierluc-codes (approved):

PR #1510, opened 2026-07-09
Handle etag conflicts for resource control policy

aastha-mittal:

This PR adds PR3 behavior for `hcp_resource_control_policy` conflict handling and update stability.

Changes included:
- implemented real in-place `Update` behavior for `hcp_resource_control_policy`
- sends `etag` on set operations when available
- detects conflict responses from the API (`ABORTED`, `FAILED_PRECONDITION`, with HTTP 409/412 fallback)
- returns actionable Terraform diagnostics for conflict and non-conflict API failures
- keeps state updates gated behind successful write + read-after-write
- switches `enabled_constraints` handling to order-insensitive set semantics to avoid invalid plan / inconsistent result issues caused by backend ordering
- adds focused unit coverage for etag handling, conflict diagnostics, non-conflict diagnostics, and single-attempt behavior

This was needed to support optimistic concurrency behavior for resource control policy updates and to make stale-plan conflicts visible to users instead of silently overwriting newer backend state.

### Testing: 

This is what the tag conflict looks like when it does happen:
image

```sh

Not run in this branch.
```

Manual validation performed:
```sh
$ go test ./internal/provider/resourcemanager/...

ok  	github.com/hashicorp/terraform-provider-hcp/internal/provider/resourcemanager	1.8s
```

Manual Terraform validation performed with local provider binary:
```sh
- verified normal in-place update of enabled constraints
- verified stale saved plan returns conflict diagnostic:
  "Conflict setting organization resource control policy"
- verified diagnostic instructs user to rerun terraform apply
- verified no internal retry loop was observed during stale-plan conflict test
```

## PCI review checklist

- [ ] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [ ] If applicable, I've worked with GRC to document the impact of any changes to security controls.

- [ ] If applicable, I've worked with GRC to ensure compliance due to a significant change to the in-scope PCI environment.


internal/provider/resourcemanager/resource_organization_resource_control_policy.go
```diff
@@ -249,17 +308,80 @@ func (r *resourceOrganizationResourceControlPolicy) setPolicy(ctx context.Contex
...
+		return false
+	}
+}
+
+func setPolicyRPCCode(serviceErr *organization_service.OrganizationServiceSetResourceControlPolicyDefault) string {
+	if serviceErr == nil || serviceErr.GetPayload() == nil {
+		return codes.Unknown.String()
+	}
+
+	return codes.Code(serviceErr.GetPayload().Code).String()
```
anushkatayal, line 374: `payload.Code` can be empty here, which makes this show as `OK` in an error path. We could default to `Unknown` when code is missing so the error message isn’t misleading.

aastha-mittal: Just made these changes Anushka! Thanks!

bhawnaverma1 (approved):

anushkatayal (approved):

